PR #23, opened 2018-11-06
Add start/end dates for training period

Natalie-Caruana:
Need to check why unit test fails

### DIFF
--- a/gordo_components/cli.py
+++ b/gordo_components/cli.py
@@ -12,6 +12,8 @@ import yaml
 import click
 from gordo_components.builder import build_model
 
+import dateutil
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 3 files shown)


epa095 (changes requested):
looks good besides somethings minor.

epa095 (changes requested):

epa095 (approved):

epa095 (changes requested):

flikka (changes requested):
Test that fails can be fixed by adding the TIME_FROM/TO env variables to the test, but we need to decide if we want these "data about the data" to be in an undetermined amount of env variables, or put into a data_config object. I lean towards the latter.

PR #25, opened 2018-11-07
Dockerfiles: Improve layer cacheability for faster builds

epa095:
Added requirements.txt and pip-installed it as seperate commands
so those layers will be re-used if all that has changed is the
code. Does nothing on travis (since it uses a fresh machine each time),
but it makes local builds much faster.


milesgranger (commented):
Seems `COPY` [is preferred](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy) for standard file additions?

milesgranger (approved):

PR #32, opened 2018-11-13
Integrate gordo_components.serializer with model builder

### LINKED ISSUE #31: Make gordo-compoents use the new model serializer
We use pickle dump, but now we have a nice new model/pipeline serializer. We should use it in the model builder and server.

milesgranger:
Will close #31


epa095 (approved):
LGTM :+1:

PR #34, opened 2018-11-13
Add KerasAutoEncoder subclass of KerasModel

### LINKED ISSUE #33: Create KerasAutoEncoder
The fit in KerasModel calls fit on X and y, while for auto encoders we want to fit X to X. We suggest to subclass KerasModel as KerasAutoEncoder, which overrides fit to have X,y=none, give a warning if y is set, and then call super on X,X. 

Also remember to change the dataset so it returns y=None again.

milesgranger:
Will close #33


epa095 (approved):
NO

PR #37, opened 2018-11-14
Add mypy to testing

milesgranger:
Not sure if this is as desirable as the doctest integration...


flikka (commented):
I think we should add this, it is cool. I also checked that it ran locally. Which it did.

epa095 (approved):

PR #40, opened 2018-11-15
Add pinned requirements

### LINKED ISSUE #19: Pin dependency versions
Versions like "requests>=2.0" is dangerous, non-reproducible builds. Pin them.

milesgranger:
Will close #19


flikka (dismissed):
Question: In times like these, should we just as well maybe upgrade at least to newest pandas/numpy? Not really important, just askin...

epa095 (commented):
What's the thing with the two requirement files? Is one our actual requirements, and the other their dependencies again(+transitive closure)?

epa095 (approved):

PR #44, opened 2018-11-21
Add TF graph & bump seldon serving image -> 0.3

### LINKED ISSUE #43: Pin selond version in gordo-server
https://github.com/Statoil/gordo-components/blob/2ea40e313070fd044d031ef37dc7eb2a01aa5e77/Dockerfile-ModelServer#L12

milesgranger:
Will close #43 

This fixes the detached graph found when upgrading seldon serving image to 0.3, which was _probably_ hidden by the older seldon image because it was indexing into something which didn't exist, because the model itself failed.

### DIFF
--- a/Dockerfile-ModelBuilder
+++ b/Dockerfile-ModelBuilder
@@ -6,8 +6,9 @@ COPY . /code
 COPY .git /code/
 
 WORKDIR /code
-RUN python setup.py sdist && \
-    mv /code/dist/$(ls /code/dist | head -1) /code/dist/gordo-components-packed.tar.gz
+RUN rm -rf /code/dist \
+    && python setup.py sdist \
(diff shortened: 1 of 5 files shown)


epa095 (changes requested):

epa095 (approved):

epa095 (dismissed):
Looks good to me if you look into the one comment i added, and maybe squash away that one "checkpoint" commit (and possible others as well). Will mark as approve as I trust you. Merge when done.

epa095 (changes requested):
I must admit that I dont understand why it works now. Except the set_learning_phase this seems identical to what did not work. Do we think the set_learning_phase magically fixed it? What about the need to call e.g. 
```python
model._make_predict_function()
model._make_test_function()
model._make_train_function()
```

epa095 (approved):

PR #46, opened 2018-11-23
Convert MVP model AutoEncoder to transformer

### LINKED ISSUE #14: Ensure a working test-build-predict setup (including Seldon API) like the DS is presented in the notebooks for Åsgård.
Look at whats outputted today, agree what we want, and implement it.

* MinMaxScaler
* FeedForwardAutoencoder and the norm (which is the Sqrt(Sum(abs(input-output)^2))

milesgranger:
Will close #14 

Depends on #44 

Will create a subclassed `KerasAutoEncoder` to return both input and output while serving.


epa095 (approved):
LGTM, lets try it and see how it works! Squashing/rebase some commits, and then you can merge.

PR #47, opened 2018-11-23
Add model metadata / seldon tags route

### LINKED ISSUE #45: Seldon: Provide reasonable metadata as custom metadata
Seldon 0.3 [added](https://github.com/SeldonIO/seldon-core/blob/master/docs/wrappers/python.md#custom-meta-data) the ability to return custom key-value pairs for metadata. It makes sense for us to return sensible metadata, like e.g. the machine-name, training start/end date, and maybe a serialization of the model config(?). 

This can either be passed along as ENV variables to the seldon server, or alternatively stored as part of the model. The latter is maybe better, since it is metadata about the model.

milesgranger:
Will close #45 
Will close https://github.com/Statoil/gordo-infrastructure/issues/76

### DIFF
--- a/gordo_components/builder/build_model.py
+++ b/gordo_components/builder/build_model.py
@@ -3,8 +3,9 @@
 import os
 import logging
 import joblib
+import copy
 
-from gordo_components import serializer
+from gordo_components import serializer, __version__
 from gordo_components.dataset import get_dataset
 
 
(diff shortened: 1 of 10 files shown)


epa095 (commented):
I like it. But it is not reachable?

epa095 (approved):
Minor stuff, mostly comments/documentation actually. When fixed and git commits are fixed/reworded etc then you can merge

PR #50, opened 2018-11-28
Support FunctionTransformer steps in config

### LINKED ISSUE #49: Support FunctionTransformer serialization with existing function
Should be able to serialize the `FunctionTransformer` while serializing the function as a path to the function held inside gordo_components.

milesgranger:
Will close #49 

Allow config files to support `FunctionTransformer` definitions, where the `func` parameter is a function held inside of gordo_components and not some random function.

### DIFF
--- a/gordo_components/model/transformer_funcs/general.py
+++ b/gordo_components/model/transformer_funcs/general.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+
+"""
+Functions to be used within sklearn's FunctionTransformer
+https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.FunctionTransformer.html
+
+Each function SHALL take an X, and optionally a y.
+
+Functions CAN take additional arguments which should be given during the initialization of the FunctionTransformer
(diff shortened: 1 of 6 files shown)


epa095 (dismissed):
So, talked to Kristian, and I have come to agree that storing the metadata in the objects (e.g. pipeline) directly is a bit too invasive/ugly. We propose the following API:

- gordo_components.serializer.pipeline_serializer.dump can take an optional dict argument which is the metadata, and it will save it in the toplevel folder as e.g. metadata.json
- add a method load_metadata which explicitly loads only the metadata and returns the dict.

The dissadvantage is that the metadata is not represented together with the model in memory, and it is up to the caller of the library to keep the model and its metadata together. I think that is fine for now. The advantage of this approach is that it keeps the serializing/loading cleaner, and probably makes it easier if we want to move to anything else. And it also does not risk screwing up the objects we try to load/save (if they contain a metadata field).

epa095 (approved):
I see no way how this can backfire. Squash/reword/rebase and merge!

PR #52, opened 2018-12-03
Fix serializer from_pipeline definition mutation

### LINKED ISSUE #51: Components serializer: Stop it from mutating its arguments
We have observed that when trying to load an architecture mutates the passed architecture. The simplest fix is probably to copy the architecture in the top-level function. Alternatively go through the lower-level functions and change there the mutation happens.

milesgranger:
Close #51


flikka (approved):
Genious

PR #53, opened 2018-12-03
Set default log level to DEBUG

### LINKED ISSUE #48: Components: Set default loglevel=debug
We want the model builder and server to use loglevel debug by default, so we get usable loging

milesgranger:
Close #48

### DIFF
--- a/gordo_components/runtime/SeldonModel.py
+++ b/gordo_components/runtime/SeldonModel.py
@@ -16,6 +16,14 @@ class SeldonModel:
     The Seldon model
     """
     def __init__(self):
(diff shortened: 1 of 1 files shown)


flikka (changes requested):
So - this probably sets the "default" log level, but wouldn't it be more professional to use eg. an environment variable for this? So that the kubernetes pod or whatever can be given a variable, and then we use this rather? DEBUG default, but look for a "LOG_LEVEL" env variable, and set accordingly if present?

flikka (approved):

Issue #54, opened 2018-12-14
Make model-builder crash if one of the sensors have no data


flikka:
For example if the mp-fetcher did not find and did not fail on the tag. Real story.

PR #58, opened 2018-12-20
Add docker-based unit test for InfluxBackedDataset 

### LINKED ISSUE #55: Add influx/docker based testing for model-builder
In gordo-demo there is [some tests](https://github.com/Statoil/gordo-demo/blob/master/tests/test_prediction_client.py) which uses influxdb in docker to run tests which does real connection to docker. This could be quite useful in gordo-components as well.

But try to make it run the influx on another port that 8086 so it does not crash with actual running influxes.

Natalie-Caruana:
This will close #55 and #54


milesgranger (changes requested):

milesgranger (approved):

epa095 (approved):
LGTM

PR #64, opened 2019-01-04
Bump docker base to 3.6.8

### LINKED ISSUE #63: Standardize on docker image python:3.6.8-slim-stretch
We should standardize on the same python base image (unless it causes problems) for better image layer caching. 
We are now using mostly python:3.6.6-slim-stretch, but 3.6.8 is the new good one. 

https://github.com/Statoil/gordo-components/blob/7e8bb052aec0eb4651f08be09b5755ffca1be242/Dockerfile-ModelServer#L13

epa095:
This closes #63


milesgranger (approved):
VNPRYCMWYAR!

PR #70, opened 2019-01-08
Save model metadata in model builder

### LINKED ISSUE #62: Save reasonable model metadata when creating model
We have the ability to save model metadata with the serializer, and we also have the ability to expose the metadata from the server now. We should save any metadata which might be useful.
Proposed data to save:
- Model config
- Tag list
- Model builder version
- Datetime of model creation
- Training start/end date (maybe part of model config, or of dataconfig?)

https://github.com/Statoil/gordo-components/blob/934ae039c4a5dfaa4eb1a14db9e1038fa3bde1bd/gordo_components/builder/build_model.py#L49

Natalie-Caruana:
This will close #62.

### DIFF
--- a/gordo_components/builder/build_model.py
+++ b/gordo_components/builder/build_model.py
@@ -4,6 +4,8 @@ import os
 import logging
 import joblib
 import copy
+import datetime
+import time
 
 from gordo_components import serializer, __version__
 from gordo_components.dataset import get_dataset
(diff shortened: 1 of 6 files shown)


epa095 (commented):
LGTM, you can merge after rebasing on newest master! Good work.

epa095 (approved):

PR #72, opened 2019-01-08
Clean up tmp folder made by doctest

### LINKED ISSUE #71: Cleanup tmp-dirs created by doctest
Some places, e.g. https://github.com/Statoil/gordo-components/blob/580eceeedd7fc68b119d7e3c70f842d9199b9e10/gordo_components/serializer/serializer.py#L185 we write stuff to /tmp/, but we do not clean up afterwards. It is possible to use pytest tmp-stuff, and maybe that is the solution?

epa095:
We had /tmp/my-model laying around after running
pytest with doctest. Now we use a tempdirectory
which is deleted after use.

Fixes #71


milesgranger (approved):
Nice!

PR #84, opened 2019-01-22
Fix static assets for swagger docs

### LINKED ISSUE #83: Fix swagger file location for API docs
Works locally, but Ambassador doesn't know how to route the request starting at the base route. Can perhaps update the location in each model server to place it where Ambassador expects to find it. Or maybe add this as a route mapping in the Ambassador annotation in infrastructure.

image

milesgranger:
Will close #83 

image

### DIFF
--- a/gordo_components/server/server.py
+++ b/gordo_components/server/server.py
@@ -3,6 +3,8 @@
 import os
 import logging
 import timeit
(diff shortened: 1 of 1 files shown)


epa095 (approved):
LGTM except a weird sentence

PR #85, opened 2019-01-23
LSTM Autoencoder

Natalie-Caruana:


milesgranger (changes requested):
Think it's pretty close, nice work.

Also _need_ tests. :100: 

- Define a pipeline in yaml and test loading and serializing it.
- Ensure the fit/predict operates as expected
- Explicitly test your generator functions (given X, or X and y) and expected outputs
- Test it will fail if asked to train on a smaller dataset than `lookback_window`
- Any other potential complications you can think of.

epa095 (commented):
I LiKe It

milesgranger (commented):
Sweet! :+1:  Will be cool to have this. Just minor comments, (sorry).. also noticed how much `black` obliterates formatting when a comment is on the same line as a statement, left a few comments on those instances.

epa095 (commented):
Good work!

milesgranger (approved):

PR #88, opened 2019-01-24
Support dumps and loads for gordo serializer

### LINKED ISSUE #86: Serialize Keras model w/ local graph
If this code can pass, it opens up a good alternative to use Dask on Kubernetes and use the native Scikit-Learn gridsearch API distributed. 

```python
>>> model = KerasAutoEncoder('feedforward_symetric')
>>> serialized = pickle.dumps(model)
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    pickle.dumps(model)
TypeError: can't pickle _thread.RLock objects
```

or alternatively, register our existing serializer with Dask (http://distributed.dask.org/en/latest/serialization.html)

It currently fails because we store the local tensorflow graph within the model, which is needed for our model serving. 

- Either find a way to serve the model without storing the graph.
- or a way to serialize the graph. 

*Should include a test to ensure all models can be pickled.*

milesgranger:
Will close #86

### DIFF
--- a/gordo_components/serializer/__init__.py
+++ b/gordo_components/serializer/__init__.py
@@ -1,3 +1,3 @@
 from .pipeline_from_definition import pipeline_from_definition
 from .pipeline_into_definition import pipeline_into_definition
-from .serializer import dump, load, load_metadata
+from .serializer import dump, dumps, load, loads, load_metadata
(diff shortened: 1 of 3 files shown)


epa095 (commented):
Add a test which compares the byte-streams dump(x) == dump(loads(dumps(x))). You might have to loop over them, who knows.

epa095 (approved):
LGTM, minor comment

PR #91, opened 2019-01-25
Add testnodocker to run tests without requiring docker on machine

### LINKED ISSUE #89: It should be possible to run tests without docker
In `setup.cfg` either change the definition for `test` so it does not include the docker things, or maybe add another one which does ever test except the ones which needs docker.

milesgranger:
Will close #89


epa095 (approved):
L.O.V.E it. Green button it when conflict fixed

PR #93, opened 2019-01-29
Update watchman to just use target names

milesgranger:
Depends on https://github.com/equinor/gordo-infrastructure/pull/161

Infrastructure is updated to only allow URL safe names for machines by default, this removes Watchman's dependency on getting sanitized target names.


flikka (commented):
LGTM

flikka (approved):

Issue #78, opened 2019-01-16
Use gordo-components to do exploratory work


flikka:
Walk through a scenario where you want to experiment with different pre-processing things and different models and architectures - USING gordo-components.

* Use jupyter or a "script like" code
* Run through a default scenario
* Note down missing features (lack of data from training, difficulties etc)

This task should be done by data scientist and dev together (at least partially)

Depends on #https://github.com/equinor/gordo-infrastructure/issues/147

PR #97, opened 2019-01-30
Documentation and running gordo locally.

### LINKED ISSUE #87: Improve README / lib documentation
- [ ] Take an existing model / pipeline and convert it to a definition
- [ ] Load a pipeline from a definition
- [ ] (De)serialize a model/pipeline
- [ ] Various examples of a valid model section for the config file.

### LINKED ISSUE #95: Run gordo locally.
Option to run a config file through the pipeline locally.

Something like:
```
gordo local /path/to/config.yml
```

milesgranger:
Will close #87 and close #95 

Also part of (and maybe closes) #78 

Creating examples / documentation on how to use Gordo, both introductory and for local iterative experiments. During the course of this, should also shape how to actually implement running Gordo locally.

@flikka and I discussed the following three good examples of running Gordo locally.
These will depend on https://github.com/equinor/gordo-components/pull/92

- [x]  *Pipeline work* - How pipelines work; from config definition, to creating a definition from an actual Pipeline object

- [x] *Mid level* - Use a `Dataset` to get sensors for a given timerange and then train a Pipeline on that data.

- [x] *High level* - Use an existing config file to run everything through into a trained Pipeline


flikka (changes requested):
As you point out @milesgranger - probably needs updating to use the dataset classes directly, and not the get_dataset.

flikka (approved):

PR #98, opened 2019-01-30
Watchman should return metadata from servers 

### LINKED ISSUE #94: Watchman forward endpoint's metadata
- [ ] The endpoint is reported as the `/healtcheck` endpoint, this should probably be the naked endpoint
- [ ] Should return the metadata from each model in it's own keyspace for each endpoint.

milesgranger:
Will close #94 

image

### DIFF
--- a/gordo_components/watchman/server.py
+++ b/gordo_components/watchman/server.py
@@ -2,7 +2,7 @@
 
 import requests
 import logging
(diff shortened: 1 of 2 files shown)


epa095 (approved):
The commit msg could be better, otherwise we are GoOoD

PR #100, opened 2019-02-04
Support time-range based predictions

### LINKED ISSUE #99: Update APIs with time based predictions (Integrate gordo-demo into components)
A client should be able to query gordo with a time range, rather than raw data specifically.
As `gordo-demo` client does this, we should integrate _some_ parts of this into the gordo ML server to deliver this functionality.

milesgranger:
Will close #99 

Support ability for client to supply a time range and get back predictions from that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ install:
   - pip install -r requirements.txt
 
 script:
-  - python setup.py test
+  - python setup.py testall
   - make images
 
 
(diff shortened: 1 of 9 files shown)


epa095 (approved):
If your happy and you know it, merge that thing. if your happy and you know it, MERGE that thing! If your happy and you know it and you really wanna show it, if you happy and you know it MERGE THAT THING!

(after making nice commit msgs)

PR #104, opened 2019-02-11
Add Travis build badge to README

milesgranger:


flikka (approved):
Cool. I see that regarding the token used, Travis states `This token is only used to access the build status image, but we recommend you not use it on a publicly available site.`

So I think that would be fine then!

PR #120, opened 2019-02-20
Add dataset/pandas filter_rows

epa095:
The function filter_rows accepts a dataframe and an filter expression,
and returns the dataframe but only containing the rows which satisfy the
filter expression. The expression is a boolean combination of
comparisons between column names or numbers.

This closes equinor/gordo-infrastructure#146

I ended up also creating a RandomDataProvider since I though I would use it for testing, but I ended up just mocking a dataprovider. I can remove it, or keep it.

### DIFF
--- a/gordo_components/data_provider/providers.py
+++ b/gordo_components/data_provider/providers.py
@@ -1,11 +1,13 @@
 # -*- coding: utf-8 -*-
-import logging
 import os
-from datetime import datetime
+import random
+import logging
 
+from datetime import datetime
 import typing
 
 from cachetools import cached, TTLCache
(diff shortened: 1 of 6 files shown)


milesgranger (changes requested):

flikka (commented):
Apart from @milesgranger's comments: From a usage point of view. If people write this wrong, which I guess they will, even if we make the docs a bit better - I guess the only way to find out for them is to run the Gordo machine? Or is there a shorter and simpler way?

milesgranger (approved):
LGTM!

PR #123, opened 2019-02-21
 Add requirement numexpr as recommended by pandas 

### LINKED ISSUE #118: Install recomended pandas dependencies
Pandas has two dependencies which, according to their [words](https://pandas.pydata.org/pandas-docs/stable/install.html#install-recommended-dependencies):

> You are highly encouraged to install these libraries, as they provide speed improvements, especially when working with large data sets. 

We have not added them as requirements, maybe we should.

epa095:
Recommended but not required by pandas.
https://pandas.pydata.org/pandas-docs/stable/install.html#install-recommended-dependencies

I tried bottleneck as well, but it needs gcc, so I dropped it for now.

Maybe silly to add without benchmarking(?), but I guess we think the Pandas people know what they are doing when they recommend installing them?

This closes #118

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -31,6 +31,7 @@ keras-preprocessing==1.0.5  # via keras, tensorflow
 keras==2.2.4
 markdown==3.0.1           # via tensorboard
 markupsafe==1.1.0         # via jinja2
+numexpr==2.6.9
 numpy==1.16.0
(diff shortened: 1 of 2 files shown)


milesgranger (approved):
I think benchmarking would be a cool thing to add, but agree if they're making a special note about it... :+1:

Issue #122, opened 2019-02-21
gordo-compose build sub-commands


milesgranger:
If possible; as an example:
`gordo build --metadata key1,key2 timeseries --resolution 2T`

where `timeseries` is a subcommand with its own args/options and receives arguments/options passed in `build`

milesgranger:
Stale :fish:

Issue #121, opened 2019-02-21
Split Dataset into ie. DataAccessor & Dataset


milesgranger:
`Dataset`s should not care about their data supplier. ie. A dataset should accept data supplied by data lake or influx resource.

PR #126, opened 2019-02-21
Update cli.build to reference data config keys and not env vars

milesgranger:
Depends on https://github.com/equinor/gordo-infrastructure/pull/190

Part of setting up for #121 and #122 

train start/end and tags are now part of data_config from infrastructure workflow generation


epa095 (approved):
LGTM

PR #128, opened 2019-02-21
Implement feedforward_hourglass model builder

### LINKED ISSUE #124: Support dynamic NN architecture based on number of sensors
Something along the the lines of https://github.com/equinor/health-indicator/blob/c8daeb2e99fc6f29a4121f78f0360fda7defeaea/src/health/feedforward.py#L147 with default, but with a middle layer that is for example num sensors / 2

epa095:
- Renamed `feedforward_symetric` to `feedforward_model`, since it is not symmetric. 
- Added `feedforward_symmetric` which is correctly spelled, and actually symmetric. 
- Added `feedforward_hourglass` which constructs the hourglass model given `n_features`.
- Change the default kind from `feedforward_model` to `feedforward_hourglass`, and changed many examples. 

The rename will definitively break a bunch of yaml-files in infrastructure, too bad. Issue about that is posted as equinor/gordo-infrastructure#191

This closes #124

### DIFF
--- a/examples/Gordo-Workflow-High-Level.ipynb
+++ b/examples/Gordo-Workflow-High-Level.ipynb
@@ -71,7 +71,7 @@
     "    steps:\n",
     "      - sklearn.preprocessing.data.MinMaxScaler\n",
     "      - gordo_components.model.models.KerasAutoEncoder:\n",
-    "          kind: feedforward_symetric\n",
+    "          kind: feedforward_model\n",
     "\"\"\""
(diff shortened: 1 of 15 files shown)


flikka (changes requested):
Looks nice, with some comments, after a quick look.

milesgranger (approved):
LGTM given @flikka 's comments

flikka (approved):
Nicey.

PR #131, opened 2019-02-25
Implement DataProvider and Dataset separation

### LINKED ISSUE #121: Split Dataset into ie. DataAccessor & Dataset

milesgranger:
Will close #121 

InfluxBackedDataset and now DataLakeBackedDataset exposed a neeed to
treat the Dataset separate from where the data itself was coming from.

We now use dependency injection from Dataset to use a 'data_provider' to
fetch the data itself. And Dataset object will have their own properties
specific to Dataset operations

### DIFF
--- a/examples/Gordo-Workflow-High-Level.ipynb
+++ b/examples/Gordo-Workflow-High-Level.ipynb
@@ -18,17 +18,9 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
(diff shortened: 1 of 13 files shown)


epa095 (approved):
AWesOmE!

PR #133, opened 2019-02-28
Support ability to download model from server

### LINKED ISSUE #132: Ability to download trained model from server

milesgranger:
via route GET /download-model

Close #132 

image


epa095 (approved):
Whats there to say? Cool stuff. If the docstring ends up in the swagger view we are good to go

PR #134, opened 2019-03-01
Support model scoring

### LINKED ISSUE #77: Feature suggestion: goodness metric from model building
Provide a metric (or several) from a model building run. Could be just the auto-encoder avg error or something, or a "overfitting indicator" where the difference between (ratio or similar) the results on training data and some test data is reported, stored, and provided as metadata. Could be a separate test period, or a cross validation / train test type split.

Need to discuss this with the whole team, and then implement. Should also be used when assessing new architectures or hyperparameters.

### LINKED ISSUE #115: When training, do validation and store metric
A n-fold CV, storing and serving the metrics. Probably as a separate step to the actual building using the full training set

milesgranger:
Will close #77 and close #115

### DIFF
--- a/gordo_components/builder/build_model.py
+++ b/gordo_components/builder/build_model.py
@@ -60,16 +60,15 @@ def build_model(
     logger.debug(f"Starting to do cross validation")
     start = time.time()
     cv_scores = cross_val_score(
-        model, X, y if y is not None else X, cv=TimeSeriesSplit(n_splits=5)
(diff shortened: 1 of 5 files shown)


epa095 (commented):
I like it. But do we really want to do CV every time, or should it be customizable ?

epa095 (approved):
I accept because I like it and I trust you to fix the stuff I commented on, and squash the commits to something reasonable

epa095 (approved):

PR #135, opened 2019-03-01
Iroc data source

epa095:
Added ability to download IROC data with the DataLakeProvider. In the process I created two "new" dataproviders, NcsReader (Norwegian Continental Shelf, extracted from the old DataLakeProvider) and IrocReader. I created the baseclass `GordoAwareDataProvider` for data-providers which are aware of what tags they can download (Influx is an example of a non-aware dataprovider, so far at least). `DataLakeProvider` now creates an instance of both NcsReader and IrocReader, and then delegates the downloading to whoever claims it can handle it. 

The IROC downloader uses 50 threads by default, for no particular reason.

Both the IROC reader and NcsReader could/should(?) be renamed to providers, since they now are exactly that. But they are also kind of not intended to be used directly, but they can.

### DIFF
--- a/gordo_components/data_provider/azure_utils.py
+++ b/gordo_components/data_provider/azure_utils.py
@@ -1,0 +1,125 @@
+# -*- coding: utf-8 -*-
+import collections
+import re
+import logging
+from typing import AnyStr
+
+from azure.datalake.store import core, lib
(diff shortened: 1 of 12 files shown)


milesgranger (approved):
:100:

Issue #137, opened 2019-03-05
Make sphinx docs and push to publicness


flikka:
---
Edit (@milesgranger )
- [x] Add sphinx documentation
- [x] Auto push to internal (SIB) docs
- [ ] After open sourcing, publish to readthedocs

PR #140, opened 2019-03-06
Add Sphinx documentation

milesgranger:
Part of #137 

Also supports github pages (not sure if that is / can be kept private, until we go open)

Can run `make docs` from root and have `index.html` in `./docs/html/index.html`


epa095 (commented):
damndamndamn very nice

epa095 (approved):
LGTM if you fix up the commit messages.

If we build pdf's then we can upload them to the github release "asset" thingy from travis like [this](https://docs.travis-ci.com/user/deployment/releases/), but I propose that we make that another issue.

PR #142, opened 2019-03-06
Update all dependencies with pip-compile

epa095:
Turns out that pip-compile does not always choose the latest fitting versions unless you use the `-U` flag, which I now tried. Lets see how it works :-p Also used "-r" to clear cache.


milesgranger (approved):
:+1:

PR #148, opened 2019-03-07
Keras history: Save and add to metadata

### LINKED ISSUE #138: Save keras training history
Keras models have a history field which contains the loss/acc for all the epochs, but it does not get saved by model.save. I propose that we, in `save_to_dir` and `load_from_dir` handle the history field as well if it is present (probably only present if `fit` has been called).

flikka:
Close #138


epa095 (approved):
Lgtm besides the todo

PR #158, opened 2019-03-11
Add helper CLI commands for ArgoCD & Cluster login steps

### LINKED ISSUE #157: Make some sweet CLI helpers
Hard to remember / lookup the commands for swapping clusters and adding apps/repos... let's wrap those with some prompt style CLI helpers.

milesgranger:
Will close #157 

With this one can go through **prompted** steps to:

- Authenticate with a new cluster `gordo-components cluster login`
- Authenticate with a new argocd resource `gordo-components argocd login`
- Add a repo to ArgoCD `gordo-components argocd add repo`
- Add or update app for ArgoCD `gordo-components argocd add app`
- See available clusters available to ArgoCD `gordo-components argocd clusters`

All take optional arguments but will default to terminal prompts and examples of valid input

...did I mention the happy emojis?

### DIFF
--- a/gordo_components/cli/__init__.py
+++ b/gordo_components/cli/__init__.py
@@ -1,0 +1,1 @@
+from .cli import mgmt, gordo
--- a/gordo_components/cli/cli.py
+++ b/gordo_components/cli/cli.py
@@ -19,6 +19,7 @@ from gordo_components.data_provider.providers import (
 )
 from gordo_components.server import server
 from gordo_components import watchman
+from gordo_components.cli import mgmt
 
 import dateutil.parser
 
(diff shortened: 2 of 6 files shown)


epa095 (commented):
Nice, but does this not belong more naturally in infrastructure?

epa095 (commented):
Can click load default values from an imported file? If so we could make a file with default values, and fill in some more default values (e.g. keyvault name, which realistically will rarely change). But the more information like that we put in the more naturally it belongs in infrastructure. But I think that is where the real value of this is, making it so I dont need to remember the name of the keyvault.

PR #161, opened 2019-03-13
Generalize kwargs in models

Natalie-Caruana:


epa095 (approved):
LGTM, but is it maybe possible to write a sentence or two in the commit msg? E.g. why

PR #162, opened 2019-03-13
Update testing configuration

### LINKED ISSUE #152: Doctest examples in Readme.md
We have sexy doctest testing all the inline examples, but we also have python examples in seperate Readme.md files which are probably outdated by now. It should be [possible](https://docs.pytest.org/en/latest/doctest.html) to have pytest doctest test those files as well.

milesgranger:
Will close #152 

- Removed old READMEs in sub modules as our current sphinx docs does a better job at showing examples as well as having our example notebooks

- Will run example code snippets in all `.rst` and `.md` files

- Added `pytest.ini` to have default pytest `--addopts` 

- Added component level tests, so when one it working on `watchman` instead of running the entire test suite or temp filter, can run `python setup.py testwatchman` for example


epa095 (approved):
Like ti!

PR #163, opened 2019-03-13
Add coverage/coveralls reporting

milesgranger:
Part of #156 
Not enabled for Coveralls.io reporting, until we're open sourced. Will print out the coverage report at the end of the tests right now. We're at... **86%**!


epa095 (commented):
Not bad

epa095 (approved):

PR #164, opened 2019-03-14
Watchman failure reporting

### LINKED ISSUE #139: Expose model-building failure reasons in model-server/watchman?
Proposal for discussion: We could maybe start up the model-server even if the model-builder fails, and add an endpoint which provides the log of the failed model-building?

milesgranger:
Will close #139


epa095 (commented):
Much goodness. But I think we should say that this PR depends on https://github.com/equinor/gordo-components/issues/168 so we limit a bit the exposing of the bearer tokens. But that PR should be quite easy.

epa095 (approved):
Looks great to me, except the minor logging and the fact that it does not pass the tests ;-)
We must remember to set the project version env variable when we add it to gordo-deploy.

PR #165, opened 2019-03-20
Gordo client

### LINKED ISSUE #160: Discussionpoint: Move some/all of gordo-demo into gordo-components
What about moving/porting most or all of the "demo not even a demo" into gordo-components? The rational for this is:
- Maybe it makes sense to ship some basic client with the servers, which we try to keep compatible inside a release?
- There is some (a lot) code-duplication across the two projects.
- By using gordo-components to create the client we might discover changes to the code which makes it easier to use in general.

The rest is musings about how we go about doing it: 

Each server endpoint (metadata / healtcheck / POST pred / GET pred / download-model) can have a corresponding client-function which takes in the endpoint-base-url, optionally a request object, and other required data (like start-end date). I think failures should be propagated (so not catched). How smart should the POST/GET functions be? The POST endpoint could for example do the matching of the return data with the DateTimeIndex of the data it gets in (it can maybe even be smart enough that it does that if it gets a Pandas, but if it gets a numpy thingy then it does not). There is shared logic here today between the demo-client and the GET prediction endpoint in the server, and it would be nice if that logic was extracted into a separate function which just got the input data and output data of the model. 

Another (wild) idea is that the client can have the possibility to download the model and then do the predictions locally (but not doing it locally should still be possible)

I think we should merge the "from-db" logic of the demo into the influx dataprovider (so it can be instantiated with fancy URL), but have the main client-logic rely on the timeseriesdataset. That way one can use the client with data from the datalake instead, and we get to delete some code (which is nice)

As output from the client we should still support pushing to influx, but maybe also a dump to csv kind of logic? Since we create a dataframe this should be quite trivial.

I also imagine it being possible to create some quite "integration-like" tests using docker + influx and both the server and the client.

One thing is we need to make sure we don't get tempted to put to much logic on the client. It should still be easy/possible to write a client in another language.

Thoughts?

milesgranger:
Will close #160

### DIFF
--- a/gordo_components/cli.py
+++ b/gordo_components/cli.py
@@ -189,7 +189,16 @@ def run_server_cli(
 )
 @click.option("--port", type=int, help="The port to run the server on.", default=5555)
 @click.option("--debug", type=bool, help="Run in debug mode", default=False)
-def run_watchman_cli(project_name, project_version, target_names, host, port, debug):
+@click.option(
+    "--namespace",
+    type=str,
+    help="Namespace watchman should make requests in",
+    default="ambassador",
+    envvar="NAMESPACE",
+)
+def run_watchman_cli(
+    project_name, project_version, target_names, host, port, debug, namespace
+):
     """
     Start the Gordo Watchman server for this project. Which is responsible
     for dynamically comparing expected URLs derived from a project config fle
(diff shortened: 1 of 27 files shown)


epa095 (commented):
A lot of goodness. But before we merge I think we should take a  deep look and see if there is more code/logic duplication between the client and the server which can be extracted out in common methods.

epa095 (approved):
Tiny comment, otherwise good!

PR #172, opened 2019-03-25
Set default azure log level to INFO

### LINKED ISSUE #168: Limit logging from azure python SDK
Currently we log a bunch of stuff from the azure SDK, including the bearer key. We should probably filter that away.

epa095:
Introduce new environment variable AZURE_DATALAKE_LOG_LEVEL which
sets the log level for the azure datalake, and default it to INFO.

This closes #168


milesgranger (approved):
Nice and clean change! :1st_place_medal:

PR #183, opened 2019-04-01
Add support for NCS tags without asset prefix

### LINKED ISSUE #175: Add support for NCS tags without field/assed prefix
Some tags (assets) don't seem to include asset specific information in the front of the tag. Example is the tag DQ-PT-T-B30L%2FMeas1%2FPRIM from Statfjord, which is actually located in /raw/corporate/IMS Statfjord/sensordata/​1101-SFB

See #174 

One possible suggestion is to do something like:

```machines:
  - name: machine-from-the-stoneage
    dataset:
      prefixpath: "/raw/corporate/IMS Statfjord/sensordata/​1101-SFB"
      tags: 
      - DQ-PT-T-B30L%2FMeas1%2FPRIM
      - TQ-PT-T-B30L%2FMeas1%2FPRIM
      - VQ-PT-T-B30L%2FMeas1%2FPRIM
      train_start_date: 2014-07-01T00:10:00+00:00
      train_end_date: 2014-11-01T00:00:00+00:00

flikka:
Close #175 

This suggestion changes little in the yaml input, but looks for a new "asset" key. Internally within the gordo_components, changes are bigger. A tag is now not longer a string, but a namedtuple called SensorTag. 

It is serialized (in the model dumping) as a dictionary, and the readers of the metadata (eg. watchman) must reconstruct the SensorTag namedtuple.


milesgranger (commented):
Sorry for the nit-picking. :slightly_frowning_face: It looks pretty good to me

epa095 (commented):
Looking very good mr Flikka.

epa095 (approved):
LOOOVE IT!!

PR #184, opened 2019-04-02
Model caching

### LINKED ISSUE #179: Discuss: Cache built models?
Currently any change in the config file causes a full rebuild off all models, but this does not need to be the case. We already persist the models after they are built, so it should be quite easy to avoid rebuilding. 

One proposal, which I call the `easy` alternative, is to simply cache the full description of the machine (including metadata and all) and use that as the cache key, and then in some folder under the project-folder on the shared drive, store the mapping from the generated cache keys to the folder where the corresponding model is already dumped. It is up to the model builder to generate the cache key, and load the existing model if it already exists. Model server does not need to know if it comes from the cache or was just built. 

We should probably add a new keyword to the machine like `nocache` or something to force a rebuild. 

Two things one can do to fancy it up is:
- Caching across AKS clusters
- Maybe remove some stuff, e.g. the metadata stuff, from the generation of the cache key, so it is only the actually trained model which is loaded. But since we store metadata with the model we must then pass the modified metadata in some way.

I think the easy model should be implementable in a day or two, should we do it?

epa095:
Optionally register and re-use old models

By using the --model-register-dir option to the build command you specify a
location for the model registry. This is not where the models are stored, but
the location of the registry. The registry contains the mapping from models to
their actual location.

This closes #179

### DIFF
--- a/gordo_components/builder/build_model.py
+++ b/gordo_components/builder/build_model.py
@@ -1,20 +1,22 @@
 # -*- coding: utf-8 -*-
-
-import os
-import logging
 import datetime
+import hashlib
+import json
+import logging
+import os
 import time
+from pathlib import Path
+from typing import Union, Optional
 
-from typing import Union
 from sklearn.base import BaseEstimator
-from sklearn.pipeline import Pipeline
 from sklearn.model_selection import cross_val_score, TimeSeriesSplit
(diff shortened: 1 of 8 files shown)


milesgranger (approved):
A majestic PR indeed! :tophat:

PR #187, opened 2019-04-04
Fix unstable testing with pinned test requirements

### LINKED ISSUE #186: Fix unstable tests
Test requirement pinning can change beneath us because we use flexible test dependencies in some cases.

Recently a formally passing commit may fail if rebuilt on Travis because of flexibility in test requirements.

milesgranger:
Will close #186 

This PR adds `test_requirements.in` where we can specify general requirements (with flexibility) and compile that to hard pinned requirements in `test_requirements.txt` which Travis will use. Added the same for install `requirements` for uniformity with test requirements.


epa095 (commented):
Mostly goodie

epa095 (approved):

PR #189, opened 2019-04-08
Fix external missing swagger

### LINKED ISSUE #177: Swagger for model not shown when using DNS/public URL
So: `https://gordotest.omnia-aurora-equinor.net/gordo/v0/grane-ioc/synthetic-lube-oil/` just shows "No API", but if you do `kubectl port-forward svc/ambassador 8889:80` and use `localhost:8889/gordo/v0/grane-ioc/synthetic-lube-oil/` you get the Swagger. For the public URL, you can add `/swagger.json`, and then it shows. 

Could be something with the header info we use in the server that is wrong/unexpected.

milesgranger:
Will close #177 

Default behavior of flask_restplus is to set the SwaggerUiBundle.url to the full url, in which case when using port forwarding an address like this gets rendered and works find from the browser:

image
---
By overriding `flask_restplus.Api.specs_url` to return the relative URL, it fixes this problem by rendering the following:

image

### DIFF
--- a/gordo_components/server/server.py
+++ b/gordo_components/server/server.py
@@ -12,8 +12,8 @@ from datetime import datetime
 import numpy as np
 import pandas as pd
 
-from flask import Flask, request, send_file, g
-from flask_restplus import Resource, fields, Api
+from flask import Flask, request, send_file, g, url_for
+from flask_restplus import Resource, fields, Api as BaseApi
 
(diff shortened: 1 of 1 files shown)


epa095 (approved):
Looks reasonable to me. Do we have some swagger in watchman as well which should have similar fix?

PR #190, opened 2019-04-08
Refactor ML Server & model input decoupling

milesgranger:
Part of #195 

Work on separating autoencoder models from having to keep track of their input as well as output.

The server can be capable of adding `output`, `transformed-model-input`, and if dealing with an auto-encoder add `inverse-transformed-output` as well as having different views into the model it's serving:

- Base prediction view:
    - Both `GET` and `POST` of the `/prediction` endpoint returns the following:

```json
{"data": [
    {"inverse-transformed-model-output": [0.21084319055080414,
                                                           -0.05984366685152054,
                                                            0.3098087012767792,
                                                           -0.23809830844402313],
    "model-output": [0.21084319055080414,
                            -0.05984366685152054,
                            0.3098087012767792,
                            -0.23809830844402313],
    "original-input": [0.03731559272851026,
                           0.56919744099208,
                           0.39446942973169974,
                           0.6009476926566822],
    "transformed-model-input": [0.03731559272851026,
                                             0.56919744099208,
                                             0.39446942973169974,
                                             0.6009476926566822]
    },
],
 "status-code": 200,
 "tags": [
               {"asset": null, "name": "tag-0"},
               {"asset": null, "name": "tag-1"},
               {"asset": null, "name": "tag-2"},
               {"asset": null, "name": "tag-3"}
],
"time-seconds": "0.1053"}
```

- Anomaly prediction view:
    - Both `GET` and `POST` of the `/anomaly/prediction` endpoint returns the same "anomaly format" data, in our case, a superset of the base view will be the same as above, but with additional keys in "error-untransformed", "error-transformed", "total-transformed-error", and "total-untransformed-error"
```


epa095 (commented):
I like the idea. Some small issues. To dumb today to read it properly, will give another stab tomorrow.

epa095 (commented):
I am a bit confused, are all the keys you show in the examples guaranteed to exists? If not, which one are optional?

milesgranger:
Moved to #288

PR #192, opened 2019-04-09
Load model and metadata before server starts

### LINKED ISSUE #191: Server: Have healtcheck not try to load the model
Currently the healtcheck endpoint is a subclass of `base`, so doing a call to healcheck will trigger a model load. Healtcheck should not do stuff, just check if things are there. So I propose healtcheck to check if `MODEL` is non-None, and return OK if, otherwise not return OK. The `metadata` endpoint can return stuff, so we should split them. 

https://github.com/equinor/gordo-components/blob/9aba3e384b1382e3d59091c4da039bf21339226d/gordo_components/server/server.py#L324

milesgranger:
Will close #191

### DIFF
--- a/gordo_components/server/server.py
+++ b/gordo_components/server/server.py
@@ -12,8 +12,9 @@ from datetime import datetime
 import numpy as np
 import pandas as pd
 
-from flask import Flask, request, send_file, g, url_for
+from flask import Flask, request, send_file, g, url_for, current_app
 from flask_restplus import Resource, fields, Api as BaseApi
+from sklearn.base import BaseEstimator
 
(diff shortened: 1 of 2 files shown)


epa095 (approved):
LGTM, assuming you have tested it a bit ;-)

epa095 (changes requested):

epa095 (commented):
Is the gevent needed to solve the most pressing issue, the loading model before server starts, or is it mostly a nice thing that came up at the same time?

epa095 (approved):
LGTM, except fixing out-of-sync docstring.

PR #193, opened 2019-04-10
ABANDONED: Move workflow logic from infrastructure into components

### LINKED ISSUE #181: Discussion: Move argo workflow & generator into components
Relates to making us useful for open source.

The workflow generator can live in components and require users, ie. gordo-infrastructure, to define a definitions template of sorts which the workflow generator template extends, and uses the defined `models`, `datasets` etc. etc. to generate the workflow. 

This way we can also move the config elements and associated elements into components and enable us to create endpoints which allow for validating config files or...?? :door: :open_hands: :rocket:

milesgranger:
Will close #181 
Move the workflow generator and related logic from gordo-infrastructure into a `workflow` component.


milesgranger:
Moved to #290

PR #196, opened 2019-04-11
Rename .load_dataframes to .load_series

### LINKED ISSUE #136: Change load_dataframes to return series?
It is after all just a single index and a column it returns

milesgranger:
Will close #136 
Rather than treating them as single column dataframes


epa095 (approved):
LGTM

PR #204, opened 2019-04-17
Make client batched POST forward correct index

epa095:
Previosly it forwarded the last indexes of the prediction for all
batches, resulting in loss of most predictions.


flikka (changes requested):

milesgranger (approved):
Looks good to me, nice catch with checking the chunk start/stop. Good news is that the client tests will get a lot cleaner with #213 I think.

flikka (approved):

PR #208, opened 2019-04-23
Wait for influx to start up properly

epa095:
Before we did 2 or 3 seconds of sleep (practices differed), now we do up to 30
seconds of waiting while pinging influx /ping endpoint and continue when it is
done starting.

Also handled failing starting of influx better by cleaning up any started
influx instead of leaving them running.


milesgranger (approved):
:1st_place_medal: Awesome-sauce! Only one pedantic comment that you can completely ignore if you'd like.

Issue #209, opened 2019-04-23
Refactor tests epic-issue


epa095:
Superissue to discuss some of the things we should/could do to the tests. We should make concrete issues as we agree on stuff, this is proposals:

- Use pytest.fixtures where appropriate (#180)
- Convert completely away from unittests by no longer inheritting from it.
   - Convert from `self.assertEquals` to `assert`
- Improve runtime by not running tests over all models and kinds, but only all models and one kind. 
- Also convert the for-loops in the above mentioned tests into using pytest parametrize, then it will print to stdout for each of the models.
- Change file-system layout so it uses folders and mirrors the structure of the actual code. This to make it easier to find where tests for "your" code belongs, but also makes it easier to run different subsets of the tests (since you can choose them by folder/files)

PR #211, opened 2019-04-24
Refactor model tests

milesgranger:
Move test files into the appropriate test modules which match their code base modules. ie. `test_dataset.py` -> `test_dataset/test_dataset.py` and all other component level tests.

Remove looping over registered factories inside each test and expand into pytest parametrization. Should help with long pauses in running individual tests, now each combination of `Model` and `kind` will run as its own test. 

Part of #209


epa095 (approved):
LGTM

PR #216, opened 2019-04-24
Add forward-resampled-sensors option to client predict

### LINKED ISSUE #214: Prediction client: Forward sensor data to influx
Add a new toggle (default to false) which when set forwards the resampled sensordata to influx. I guess it only makes sense when we are already forwarding predictions to influx.

Here is some code I wrote that does something like it
```python
... 
    influx_df_client = DataFrameClient(host=influx_name, database=database_name)
    for tag_name, tag_data in explode_df(X).items():
        if not influx_df_client.write_points(
            tag_data, measurement="resampled", batch_size=10000, tags={"tag": tag_name}
        ):
            raise ValueError(f"Failed at writing tag data for {tag_name} to Influx")
        else:
            logger.debug(f"Wrote resampled tag data for {tag_name} to Influx")
    logger.debug("Done writing resamples sensor values to influx")


def explode_df(df, field_name="value"):
    ret = dict()
    for col in df.columns:
        ret[col] = df[[col]].rename(columns={col: field_name}, inplace=False)
    return ret
```

epa095:
Forwards the resampled sensor-data to influx.

This closes #214

### DIFF
--- a/gordo_components/cli/client.py
+++ b/gordo_components/cli/client.py
@@ -69,6 +69,12 @@ def client(ctx: click.Context, *args, **kwargs):
     is_flag=True,
     default=False,
 )
+@click.option(
+    "--forward-resampled-sensors",
+    help="forward the resampled sensor values",
+    is_flag=True,
+    default=False,
+)
 @click.pass_context
 def predict(
     ctx: click.Context,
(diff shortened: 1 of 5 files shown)


milesgranger (commented):
I'm not opposed to this approach; I too would have had it inside the `forwarder`, but rather changing the signature of the callable to _optionally_ accept the raw data associated with the predictions it was given.

Then as you have here, based on the flag, either pass those or not to the forwarder, if it exists.  And of course, if no data provider was supplied, then it couldn't pass the raw data anyway.

milesgranger (approved):
Only a small detail, but LGTM!

PR #219, opened 2019-04-26
Final test optimization pass through

### LINKED ISSUE #180: Refactor unittests
Probably better to see where `pytest.fixture`s can be used, both in terms of model caching as well as `pytest.yield_fixture` to share `watchman` deployments and `influxdb` usages, where appropriate.

milesgranger:
Part of #209 and will close #180


epa095 (approved):
A lot of niceness here. For some reason it seem to run a bit slower (!) on travis than the latest master, but it is worth it. But a bit strange.

PR #220, opened 2019-04-26
Client add dockerfile

### LINKED ISSUE #217: Add Dockerfile-client

epa095:
Adds a dockerfile for the client, and adds it to the makefile. Why? I need it for grafana, and I guess it is kind of nice to be able to use it from docker without installing it at all. 

This closes #217


milesgranger (approved):
Aside from the minor comment, LGTM

PR #231, opened 2019-05-03
Change running of black tests to use os.system, update black

flikka:
The CLI runner together with black seems to be a bit unstable, suggest using os.system for a maybe more predictable running.


epa095 (commented):
I like except question

epa095 (approved):

PR #235, opened 2019-05-03
Non-zero exit code for client on errors

### LINKED ISSUE #223: Client: Should exit with non-0 exit code on failure
If the client fails POST because the upstream server fails then the client still exits with return code 0, indicating sucsess. I think if a single fails it should exit with 1, or some other exit code than 0 so it can be detected pragmatically that it failed.

milesgranger:
Close #223

### DIFF
--- a/gordo_components/cli/client.py
+++ b/gordo_components/cli/client.py
@@ -2,6 +2,7 @@
 
 import os
 import typing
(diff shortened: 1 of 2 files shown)


epa095 (approved):
LGTM

PR #240, opened 2019-05-07
Make client optionally ignore unhealthy endpoints

### LINKED ISSUE #239: Client: Handle failure of non-relevant servers
If a single server in a project fails (is non-healthy from watchmans perspective), then the client will fail, even with an explicit target to some other server set. This is because it tries to built `EndpointMetadata` for all the endpoints, and in doing so it tries to reference the metadata, which is not available since the server is down. It will crash at:

https://github.com/equinor/gordo-components/blob/c635a44d1905d9fce2bb5dfad3ab3677b343ded0/gordo_components/client/client.py#L125

One alternative is to just have `tag_list` and `resolution` be None in that case, and then check for healthiness where the endpoint is actually used. 

It seems clear that if you ask for predictions for a single target A then that should work if A is up, even though some other target B is down at the moment. But if you ask for all targets in a project and some is down, then what do we do? Maybe do the ones we can, and exit with 1 indicating a failure?

milesgranger:
Will close #239

### DIFF
--- a/gordo_components/cli/client.py
+++ b/gordo_components/cli/client.py
@@ -76,6 +76,13 @@ def client(ctx: click.Context, *args, **kwargs):
     is_flag=True,
     default=False,
 )
+@click.option(
+    "--ignore-unhealthy-targets",
+    help="Ignore any unhealthy targets. By default the client will raise an "
+    "error if any unhealthy endpoints are encountered.",
+    is_flag=True,
+    default=False,
+)
 @click.pass_context
 def predict(
     ctx: click.Context,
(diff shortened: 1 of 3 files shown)


epa095 (changes requested):

epa095 (commented):
We dont do tests anymore? I think if you extract the new logic from the init into a seperate function you can easily test it without having to mock anything, just provide it a list of `EndpointMetadata`-elements and check a few of the cases.

epa095 (approved):
LGTM

PR #249, opened 2019-05-09
Rename Watchman's "metadata" top-level key to "endpoint-metadata"

### LINKED ISSUE #178: "Two levels" of metadata in the the JSON returned from APIs
Example, if you look at the "endpoints" part of a project (eg. `https://gordotest.omnia-aurora-equinor.net/gordo/v0/grane-ioc/`) - each of them has a "metadata" key. In there, you have some "global metadata", and then another "metadata" key. This is a bit "un-beautiful".

Example:

``` 
 "endpoint": "/gordo/v0/grane-ioc/synthetic-lube-oil/",
  "healthy": true,
  "metadata": {
    "env": {
      "MODEL_LOCATION": "/gordo/models/grane-ioc/1553603389.1011672/synthetic-lube-oil"
    },
    "gordo-server-version": "0.13.0",
    "metadata": {
      "dataset": {
        "filter": "",
        "resolution": "10T",
        "tag_list": [
          "GRA-LIT -23-0645.PV",
          "GRA-NRA -23-0413.PV",
          "GRA-PDIT-23-0646.PV",

milesgranger:
Will close #178

### DIFF
--- a/gordo_components/client/client.py
+++ b/gordo_components/client/client.py
@@ -180,15 +180,17 @@ class Client:
 
         return [
             EndpointMetadata(
-                target_name=data["metadata"]["metadata"]["user-defined"][
+                target_name=data["endpoint-metadata"]["metadata"]["user-defined"][
                     "machine-name"
                 ],
                 healthy=data["healthy"],
                 endpoint=f'{self.base_url}{data["endpoint"].rstrip("/")}',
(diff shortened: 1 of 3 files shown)


epa095 (approved):
LGTM. But what about the other metadata, are we renaming that as well? If so I suggest we do it as fast as possible so we break as much as possible in a single go.

PR #250, opened 2019-05-09
Add ability to have template variables in model config

### LINKED ISSUE #241: Template the model config

epa095:
This PR introduces a new option to model building  `--model-parameter`:
> Key-Value pair for a model parameter and its value, may use this option 
   multiple times. Separate key,valye by a comma. ie: --model-parameter key,val
   --model-parameter some_key,some_value

It also allows you to specify the model with jinja-holes in them. An example of a legal model is:
 ```yaml        {
         "sklearn.decomposition.pca.PCA":
          {
            "svd_solver": "{{svd_solver}}",
            "n_components": {{n_components}}
          }
        }
```
And this can then be built with e.g. ` --model-parameter svc_solver,auto --model-parameter n_components,0.5`.

I also found that by replacing Keras with PCA in the CLI tests we got a speedup from 20s to 0.2s, so  I added that as a separate commit. That is related to #253. 

This closes #241 .

### DIFF
--- a/gordo_components/cli/cli.py
+++ b/gordo_components/cli/cli.py
@@ -9,6 +9,7 @@ import ast
 import os
 from ast import literal_eval
 
+import jinja2
 import yaml
 import click
 from gordo_components.builder.build_model import provide_saved_model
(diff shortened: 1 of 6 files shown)


milesgranger (approved):
:birthday: LGTM :rocket: 

Aside from my minor comment about logging, but to your judgement.

PR #254, opened 2019-05-10
Remove module level import of _get_dataset

epa095:
It caused import loop error


milesgranger (approved):
Nice! When we made it private, I should have taken it out from the `__init__` in the first place, thanks!

PR #256, opened 2019-05-10
Print vc scores

### LINKED ISSUE #243: Modelbuilder: Output scoring to stdout (option/toggle)

epa095:
Prints out the cross validation scores on a katlib-readable format to stdout.
If we ask for CV printing but the loaded model from cache does not have that metadata stored then we delete it from the cache and rebuild the model. 

If/when we support doing CV with several choosen scoring methods this logic can be extended to rebuild the model in case at least one of them are not present.

This closes #243 .


flikka (approved):
LGBT!

PR #261, opened 2019-05-14
Support caching for Watchman

### LINKED ISSUE #260: Cache on Watchman
It's main endpoint requires it to ping kubernetes to inquire about the model status. When projects get large with many models, and then many clients connecting to the same endpoint to fill model predictions, it can be overly expensive.

milesgranger:
Will close #260

### DIFF
--- a/gordo_components/watchman/server.py
+++ b/gordo_components/watchman/server.py
@@ -4,9 +4,10 @@ import os
 import logging
 from collections import namedtuple
 from concurrent.futures import ThreadPoolExecutor
-from typing import Iterable, Optional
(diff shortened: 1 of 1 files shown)


epa095 (commented):
I think you need to actually anotate some of the getters?

Using `@cache.memoize(timeout=5)`

epa095 (commented):
Have you tested it? Particularly, what happends if you do one request with `logs` and then another one without `logs` right (< 5s) after? Does it return logs or not? I am wondering if the annotation is smart enoug to know about the `request.args` object.

epa095 (approved):

PR #263, opened 2019-05-16
Convert several test Keras -> PCA for speedup

### LINKED ISSUE #253: Speedup tests by removing keras from them
It takes a long time to make Keras NN models, and we might have places where we use KerasAutoEncoder in our tests when we could really just have used a fast scikit learn model instead, since it is not actually the Keras stuff we want to test.

epa095:
This closes #253


flikka (approved):
LOVELY. 40 to 4 seconds on my computerator.

PR #267, opened 2019-05-20
Fix filter rows to use back ticks in yaml

### LINKED ISSUE #251: Filter rows does not work from yaml
The problem is that to reference column names (i.e. tags) you need to use single ticks `'`, but that generates an invalid argo workflow. One proposed sollution is to instead use backticks, and then just replace them with single ticks inside row_filter.py. The advantage of this is that in pandas 25 the `query` method will use exactly backticks to support column names with spaces in them, so it should make row filters forward compatible.

Natalie-Caruana:
Replace quotations with back ticks to select columns for row filtering.

This will close #251.

### DIFF
--- a/gordo_components/dataset/filter_rows.py
+++ b/gordo_components/dataset/filter_rows.py
@@ -51,8 +51,10 @@ def pandas_filter_rows(df, filter_str: str):
     filter_str: str
       String representing the filter. Can be a boolean combination of conditions,
       where conditions are comparisons of column names and either other columns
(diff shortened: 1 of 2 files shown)


epa095 (commented):
Minor comment, otherwise looks good

flikka (changes requested):
Looks good, but maybe we could allow both single, double and backtics around the column names.

flikka (approved):

epa095 (approved):
Lovely

PR #269, opened 2019-05-21
Add resampling_endpoint up to which data will be forward filled

### LINKED ISSUE #266: dataset resample: Add a point at the to_ts
Today we resample each of the timeseries individually, and each timeseries is resampled from the `start_ts` to the last timestamp they have. That means that if some timeseries lacks e.g. one day of data before `end_ts`, the resampled timeseries will also lack one day. Then we do a **inner** join between the timeseries, so the resulting index is the intersection of all the timeseries, i.e. it becomes equal to the shortes timeseries. I propose that we forward fill the value at the last timestamp `t` such that `start_ts <= t <= end_ts` to `end_ts`. Then all the resampled timeseries will be equally long. One alternative is to do an outer join instead, and the forward fill after joining. But that just makes the resulting dataset larger than today, we can still not guarantee anything about how large it actually becomes.

When we one day support something else than forward filling (e.g. linear interpolation) it must be handled somewhat differently. We then need to look for one point after (or equal) to `end_ts` and use it to interpolate the value at `end_ts`, but we handle that in the future.

Natalie-Caruana:
This will close #266.

### DIFF
--- a/gordo_components/client/client.py
+++ b/gordo_components/client/client.py
@@ -607,11 +607,10 @@ class Client:
         pandas.core.DataFrame
             Dataframe of required tags and index reflecting the datetime point
         """
-        freq = pd.tseries.frequencies.to_offset(endpoint.resolution)
 
         dataset = TimeSeriesDataset(  # type: ignore
             data_provider=self.data_provider,
-            from_ts=start - freq.delta,
+            from_ts=start,
             to_ts=end,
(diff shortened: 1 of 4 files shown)


epa095 (commented):
Looks good in generall

epa095 (approved):
LGTM, although I propose you squash the two commits together.

PR #279, opened 2019-05-24
Watchman set endpoint healthy via metadata response

### LINKED ISSUE #272: Watchman: Set healthy based on get metadata
Today we do two requests, one to healtcheck and one to metadata. But if the server becomes unhealthy in between we will say it is healthy but respond with an empty metadata. There are two solutions:
1: Set the `healthy` field as the *and* of the healtcheck and the get(metadata)
2: Just to a get(metadata) and set healthy=True if it worked.

I vote for 2 since that gives less load on our servers. 

https://github.com/equinor/gordo-components/blob/5f1b711397c56b8fd4d2bd76fd9db61975b8588e/gordo_components/watchman/server.py#L84

milesgranger:
Will close #272


epa095 (approved):
Lovely rita, lovely

Issue #195, opened 2019-04-11
Add "anomaly" concept in gordo?


flikka:
This must be discussed, but there appears to be work going on wrt providing "anomaly" signals that transcends the "reconstruction error" and sensor errors. This will probably ongoing development for a long time, looking at trends/weak signals, patterns etc - but we should probably introduce the concept into gordo soon. 

For example as "extra" data series that are supplied as a part of the output/results, could be several "anomaly" series in many variants and levels - for the whole machine and for sensors.

epa095:
This seems to be what miles is doing in #368 with `DiffBasedAnomalyDetector` ?

milesgranger:
@flikka can this be closed?

Seems any additional variations of anomalies have a clear way forward now, following the pattern of `DiffBasedAnomalyDetector`. While #449 is something that is related but not complete?

flikka:
I agree! This is fine done, and the “flagging” stuff is in #449

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #288, opened 2019-05-28
Refactor ML Server & model input decoupling

milesgranger:
Part of #195 

Moved from https://github.com/equinor/gordo-components/pull/190

Work on separating autoencoder models from having to keep track of their input as well as output.

The server can be capable of adding `transformed-model-input`, `model-output`, `original-input` and if output shape matches input shape, and all steps in pipeline support `inverse_transform`, will add `inverse-transformed-model-output`

- Base prediction view:
    - Both `GET` and `POST` of the `/prediction` endpoint returns the following:

```json
{"data": [
    {"model-output": [0.21084319055080414,
                            -0.05984366685152054,
                            0.3098087012767792,
                            -0.23809830844402313],
    "original-input": [0.03731559272851026,
                           0.56919744099208,
                           0.39446942973169974,
                           0.6009476926566822],
    "transformed-model-input": [0.03731559272851026,
                                             0.56919744099208,
                                             0.39446942973169974,
                                             0.6009476926566822]
    },
],
 "status-code": 200,
 "tags": [
               {"asset": null, "name": "tag-0"},
               {"asset": null, "name": "tag-1"},
               {"asset": null, "name": "tag-2"},
               {"asset": null, "name": "tag-3"}
],
"time-seconds": "0.1053"}
```

- Anomaly prediction view:
    - Both `GET` and `POST` of the `/anomaly/prediction` endpoint returns the same "anomaly format" data, in our case, a superset of the base view will be the same as above, but with additional keys in `error-untransformed`, `error-transformed`, `total-transformed-error` and `total-untransformed-error`
```

### CODE REVIEW

**Comment by milesgranger** [comment]
@epa095 @flikka Review requested 6 days ago. :watch:

**Comment by milesgranger** [comment]
@jgaz Regarding the example payload, do you approve? Keeping in mind, the `/anomaly/prediction` would additionally return keys that you use now, like `total-transformed-error`, `error-transformed`, etc.

**Comment by jgaz** [comment]
> @jgaz Regarding the example payload, do you approve? Keeping in mind, the `/anomaly/prediction` would additionally return keys that you use now, like `total-transformed-error`, `error-transformed`, etc.

Yes, no problem at all, just let me know when this hits production to check the client.

**Review by flikka — COMMENTED** [review]
The use of "lvl" for "level" seems a bit over the top, as long as we write all kinds of other words out?

File: tests/gordo_components/server/test_anomaly_view.py [hunk 1]
```diff
@@ -0,0 +1,119 @@
...
+        {
+            "X": np.random.random(size=(10, len(tu.SENSORS_STR_LIST))).tolist()
+        },  # Nested records
+        {
+            "X": np.random.random(size=len(tu.SENSORS_STR_LIST)).tolist()
+        },  # Single records 1d
+        None,  # No data, use GET
+    ],
+)
+def test_anomaly_prediction_endpoint(
```
flikka, line 21: Generally, I feel this:
* We no longer support "non-pipeline" models. We should explicitly show this in tests? With something failing?
* Maybe we should have a pipeline with only one step? I think it is the `trained_model_directory` in conftest which is the source for many of the "end-to-end" (TM) tests.

Could we do some parametrization or maybe just explicitly have a couple of different `trained_model_directory` versions (no pipeline, will fail in the server region, due to .steps and stuff, and single step, and more steps (like today))?

flikka (changes requested):
Lots of very good improvements here, impressive amount of work, in particular if it works.

Main comment is that it seems to me that the only "end-to-end" tests (meaning from a model pipeline to an API endpoint) use a single model pipeline, with a scaler and a KerasAutoEncoder. Would it make sense to add a couple more, maybe one without a Pipeline (that will fail) and one with only the actual model (one single step)?

tests/gordo_components/server/test_anomaly_view.py
milesgranger: Check out the latest tests, and see if that doesn't suit your fancy. Thought about parametrizing the tests but I think this route it's quicker (both test timing and less complex tests) and is what the server  uses anyway to interact with the model `ModelMixin`

flikka (approved):

gordo_components/server/mixins.py
```diff
@@ -0,0 +1,104 @@
...
+
+import numpy as np
+import pandas as pd
+
+from sklearn.pipeline import Pipeline
+
+logger = logging.getLogger(__name__)
+
+
+class ModelMixin:
```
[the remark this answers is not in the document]
milesgranger: Very good point.. In my head while I was writing it, I expected to use it in both views, but then found it was easier (and cleaner?) to use propagate the base view's result. But I agree, making these into functions is a better approach now. :100:

milesgranger:
Merging this bad-jackson after a successful run of the gauntlet by @flikka  (see attch) and a verbal approval of Erik this morning.

Between that, I just needed to add a few fixes (of which were exposed with the new test project :tada: ) to allow the `make_base_dataframe` which would offset the index (if applicable) and the input data to match the model output if it was different. Along with a silly catching of a `ValueError` in the `inverse_transform` model io as we did with the `transform`

Gauntlet:
image

Dashboard
image

PR #291, opened 2019-05-28
Add name as a mandatory input to model building, and insert into metadata at higher level

### LINKED ISSUE #238: Move machine-name higher up in metadata
Currently it resides in `["metadata"]["metadata"]["user-defined"]["machine-name"]`, but that is a bit weird. It should reside higher up, next to the `dataset` key.

flikka:
The output of this is that the metadata structure gets a "name": "machine-whateveritis-name" next to the "dataset" etc. So the endpoints will change. The config is not changed.

Require https://github.com/equinor/gordo-infrastructure/pull/274
Close #238 

Old structure: endpoint-metadata -> metadata-> user-defined -> machine-name
New structure: endpoint-metadata -> metadata -> name

Old example:
Screenshot 2019-05-29 at 15 17 08

New example:
Screenshot 2019-05-29 at 15 18 35

### DIFF
--- a/examples/Gordo-Workflow-High-Level.ipynb
+++ b/examples/Gordo-Workflow-High-Level.ipynb
@@ -120,6 +120,7 @@
    "outputs": [],
    "source": [
     "pipe, metadata = build_model(\n",
+    "    name=config['model'],\n",
     "    model_config=model_config, \n",
     "    data_config=data_config,\n",
     "    metadata=machine_config['metadata']\n",
--- a/gordo_components/builder/build_model.py
+++ b/gordo_components/builder/build_model.py
@@ -167,7 +167,7 @@ def _get_final_gordo_base_step(model: BaseEstimator):
 
 
 def calculate_model_key(
-    model_config: dict, data_config: dict, metadata: Optional[dict] = None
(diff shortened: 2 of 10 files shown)


flikka:
New PR due to fork-trouble when going over to open source gordo

@epa095 pointed out that the caching will not take the name into account and must be fixed

flikka:
Old PR: https://github.com/equinor/gordo-components/pull/282

flikka:
Tested on auroraweek22 together with https://github.com/equinor/gordo-infrastructure/pull/274 , and it (seems to work), including the client and the Grafana dashboards

milesgranger (commented):
As commented in the infrastructure PR, while I think it looks fine, if not positively excellent. Maybe we actually want to not have this as a `model name`? or maybe we do.. depending on our understanding of what a single definition is (currently our understanding as a "Machine").. but maybe, as I said, should be `GordoModel` which is comprised of things like dataset, metadata, the actual ML model, etc.  I'm not sure if it should be changed in this PR or later.

Also, I think adding an example of how the JSON response(s) will change as a result of this PR.

epa095 (approved):
LGTM

PR #295, opened 2019-05-28
Client should also retry on BadRequest errors

### LINKED ISSUE #294: Client: Retry on any error.
The server is weird, and can return 4xx even when it is not really our fault. Change the client so it retries on every error, not only the 500.

milesgranger:
Will close #294

### DIFF
--- a/gordo_components/client/client.py
+++ b/gordo_components/client/client.py
@@ -384,7 +384,7 @@ class Client:
                 )
 
             # If it was an IO or TimeoutError, we can retry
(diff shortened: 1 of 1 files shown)


epa095 (commented):
LGTM

milesgranger:
Press :green_apple: :radio_button: please. :tophat:

epa095 (approved):

epa095:
For the future : I won't be angry if you just use your admin privileges to merge it after I day lgtm but press the wrong button!

PR #296, opened 2019-05-28
WIP: Client fetch metadata from target directly

epa095:
This PR makes the client contact the server directly for metadata instead of watchman, which helps when the projects are big and watchman are slow. But we should still fix watchman, so maybe not merge this? I dont know.


epa095:
I think maybe even after we fix watchman this can be useful, since we will make watchman eventually consistent. And in the workflow the client is ran directly after the server is up, but then maybe watchman has not picked it up yet? If course one can always put a tiny sleep before the client, but then we must have a bound on how long to sleep

epa095:
Nha, I think the new watchman in #330 solves this better

PR #297, opened 2019-05-29
Support eventually consistent endpoint metadata in Watchman

### LINKED ISSUE #281: Discussion: Design watchman for scale
Today a get-request to watchman is either answered from the 5-second cache, or it triggers a full scan of all the `n` machines, with `n` sequential HTTP-get requests. This is unlikely to scale to instances where `n` is large. Thus this issue to discuss what to do.

Today watchman servers 3 purposes:
1. Lists all the machines/endpoints of a project,
2. Gives the status of all of them, and
3. Gives the metadata for the ones which are up. 

Note that it can actually do 1 without any network traffic at all, it knows the list of machines at startup. It does 2 and 3 combined today, by fetching the metadata for each machine. But it can do point 2 much more efficiently, by just asking kubernetes. The server-deployments have `healtcheck` defined, and kubernetes checks it every 10 seconds. So we can simply ask kubernetes for all deployments matching the same project / project version, and we will get how many live pods each of them have. This should be significantly faster than scanning them all, and utilizes the fact that kubernetes is already scanning them for us. 

Regarding point 3, notice that the metadata is pretty much immutable, as long as the project version stays the same. So if we manage to get the metadata for a server, and it has the same version as us, we can actually cache it indefinitely. But the cache must be populated somehow, and if the first request fires off e.g. 600 request sequentially that will take a long time. One option is to do those requests in parallel, but that assumes that ambassador/envoy/watchman is able to handle that many parallell request. The other alternative is to have a backgroundthread which, from the beginning, tries continually to fetch a metadata until it manages, and then it puts it into a map. And then when we get a GET request for metadata we serve whatever we have in that map (so a GET request never directly initializes a scan)

milesgranger:
Will close #281 

Uses [APScheduler](https://apscheduler.readthedocs.io/en/latest/index.html) to update `EndpointStatuses` every 5 minutes. `EndpointStatuses` holds a list of the endpoint statuses as `dict`s which are locked when read with `.statuses()` or updated with `.update()` 

Will get "fresh" statuses when requesting `?logs`/`?logs=20` in the URL to get logs of the model server and builder.

### DIFF
--- a/gordo_components/watchman/endpoints_status.py
+++ b/gordo_components/watchman/endpoints_status.py
@@ -1,0 +1,205 @@
+# -*- coding: utf-8 -*-
+
+import threading
+from typing import List, Optional
+from collections import namedtuple
+from concurrent.futures import ThreadPoolExecutor
+
+import requests
+from flask import current_app
+
+from gordo_components.watchman.gordo_k8s_interface import (
+    Service,
+    list_model_builders,
+    ModelBuilderPod,
+)
+
+
+EndpointStatus = namedtuple(
+    "EndpointStatus",
+    [
+        "endpoint",
+        "target",
+        "metadata",
+        "healthy",
+        "model_builder_status",
+        "model_server_status",
+    ],
+)
+ModelBuilderStatus = namedtuple("ModelBuilderStatus", "status logs")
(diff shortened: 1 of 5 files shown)


milesgranger:
@epa095 has a level 3 version of this.

PR #299, opened 2019-05-31
Replace generate_window with keras functionality

### LINKED ISSUE #293: Understand LSTM predict performance
It is much much slower than feed_forward, and it seems quite slow compared to the time it takes to fit it as well. Have we done something weird, or is it just slow?

epa095:
- Replaces generate_window with keras functionality
- Sets 10000 as default batch_size on predict. Batch_size on predict is *not* the same as the on for training, and is purely a performance trade-of (memory vs speed). Thus it is not customizable (at least not yet) from yaml. Did some rough benchmarks with different loookback-windows (144 / 72 / 36) and n_features (3,6) on transforming 100000 datapoints, and using a batch-size around 10000 was the fastest in general. 

- The previous implementation was equivalent with having batch_size=1 on predict, and that is very much slower than having it higher. Because of this we get this performance boost (measured on my laptop):
`model.transform(np.random.rand(1000,3))`
Old:
1min 14s  ± 1.7s per loop
new: 
132 ms ± 47.1 ms per loop
- Currently there is a small change in that in `fit` the first sample will be used twice. Can change that by removing it, but do we care? (see TODO)
- Made it such that all calls to `fit` calls `super().fit`, which generates a completely fresh model. This ensures that we are compatible with sklearn, especially for cross-validation, which of course expects a fresh model for each of the folds.
- There is also the change that we fit on all the data, and not only on the first 1/32th of the data, which we did before (bug)
- Removed the ability to output without a `y`, since it was not used.
- The new method might use a bit more memory. The reason is that we have to insert upto several rows at the start and end of the dataset for `TimeseriesGenerator` to generate the lookahead, and that probably causes numpy to copy the whole matrix. My current feel is that we should be OK with this, and rather implement something smarter if we need it later.

This closes #293

### DIFF
--- a/gordo_components/model/models.py
+++ b/gordo_components/model/models.py
@@ -2,17 +2,16 @@
 
 import logging
 import json
-import math
 
 from typing import Union, Callable, Dict, Any, Optional
 from os import path
 from contextlib import contextmanager
 import pickle
-import abc
 from abc import ABCMeta
 
 import keras.models
 import keras.backend as K
+from keras.preprocessing.sequence import pad_sequences, TimeseriesGenerator
 from keras.wrappers.scikit_learn import BaseWrapper
 from keras.models import load_model
 import numpy as np
@@ -48,7 +47,8 @@ def possible_tf_mgmt(keras_model):
     Parameters
     ----------
     keras_model: KerasBaseEstimator
-        An instance of a `KerasBaseEstimator` which can potentially have a tensorflow backend
+        An instance of a `KerasBaseEstimator` which can potentially have a tensorflow
+        backend
     """
     logger.info(f"Keras backend: {K.backend()}")
     if K.backend() == "tensorflow":
@@ -66,17 +66,19 @@ class KerasBaseEstimator(BaseWrapper, GordoBase):
         **kwargs,
     ) -> None:
         """
-        Initialized a Scikit-Learn API compatitble Keras model with a pre-registered function or a builder function
+        Initialized a Scikit-Learn API compatitble Keras model with a pre-registered
+        function or a builder function
         directly.
 
         Parameters
         ----------
         kind: Union[callable, str]
             The structure of the model to build. As designated by any registered builder
-            functions, registered with gordo_compontents.model.register.register_model_builder
-            Alternatively, one may pass a builder function directly to this argument. Such a
-            function should accept `n_features` as it's first argument, and pass any additional
-            parameters to `**kwargs`
+            functions, registered with
+            `gordo_compontents.model.register.register_model_builder`.
+            Alternatively, one may pass a builder function directly to this argument.
+            Such a function should accept `n_features` as it's first argument, and pass
+            any additional parameters to `**kwargs`
 
         kwargs: dict
             Any additional args which are passed to the factory
@@ -276,7 +278,8 @@ class KerasAutoEncoder(KerasBaseEstimator, TransformerMixin):
         The data returns from this method is double the feature length of its input.
         The first half of each sample output is the _input_ of the model, and the
         output is concatenated (axis=1) with the input. ie. If the input has 4 features,
-        the output will have 8, where the first 4 are the values which went into the model.
+        the output will have 8, where the first 4 are the values which went into the
+        model.
         """
         with possible_tf_mgmt(self):
             xhat = self.model.predict(X, **kwargs)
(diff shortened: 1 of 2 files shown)


gordo_components/model/models.py
```diff
@@ -761,7 +626,7 @@ def fit(
     ) -> "KerasLSTMForecast":
 
         """
-        This fits a one step forecast LSTM architecture using Keras fit_generator method.  For further details
+        This fits a one step forecast LSTM architecture.  For further details
         on the input/output samples fed to this method, refer to the generate_window method.
```
flikka: Update docstring (generate_window -> get_TimeseriesGenerator) - same for KerasLSTMAutoencoder.

flikka (commented):
Nicely spotted!

Need to look more at it to understand the stuff in the get_TimeseriesGenerator.

Love the compact commit message, that still captures the essence of the PR!

gordo_components/model/models.py
```diff
@@ -902,3 +776,80 @@ def score(
         out = self.transform(X)
 
         return explained_variance_score(out[:, : X.shape[1]], out[:, X.shape[1] :])
+
+
+def get_TimeseriesGenerator(data, batch_size, lookback_window, lookahead):
```
milesgranger: I find `get_TimeseriesGenerator` a touch strange, in that this is basically a wrapper to instantiating `TimeseriesGenerator`. It doesn't follow standard naming convention for functions (snake case). I suggest either create a fully qualified wrapper `TimeseriesGenerator(sequence.TimeseriesGenerator)` and make this function's signature the `__init__` or classmethod of some kind, or if you prefer to have a free function for this, think the naming at least should be fixed.

milesgranger: Also, if we want to continue to type our parameters in the signature.. not sure how strict we're being with that.

tests/gordo_components/model/test_model.py
```diff
@@ -203,62 +204,35 @@ def test_lookback_window_forecast_valueerror_during_transform(self):
...
-
         lookback_window = 3
-        model = KerasLSTMForecast(
-            kind=lstm_autoencoder.lstm_model, lookback_window=lookback_window
+
+        gen = get_TimeseriesGenerator(
+            X, batch_size=1, lookback_window=lookback_window, lookahead=1
         )
-        gen = model.generate_window(X)
+        gen = (e for e in gen)
```
milesgranger: So `TimeseriesGenerator` is not a generator? Otherwise, maybe adding `iter()` to the returned value is better, or?

gordo_components/model/models.py
```diff
@@ -786,25 +652,33 @@ def fit(
...
-        steps_per_epoch = self.calc_steps_per_epoch(X)
 
-        self.kwargs["steps_per_epoch"] = steps_per_epoch
+        # these are only used for the intermediate fit method (fit method of KerasBaseEstimator),
+        # called only to initiate the model
+        # batch_size 1 so we only re-train on a single sample
+        tsg = get_TimeseriesGenerator(
+            data=X, batch_size=1, lookback_window=self.lookback_window, lookahead=1
+        )
+        super().fit(*tsg[0], epochs=1, verbose=0)
```
flikka: Probably fine. For the cross_val_score it was "ok" already, because the model sent to the cross val was not trained (fit) before that point. And the first step of the cross_val is to clone the model, so that each fold gets a separate thing.

gordo_components/model/models.py
```diff
@@ -390,10 +390,6 @@ def __init__(
         }
         super().__init__(**kwargs)
 
-    @abc.abstractmethod
-    def generate_window(self, X: np.ndarray, output_y: bool = True):
-        ...
-
     def _reshape_samples(
```
Natalie-Caruana: I guess we can remove this method now since we're not using it anymore?

epa095:
Jajaja @codecov , I will add some more test!

gordo_components/model/models.py
```diff
@@ -549,26 +437,31 @@ def fit(
...
-
-        steps_per_epoch = self.calc_steps_per_epoch(X)
-
-        self.kwargs["steps_per_epoch"] = steps_per_epoch
+        X = self._validate_and_fix_size_of_X(X)
 
+        # We call super.fit on a single sample (notice the batch_size=1) to initiate the
+        # model using the scikit-learn wrapper.
+        tsd = get_TimeseriesGenerator(
+            data=X, batch_size=1, lookback_window=self.lookback_window, lookahead=0
```
flikka: Would it be more efficient to just send in a slice of X? Since we only need a single sample? I guess not much point, since it doesnt actually do much before the batch is yielded...

tests/gordo_components/model/test_model.py
```diff
@@ -203,62 +222,35 @@ def test_lookback_window_forecast_valueerror_during_transform(self):
             with self.assertRaises(ValueError):
                 model.fit(X)
 
-    def test_generate_window_lstm_ae_output(self):
-        # Check that right output is generated from generate_window (for KerasLSTMAutoEncoder)
+    def test_get_TimeseriesGenerator(self):
```
flikka: Are the lookahead (and less importantly batch_size) features properly tested here? Seems like all tests have lookahead=1 and batch_size=1. I'm not saying that we should test the TimeseriesGenerator, but perhaps the shifting we do with the lookahead should be tested a bit more?

gordo_components/model/models.py
```diff
@@ -902,3 +750,87 @@ def score(
         out = self.transform(X)
 
         return explained_variance_score(out[:, : X.shape[1]], out[:, X.shape[1] :])
+
+
+def get_TimeseriesGenerator(
```
flikka: maybe create_keras_timeseriesgenerator?

flikka (changes requested):
Very nice that you found these improvements. Some comments attached...

tests/gordo_components/model/test_model.py
epa095: No, pretty confusing name given to it by Keras guys ....

gordo_components/model/models.py
epa095: We should type them, I update.

epa095: I want it as a free function, and I will change the name.

epa095: What about `get_keras_sequence` since the point is really that it returns a Keras Sequence object? Idk, naming is hard.

tests/gordo_components/model/test_model.py
```diff
@@ -222,34 +222,66 @@ def test_lookback_window_forecast_valueerror_during_transform(self):
             with self.assertRaises(ValueError):
                 model.fit(X)
 
-    def test_get_TimeseriesGenerator(self):
+    def test_get_TimeseriesGenerator_lb3_loah0_bs2(self):
```
flikka: Beautiful, thank you.

tests/gordo_components/model/test_model.py
epa095: it should, I updated now.

gordo_components/model/models.py
epa095: Very good name!

gordo_components/model/models.py
epa095: I went with @flikka 's suggestion `create_keras_timeseriesgenerator`

gordo_components/model/models.py
epa095: it would. I did that. Good tip

epa095:
Whatcha think @flikka ?
I will of course rebase and reword the commits

flikka (approved):

gordo_components/model/models.py
```diff
@@ -608,23 +510,26 @@ def transform(self, X: np.ndarray) -> np.ndarray:
...
         >>> model_transform.shape
         (3, 4)
         """
-        self._validate_size_of_X(X)
-        gen = self.generate_window(X, output_y=False)
-
+        X = self._validate_and_fix_size_of_X(X)
+        # predict batch_size does not need to correspond to
+        # training batch_size
+        tsd = create_keras_timeseriesgenerator(
```
Natalie-Caruana: replace tsd with tsg

gordo_components/model/models.py
```diff
@@ -608,23 +510,26 @@ def transform(self, X: np.ndarray) -> np.ndarray:
...
+            data=X, batch_size=10000, lookback_window=self.lookback_window, lookahead=0
+        )
         with possible_tf_mgmt(self):
-            xhat = self.model.predict_generator(
-                gen, steps=X.shape[0] - (self.lookback_window - 1)
-            )
+            xhat = self.model.predict_generator(tsd)
```
Natalie-Caruana: same here "tsg"

gordo_components/model/models.py
```diff
@@ -902,3 +762,87 @@ def score(
...
+
+
+def create_keras_timeseriesgenerator(
+    data: np.ndarray, batch_size: int, lookback_window: int, lookahead: int
+) -> keras.preprocessing.sequence.TimeseriesGenerator:
+    """
+    Provides a `keras.preprocessing.sequence.TimeseriesGenerator` for use with
+    LSTM's, but with the added ability to specify the lookahead of the target in y.
+
+    If lookahead==0 then the generated X is equal to the generated Y. If
```
Natalie-Caruana: isn't y = X[lookback_window-1:] if lookahead==0?

epa095: Fixed docstring slighty, thanks

PR #303, opened 2019-06-01
Push pypi packages from travis

### LINKED ISSUE #302: Build pip packages
According to travis documentation it is [easy](https://docs.travis-ci.com/user/deployment/pypi/). But we need a pypi user, but there seems to already be one called [statoil-travis](https://pypi.org/user/statoil-travis/). Maybe we can just use it? Like [this](https://github.com/equinor/cwrap/blob/master/.travis.yml).

epa095:
Only need to get the encrypted password, can probably ask @jokva  or  @markusdregi 
This closes #302

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,3 +43,12 @@ deploy:
     script: make images push-prod-images
     on:
       tags: true
+  - provider: pypi
+    user: statoil-travis
+    distributions: bdist_wheel sdist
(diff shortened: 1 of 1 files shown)


epa095:
Try it @milesgranger ?

milesgranger (approved):
LGTM!

Issue #221, opened 2019-04-29
Epic: Aftermath actions of Open Source


flikka:
Please add elements that we need to do (tasks) here, or make issues and label them with open source. 

The key approvals for making Gordo open source is as far as flikka knows to get approval from relevant line managers, for Gordo flikka believes this to be the Data Science line, and the IT line.  VP of data science and SIB leader are positive.

Things to do:

- [x] Decide on licensing - discuss with Chief Eng IT representatives.
- [x] Make better "overall" documentation, architecture drawing or flow. Issues #137 & #323 
- [x] Add more example data?
- [x] Readthedocs
- [x] Add coveralls or similar? #156 
- [x] Push to PyPi (https://github.com/equinor/gordo-components/pull/303)
- [x] Move stuff from infrastructure into components? Maybe keep equinor specific stuff in infrastructure and move e.g. the workflow template into components? At least it makes sense to open source it as well. (#534) 
- [x] In workflow template: parameterize the docker container registry name, which today is hard-coded to auroradevacr (#548)

epa095:
We just did it

milesgranger:
Think this was more serving as a tracking issue, no? Lots of unchecked boxes left.

flikka:
I agree, think that was a bit premature, look - a new title!

PR #306, opened 2019-06-03
Add .pyup.yml config

milesgranger:
Part of #221 
Do we want this?

### DIFF
--- a/.pyup.yml
+++ b/.pyup.yml
@@ -1,0 +1,40 @@
+# set the default branch
+# default: empty, the default branch on GitHub
+branch: master
+
+# update schedule
+# default: empty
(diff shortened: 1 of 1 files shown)


.pyup.yml
```diff
@@ -0,0 +1,38 @@
...
+
+# Specify requirement files by hand, default is empty
+# default: empty
+# allowed: list
+requirements:
+  - requirements.in:
+      update: all
+      pin: False
+  - requirements.txt:
+      update: False
```
epa095: Tjatjo, maybe try to have it at all and see how it works out?

epa095 (approved):

PR #309, opened 2019-06-03
Update test_requirements

epa095:
First running pip-tools to update test_requirements, and then changing several items from being pinned version in `test_requirements.in` to be abstract requirements. 

I can squash them all, or just reword the first to something better.


flikka (approved):
🍭

PR #312, opened 2019-06-04
Add semver cache key

### LINKED ISSUE #301: Add some versioning to the model-cache
Sometimes we might find that we do changes to the model-builder of such a nature that we want to force a rebuild of cached models. Maybe all models, or just those who are of a specific type. There are two ways (at least) to do this:
- Add a magic number to the `calculate_model_key` function which will be used as part of the hash-key, and which we increment for each release where we want to force a rebuild. Then all models will be rebuilt.
- Let the models contain a magic number which will be gathered up before `calculate_model_key` is called (or by it) and used as part of the hash-key. 

The advantage of the latter method is that it allows us to only force a rebuild of the specific mode types we want. The disadvantage is that it realistically will never be done, so maybe go for the first ?

In a fancy future we might want to add e.g. the major and minor version of model-builder to the cache key, but we bump them so often no that we might not want to do that yet

milesgranger:
Will close #301

### DIFF
--- a/gordo_components/builder/build_model.py
+++ b/gordo_components/builder/build_model.py
@@ -13,7 +13,7 @@ from sklearn.model_selection import cross_val_score, TimeSeriesSplit
 from sklearn.pipeline import Pipeline
 
 from gordo_components.util import disk_registry
-from gordo_components import serializer, __version__
+from gordo_components import serializer, __version__, MAJOR_VERSION, MINOR_VERSION
 from gordo_components.dataset.dataset import _get_dataset
 from gordo_components.dataset.base import GordoBaseDataset
 from gordo_components.model.base import GordoBase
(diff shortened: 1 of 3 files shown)


gordo_components/builder/build_model.py
```diff
@@ -214,12 +214,15 @@ def calculate_model_key(
     # Sets a lot of the parameters to json.dumps explicitly to ensure that we get
     # consistent hash-values even if json.dumps changes their default values (and as such might
     # generate different json which again gives different hash)
+    major, minor, _path = _parse_version(__version__)
```
epa095: `_path` or `_patch`

epa095 (approved):
minor fix

gordo_components/builder/build_model.py
milesgranger: ha, misspelled that every time I wrote it, this one slipped by. Thanks!

PR #313, opened 2019-06-04
Ensure model-config in builder is fully expanded

### LINKED ISSUE #300: Model-cache: Use fully expanded model-pipeline
We allow people to write model-pipelines without all parameters set, and the defaults are used. But if e.g. sklearn changes a default this is a bit shit, since we will then pick up the old model from the cache, but it should really have had a new behaviour. Also, it is maybe nice to store the fully expanded model-pipeline in the metadata so one can see what parameters were actually there.

It should be as easy as doing 
```python
pipeline_into_definition(pipeline_from_definition(json-pipeline))
```
and then storing this in the metadata before the cache-key is generated.

milesgranger:
Closes #300

### DIFF
--- a/gordo_components/cli/cli.py
+++ b/gordo_components/cli/cli.py
@@ -17,7 +17,11 @@ from gordo_components.data_provider.providers import (
     DataLakeProvider,
     InfluxDataProvider,
 )
-from gordo_components.serializer import load_metadata
+from gordo_components.serializer import (
+    load_metadata,
+    pipeline_into_definition,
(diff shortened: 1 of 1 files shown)


epa095 (commented):
If you place this further out, either in `provide_saved_model` or all the way out in `gordo_components.cli.cli.build` then it will be part of the cache-key as well, ensuring that "magically" whever scikit changes / adds / removes some default parameters the models will be rebuilt.

epa095 (approved):
LGTM

PR #314, opened 2019-06-04
Fix setup.py license and supported python versions

milesgranger:


setup.py
```diff
@@ -14,18 +14,15 @@ def requirements(fp: str):
     author_email="milg@equinor.no",
```
epa095: Can you also change this to `fg_gpl@equinor.com`

epa095: And can you also set `author="Equinor ASA"`

epa095 (approved):
looks GREAT to me

PR #315, opened 2019-06-04
Set GordoBaseDataset.join_timeseries to use label=left

### LINKED ISSUE #265: dataset resample: explicitly say "label": "left"
It is left for most cases, except some resample rates which makes it right. That is confusing. Lets make it constant left.

milesgranger:
Will close #265


epa095 (approved):
Whats to say. Looks good.

PR #316, opened 2019-06-04
Upgrade Travis ubuntu to xenial 16.04

epa095:
This removes the httpbin-based tests which downloaded a httpbin docker image and used it to test our async get/post wrappers.

It also updates the travis ubuntu version to 16.04. This is soon the default, so good to be on it to be ready for the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ cache: pip
 os:
   - linux
 
+dist: xenial
 
 jobs:
   include:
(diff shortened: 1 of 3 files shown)


flikka (approved):
I don't mind removing the tests, but would not have hated a small explanation about why it is not important to have them there. Not an essay, but more like a Haiku.

milesgranger:
Invalidated by #442  
and the httpbin tests removed in #451

PR #317, opened 2019-06-04
Upgrade sklearn

### LINKED ISSUE #298: Try sklearn 0.21
We have `scikit-learn~=0.20`, but latest stable is `0.21.2`. We should upgrade and make sure stuff works.

milesgranger:
Will close #298


epa095 (approved):
LGTM. I guess there is no pressing reason to try to stay compatible with 0.20 as well?

milesgranger:
Not that I see.. and that's an interesting view.. these tests vs. _compatibility_ tests? :man_shrugging:

PR #319, opened 2019-06-04
Switch to intel-tensorflow

### LINKED ISSUE #245: Try intel-tensorflow
There is a intel-tensorflow [on pypi](https://pypi.org/project/intel-tensorflow) and supposedly it can give [quite a speedup](https://software.intel.com/en-us/articles/tensorflow-optimizations-on-modern-intel-architecture).
Maybe we should try to build our docker images with it, and see it things are faster on the azure world. According to [tensorflow](https://www.tensorflow.org/guide/performance/overview#tensorflow_with_intel%C2%AE_mkl_dnn):
> In addition to providing significant performance improvements for training CNN based models, compiling with the MKL creates a binary that is optimized for AVX and AVX2. The result is a single binary that is optimized and compatible with most modern (post-2011) processors.

Sounds OK. But we should maybe have some benchmark?

milesgranger:
Closes #245


flikka (changes requested):
This actually doesn't work on mac. Not sure if we care... intel-tensorflow only comes in 0.0.1 for things other than "anylinux". 0.0.1 doesn't work (fails on import in a Keras tensorflow_backend.py file)

epa095:
Good point @flikka . Then if we want to try intel-tensorflow maybe a solution is to add it to the docker images only, so as a step in the Dockerfile we installed intel-tensorflow. The question is how to make that pretty together with having `tensorflow` in the `requirements.txt`. Will installing intel-tensorflow in the image before stop it from installing `tensorflow` (because it somehow registers `intel-tensorflow` as being  also `tensorflow`, or should we have a step after installing `requirements.txt` where we remove `tensorflow` and install `intel-tensorflow`? Some lameness anyway, so I think at least it warants an hour or two of benchmarking modelbuilding *in the cluster* with `intel-tensorflow` and see if it makes a difference. If it significantly decreases model-building time (half?) then maybe(?) it is worth it?

flikka:
I did a bit of testing, and if I have both tensorflow and intel-tensorflow installed I don't know how things work. Only the intel-tensorflow causes keras to fail, but when both are installed (order of install doesnt matter) keras doesnt fail. But I don't know what that means. I guess if we want it "server only", then maybe a pip uninstall tensorflow is a safe but hacky approach.

milesgranger:
Now I'm more convinced maybe we should do what @epa095 suggested, shelf this, and if someone asks/claims the models are slow.. we can re-investigate this?

milesgranger:
Closing, unless @epa095 or @flikka objects. Feel free to reopen.

PR #320, opened 2019-06-04
Support parametrizing model location file

### LINKED ISSUE #188: Model builder: parameterize the /tmp/model-location.txt
We are [writing](https://github.com/equinor/gordo-components/blob/87b5c56b82af8f134e8af9c56be8b7f938ba0cb7/gordo_components/builder/build_model.py#L130) the model location to a tmp-file. We should definitely parameterize this, so it is possible to not write the location, or write it somewhere else. Maybe wait until  #184 is merged, as it changes things a tiny bit.

milesgranger:
Closes #188 
Was hardcoded to `/tmp/model-location.txt` now `gordo-components build --model-location-file /whatevs.txt`

### DIFF
--- a/gordo_components/cli/cli.py
+++ b/gordo_components/cli/cli.py
@@ -89,6 +89,12 @@ DEFAULT_MODEL_CONFIG = (
     "multiple times. Separate key,valye by a comma. ie: --model-parameter key,val "
     "--model-parameter some_key,some_value",
 )
+@click.option(
+    "--model-location-file",
+    help="Full path to a file to create and write the location of where the serialized model is placed.",
+    type=click.File(mode="w", lazy=False),
+    default="/tmp/model-location.txt",
+)
 def build(
     name,
     output_dir,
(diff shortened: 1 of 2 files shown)


gordo_components/cli/cli.py
```diff
@@ -89,6 +89,12 @@ def gordo():
     "multiple times. Separate key,valye by a comma. ie: --model-parameter key,val "
     "--model-parameter some_key,some_value",
 )
+@click.option(
+    "--model-location-file",
+    help="Full path to a file to create and write the location of where the serialized model is placed.",
+    type=click.File(mode="w"),
```
epa095: I think maybe you want `lazy=False` to force any errors to occur early (before model build)

epa095 (commented):
What happens if you use it twice with the same location, does it clear it before the second write, or do you get two lines in the file?

milesgranger:
My _guess_ would be that because the mode is `w` it would overwrite the file, but it's a good point to bring up and I'll update tomorrow. :+1:

epa095 (approved):
LGTM

PR #321, opened 2019-06-04
Support predicting N number of tags

milesgranger:
Add ability to use models meant for single target predictions, to produce predictions for N number of targets. So that we can use some tags to predict some other tags.

This will depend on a future implementation where the `TimeSeriesDataset` is capable of returning a `y` 

Part of #354

### DIFF
--- a/gordo_components/model/models.py
+++ b/gordo_components/model/models.py
@@ -22,7 +22,6 @@ from sklearn.metrics import explained_variance_score
 from sklearn.exceptions import NotFittedError
 from gordo_components.model.base import GordoBase
 
-
 # This is required to run `register_model_builder` against registered factories
 from gordo_components.model.factories import *  # pragma: no flakes
 
--- a/gordo_components/serializer/pipeline_from_definition.py
+++ b/gordo_components/serializer/pipeline_from_definition.py
@@ -115,6 +115,10 @@ def _build_step(
         import_str = list(step.keys())[0]
         params = step.get(import_str, dict())
 
+        # Load any possible classes in the params if this is a dict of maybe kwargs
+        if isinstance(params, dict):
+            params = _load_param_classes(params)
+
         StepClass = pydoc.locate(
             import_str
         )  # type: Union[FeatureUnion, Pipeline, BaseEstimator]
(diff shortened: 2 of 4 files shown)


gordo_components/model/multi_regressor.py
```diff
@@ -0,0 +1,142 @@
+# -*- coding: utf-8 -*-
+
+from typing import Union, List, Optional
+
+import numpy as np
+
+from sklearn.ensemble import RandomForestRegressor
+from sklearn.base import BaseEstimator, RegressorMixin, clone
+from sklearn.utils import Parallel, delayed
```
epa095: [DEPRECATED](https://scikit-learn.org/stable/modules/generated/sklearn.utils.Parallel.html#sklearn.utils.Parallel)
> DEPRECATED: deprecated in version 0.20.1 to be removed in version 0.23. Please import this functionality directly from joblib, which can be installed with: pip install joblib.

gordo_components/model/multi_regressor.py
```diff
@@ -0,0 +1,142 @@
...
+        ----------
+        base_estimator: BaseEstimator
+            Model which is used to predict individual targets of y.
+            Internally, this model is cloned and used for each axis=1 of y.
+        n_jobs : int or None, optional (default=None)
+            The number of CPUs to use to do the computation.
+            ``None`` means 1 unless in a :obj:`joblib.parallel_backend` context.
+            ``-1`` means using all processors. See :term:`Glossary <n_jobs>`
+            for more details.
+        verbose : integer, optional
```
epa095: Should the type not be `Optional[int]` then?

epa095: (and similar for others)

gordo_components/model/multi_regressor.py
```diff
@@ -0,0 +1,142 @@
...
+        Parameters
+        ----------
+        X: np.ndarray
+        y: np.ndarray
+
+        Returns
+        -------
+        self
+        """
+        parallel = Parallel(
```
epa095: If I understand correctly this will use the `loky` backend, which causes new processes. There is some fancyness to how the shared data (X) is shared among them, it writes it to `/dev/shm` (a RAM backed FS) and then the different processes reads from there.  But in docker this seems to be default 64MB. Luckily joblib [now](https://github.com/joblib/joblib/pull/609) detects the size, and does not use it if it is too smal. So I think when we run this under docker joblib would "transparantly" copy X into each of the subprocesses, causing more memory usage. One way to fix this is to set `require='sharedmem'` to `Parallel`. Then it will use threads, but I dont know how much of a problem that is. If the random forrest uses numpy then the GIL is released whenever numpy does something. But anyway probably better to be a bit slower than to crash. But feel free to make a parameter out of it, so you can choose the "use-much-memory-but-maybe-be-faster" option if you want.

gordo_components/model/multi_regressor.py
milesgranger: :male_detective:

gordo_components/model/multi_regressor.py
milesgranger: Is this karma? :sweat_smile:

gordo_components/model/multi_regressor.py
milesgranger: :boom: :computer: :boom:

flikka:
@epa095 what do you think about this? Approval awaits?

flikka (changes requested):
Cool stuff, I assume the "missing" x,y thing is the one in https://github.com/equinor/gordo-components/pull/327

It would be useful with an issue attached, to describe what we are fixing/solving in a bit more detail. Can I now make a config file with these multi-targets, and the /prediction endpoint will give predictions?

flikka:
@milesgranger - is https://scikit-learn.org/stable/modules/generated/sklearn.multioutput.MultiOutputRegressor.html relevant here? From a birds eye it looks a bit like the same thing...

flikka (approved):

flikka:
Very nice, only one thing: The endpoint (/predict) also returns the tags, which is cool - but only the "tags" tags, not the "target_tag_list" tags. Would be nice to have that too.

flikka:
Beautiful! Let's merge and release 0.25!

PR #326, opened 2019-06-05
Support building models without scoring/cross val

### LINKED ISSUE #325: Builder: Support building models without a score(?)
Today we crash on the cross-validation if the model does not have a score. But maybe we should catch that and still build the model, just without the metrics saved?

milesgranger:
closes #325

### DIFF
--- a/gordo_components/builder/build_model.py
+++ b/gordo_components/builder/build_model.py
@@ -6,7 +6,7 @@ import logging
 import os
 import time
 from pathlib import Path
-from typing import Union, Optional
+from typing import Union, Optional, Dict, Any
 
 from sklearn.base import BaseEstimator
 from sklearn.model_selection import cross_val_score, TimeSeriesSplit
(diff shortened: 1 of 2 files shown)


gordo_components/builder/build_model.py
```diff
@@ -72,7 +72,22 @@ def build_model(
...
+            "explained-variance": {
+                "mean": cv_scores.mean(),
+                "std": cv_scores.std(),
+                "max": cv_scores.max(),
+                "min": cv_scores.min(),
+                "raw-scores": cv_scores.tolist(),
+            }
+        }
+    else:
+        scores = {"message": "Unable to obtain scores for this model"}
```
epa095: I think i prefer scores to just be an empty dict (or not exists at all). As a consumer I think maybe I would like to loop over all scores and e.g. print them (as we actually do ourself when `--print-vc-scores` is set`, and this means I need to look for the magic key "message" and try to interpret that to know if we have scores or not. If you want this message then maybe there is some other place we can put it? Or maybe just log it?

milesgranger: Oh ya, I agree, that's much better.

epa095 (approved):
LGTM. Remember squash commits

PR #327, opened 2019-06-06
Allow TimeseriesDataset to take and output target tags

milesgranger:
Part of #354

### DIFF
--- a/gordo_components/dataset/datasets.py
+++ b/gordo_components/dataset/datasets.py
@@ -1,9 +1,11 @@
 # -*- coding: utf-8 -*-
 
 import logging
-from typing import Tuple, List, Dict, Union
-import pandas as pd
+
+from typing import Tuple, List, Dict, Union, Optional, Iterable
 from datetime import datetime
+
+import pandas as pd
 
 from gordo_components.data_provider.providers import RandomDataProvider
 from gordo_components.dataset.base import GordoBaseDataset
@@ -22,6 +24,7 @@ class TimeSeriesDataset(GordoBaseDataset):
         from_ts: datetime,
         to_ts: datetime,
         tag_list: List[Union[str, Dict, SensorTag]],
+        target_tag_list: Optional[List[Union[str, Dict, SensorTag]]] = None,
         resolution: str = "10T",
         row_filter: str = "",
         **_kwargs,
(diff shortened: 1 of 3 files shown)


gordo_components/dataset/datasets.py
```diff
@@ -64,20 +74,31 @@ def __init__(
...
-        y = None
         if self.row_filter:
-            X = pandas_filter_rows(X, self.row_filter)
-        logger.info(f"First five rows of the filtered dataset are {X.head()}")
+            data = pandas_filter_rows(data, self.row_filter)
+
+        x_tag_names = [tag.name for tag in self.tag_list]
+        y_tag_names = [tag.name for tag in self.target_tag_list]
+
+        X = data[x_tag_names]
```
epa095, line 93: So this is the thing which ensures that you get the tags out again in the expected order (i.e. in the same order as they were in the `tag_list`? Can you maybe add a test which verifies this, i.e. one where the sensors are listed as `["Sensor 2", "Sensor 1"]`, and ensure that they come out with the first columns being `Sensor 2`

milesgranger: Very good, as of now, it will return them in the same order, but definitely good to make a test to verify it stays that way. :100:

tests/gordo_components/dataset/test_dataset.py
```diff
@@ -194,6 +195,38 @@ def test_row_filter(self):
...
+        [],
+    ],
+)
+def test_timeseries_target_tags(tag_list, target_tag_list):
+    start = dateutil.parser.isoparse("2017-12-25 06:00:00Z")
+    end = dateutil.parser.isoparse("2017-12-29 06:00:00Z")
+    tsd = TimeSeriesDataset(MockDataSource(), start, end, tag_list, target_tag_list)
+    X, y = tsd.get_data()
+
+    # If we have targets, y and X should be equal and axis=1 == N in target tag list
```
flikka: It sounds from the comment that y and X should be equal, but you probably mean equal _length_?

milesgranger: :male_detective:

gordo_components/dataset/datasets.py
milesgranger: :boom: :checkered_flag:

flikka (changes requested):
Cool stuff. To pick this up again. ON its own, it is fine, but it would be nice to have an issue associated with the PR, to make it clear what we expect this to do. For example if we make a config file with a target_tag_list and a supervised model in the pipeline, will I get a working /prediction endpoint, accepting len(non target tags) as input and providing len(target_tags) as output?

flikka:
Small issue: if we write a "target_tag_list" in the config, that relies on the "asset" keyword, it wont work - since we need to add tag normalization in the build part of the cli too.

flikka (approved):

PR #329, opened 2019-06-07
Update README

milesgranger:
Adds a first attempt at a logo, probably a placeholder for now.. but more importantly updates that we can install the project via `pip`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,28 @@
-# Gordo Components
 
-[![Build Status](https://travis-ci.com/equinor/gordo-components.svg?token=9cHSKigsoXktTGTEJsVA&branch=master)](https://travis-ci.com/equinor/gordo-components)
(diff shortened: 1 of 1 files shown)


epa095:
Can we actually use the logo? I.e., how was it made? If it is a stitched-together collection of symbols then where did they come from? Thinking about copyright here.

milesgranger:
https://www.freelogodesign.org/

Guess the burden of responsibility is on them.

flikka (dismissed):
I love it! BUT: The logo is too big and centered. Can you make it a bit smaller, and write Gordo on the side?

Natalie-Caruana:
Other than the arrows I like it :) ...how about if we replace the arrows with regular lines and enclose logo in a circle?

epa095 (commented):
I propose you remove the logo, and we can bikeshed over that forever, but then we can merge the other stuff

epa095 (approved):
LGTM. @flikka  remove your objections plz!

PR #331, opened 2019-06-09
Testing splitting into 12

epa095:
Another test at splitting the tests, now with coverall which claims they are able to merge the results from all the 12 tests!

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,22 +7,22 @@ test             = pytest --addopts "-n auto -m 'not dockertest'"
 testall          = pytest --addopts "--cov-report=xml --cov=gordo_components"
 
 # Only run tests which use docker
-testdocker       = pytest --addopts "-m 'dockertest'"
+testdocker       = pytest --addopts "-m 'dockertest' --cov-report=xml --cov=gordo_components"
 
 # Component level filtering of tests
-testbuilder      = pytest --addopts "tests/gordo_components/builder"
-testcli          = pytest --addopts "tests/gordo_components/cli"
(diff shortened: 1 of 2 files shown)


epa095:
... And it seems to actually work! A bit annoying that when it gets the results from the first test it immediately reports it as failing with a very low test-%, but then as it receives the rest it changes its mind. So thats nice.

setup.cfg
```diff
@@ -7,22 +7,22 @@ test             = pytest --addopts "-n auto -m 'not dockertest'"
...
-testcli          = pytest --addopts "tests/gordo_components/cli"
-testclient       = pytest --addopts "tests/gordo_components/client"
-testdataprovider = pytest --addopts "tests/gordo_components/data_provider"
-testdataset      = pytest --addopts "tests/gordo_components/dataset"
-testmodel        = pytest --addopts "tests/gordo_components/model"
-testserializer   = pytest --addopts "tests/gordo_components/serializer"
-testserver       = pytest --addopts "tests/gordo_components/server"
-testutil         = pytest --addopts "tests/gordo_components/util"
-testwatchman     = pytest --addopts "tests/gordo_components/watchman"
+testbuilder      = pytest --addopts "tests/gordo_components/builder --cov-report=xml --cov=gordo_components"
```
milesgranger, line 13: I _think_ you would only need to at this to the `pytest.ini` where we set a bunch of the default opts.

flikka:
@epa095 you want this PR?

milesgranger:
@epa095 should we close this ol girl? 
 #442

PR #333, opened 2019-06-11
Use .get_params() in pipeline_into_definition

### LINKED ISSUE #332: Use .get_params() in pipeline_into_definition
https://github.com/equinor/gordo-components/blob/167fce805c6e5bada8b7c601ae9edf2775b64e20/gordo_components/serializer/pipeline_into_definition.py#L87

Currently we only inspect the `__init__` of the `BaseEstimator` when decomposing a node. However our Keras models allow for `kwargs` which are passed to either subsequent fits to the underlying Keras model or to the model builder factory. By using `.get_params()` we ensure we get all the parameters the model can be initialized with.

milesgranger:
Will close #332 

Before we only looked at the step's __init__ args, which didn't
account for kwargs sent to the model.

### DIFF
--- a/gordo_components/serializer/pipeline_into_definition.py
+++ b/gordo_components/serializer/pipeline_into_definition.py
@@ -2,10 +2,10 @@
 
 import inspect
 import logging
-from typing import Iterable, Dict, Union, List  # pragma: no flakes
+from typing import Iterable, Dict, Union, List, Any  # pragma: no flakes
 
-from sklearn.pipeline import Pipeline, FeatureUnion
-from sklearn.preprocessing import FunctionTransformer
+from sklearn.base import BaseEstimator
+from sklearn.pipeline import Pipeline
 
 
 logger = logging.getLogger(__name__)
@@ -63,7 +63,7 @@ def pipeline_into_definition(
     return steps
 
 
-def _decompose_node(step: object, prune_default_params: bool = False):
+def _decompose_node(step: BaseEstimator, prune_default_params: bool = False):
     """
     Decompose a specific instance of a scikit-learn transformer,
     including Pipelines or FeatureUnions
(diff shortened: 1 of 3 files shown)


gordo_components/serializer/pipeline_into_definition.py
```diff
@@ -84,34 +85,27 @@ def _decompose_node(step: object, prune_default_params: bool = False):
     """
 
     import_str = f"{step.__module__}.{step.__class__.__name__}"
-    init_params = inspect.getfullargspec(step.__class__.__init__).args
+    params = step.get_params(deep=False)
```
epa095, line 87: Should this not be guarded by a hasattr?

epa095: Or do we not support non-scikit-learn stuff (although that would be a bit lame)

milesgranger: ya.. I saw this and thought too. But I mean, I don't know _how_ you could fit a step of a pipeline, into a pipeline, without it being a BaseEstimator. And by that time, seems like a bag of worms?

flikka (approved):
LGTM

gordo_components/serializer/pipeline_into_definition.py
flikka: I think it is ok to assume BaseEstimator...

tests/gordo_components/serializer/test_serializer_into_definition.py
```diff
@@ -126,46 +126,47 @@ def test_pipeline_into_definition(self):
...
                                 n_iter: 5
                                 random_state:
                                 tol: 0.0
-                            memory:
                             verbose: false
-                        n_jobs:
                         transformer_weights:
                         verbose: false
                     - gordo_components.model.models.KerasAutoEncoder:
+                        build_fn:
```
epa095: It is a bit odd that we pass this one around, but we "explicitly" ignore it in our constructor in `KerasAutoEncoder`. I see how it comes there, the call to `params = super().get_params(**params)` in `get_params` provides it. Should we remove it before we return in `get_params`, or just ignore this whole "problem"?

epa095 (commented):
Small question, otherwise LGTM

tests/gordo_components/serializer/test_serializer_into_definition.py
milesgranger: I have no problem removing it, certainly something which is used "internally", but the only potential problem is if some froggy data scientist starts playing around with parameters within their notebook and are passing functions directly. I think @Natalie-Caruana was doing this at one point; passing the builder function directly to the model

epa095: But we already dont support that really, since the constructor of `KerasAutoEncoder` completly ignores any `build_fn` parameter

milesgranger:
Have ran this through @flikka's gauntlet with updated parameters such as `epochs` and ensured it works. :+1: 
image

epa095 (approved):
Here, take my approval and fare well with it!

tests/gordo_components/serializer/test_serializer_into_definition.py
milesgranger: ah, so right, I was thinking about how you can pass `kind` a function. I'll update to remove the `build_fn` :+1:

PR #337, opened 2019-06-13
Document endpoints for deployed models

### LINKED ISSUE #323: Update Sphinx docs for endpoints
Should add an additional tab which explains, as the Swagger API UI does, what endpoints will be available and what they return.

milesgranger:
Will close #323

### DIFF
--- a/docs/general/endpoints.rst
+++ b/docs/general/endpoints.rst
@@ -1,0 +1,219 @@
+Endpoints
+---------
+
+==================
+Project index page
+==================
+
+Going to the base path of the project, ie. ``/gordo/v0/my-project/`` will return the
+project level index, with returns a collection of the metadata surrounding the models currently deployed and their status.
+Each ``endpoint`` key has an associated ``endpoint-metadata`` key which is the direct transferal of metadata returned from
+the ML servers at their :ref:`ml-server-metadata-route` route.
+
+.. code-block:: python
+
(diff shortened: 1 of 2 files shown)


flikka:
Nice. Professional.

docs/general/endpoints.rst
```diff
@@ -0,0 +1,163 @@
+Endpoints
+---------
+
+When a model is deployed from a config file, it results in a ML
+server capable of the following paths:
+
+All paths listed below should be prefixed with ``/gordo/v0/<project-name>/<model-name>``
```
epa095: Maybe a slightly different wording. From the model-servers perspective this is not the required path right, it could actually be served anywhere. But maybe a sentence that in a "normal" deployment or something like that?

docs/general/endpoints.rst
```diff
@@ -0,0 +1,163 @@
...
+Various metadata surrounding the current model and environment.
+
+
+/
+=
+
+This is the Swagger UI for the given model
+
+
+Project index page
```
epa095: Maybe a comment noting that this is the output from watchman, which by default is routed at ... ?

flikka (changes requested):
I am missing some information about the _input_ to the endpoints. It says what it will output, but what one must supply as input should also be documented, no?

Also - a bit confused (would be clearer if the input was specified) - do we still have two anomaly endpoints, one "get" and one "post" (ie supply dates or supply data?)

flikka (approved):

PR #342, opened 2019-06-20
Ensure that start and end has UTC tzone on date_range call

flikka:
Fix https://github.com/equinor/gordo-infrastructure/issues/287

Essentially - the date_range() method could have start and end datetime inputs with different time-zone, since we get the these timestamps from various sources (config file, date in bash etc). 

Convert them to utc.


epa095 (approved):
LGTM

PR #344, opened 2019-06-23
Add multithreaded download of tags from NCS

### LINKED ISSUE #341: Fetch data(lake?) data in parallell
This feature is not too important for the model-builder / gordo-style on cluster, but if you use gordo-components more interactively, you'll find that loading the tag files sequentially is slow.

If implemented, it should be possible to still run single-threaded (perhaps that should be default).

epa095:
The default is now to have threads `None`, which leaves it to the data-provider to choose. For IROC it is constant 50 threads, for NCS it is your nr of cores * 5 (this is the default "smart" choice for ThreadPoolExecutor). 

Gordo model builder hard-codes it to 1 thread to save memory.

This closes #341

### DIFF
--- a/gordo_components/cli/cli.py
+++ b/gordo_components/cli/cli.py
@@ -95,6 +95,13 @@ DEFAULT_MODEL_CONFIG = (
     type=click.File(mode="w", lazy=False),
     default="/tmp/model-location.txt",
 )
+@click.option(
+    "--data-provider-threads",
+    help="Number of threads to use for the data provider when fetching data",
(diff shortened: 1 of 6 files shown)


epa095:
I have two things left to do:
1. Expose the threads as parameters for both the model builder and the client
2. Set a default threads nr for the client, as it is it will be `None`

flikka (commented):
Good stuff, except the failing test.

epa095:
Cool, do I get an accept @flikka ?

flikka (approved):
BIEN

PR #346, opened 2019-06-24
Change default keras activation function to tanh

### LINKED ISSUE #268: Change default activation function from relu to tanh
`relu` has a tendency to give infinity and NaNs all over. Lets change the default keras activation functions to `tanh`.

epa095:
It was `relu` before, but that had a tendency to cause infinity and NaNs
all over the place.

This closes #268


epa095:
@Natalie-Caruana  said `tanh` was good, and *[the internet](https://stats.stackexchange.com/questions/330559/why-is-tanh-almost-always-better-than-sigmoid-as-an-activation-function)* seems to agree

flikka (approved):
Good

PR #351, opened 2019-06-25
Added PERA as New Asset

choukha:
Added PERA from Brazil data as new asset


gordo_components/data_provider/ncs_reader.py
```diff
@@ -24,6 +24,7 @@ class NcsReader(GordoBaseDataProvider):
         "1130-troa": "/raw/corporate/Aspen MS - IP21 Troll A/sensordata/1130-TROA",
         "1101-sfb": "/raw/corporate/IMS Statfjord/sensordata/​1101-SFB",
         "1218-gkr": "/raw/corporate/PI System Manager Sleipner/sensordata/1218-GKR",
+        "1295-pera": "raw/corporate/Aspen MS - IP21 Peregrino/sensordata/1295-PERA",
```
flikka: Don't know if it is required, but I would like to see a leading forward slash before raw on this too.

flikka (changes requested):

gordo_components/data_provider/ncs_reader.py
choukha: Yes, Putting it there. my mistake.

flikka (approved):

PR #352, opened 2019-06-25
Add output activation function for feedforward NN as a parameter

Natalie-Caruana:


gordo_components/model/factories/feedforward_autoencoder.py
```diff
@@ -46,6 +47,7 @@ def feedforward_model(
     decoding_dim = dec_dim or [64, 128, 256]
     encoding_func = enc_func or ["tanh", "tanh", "tanh"]
     decoding_func = dec_func or ["tanh", "tanh", "tanh"]
+    out_func = out_func or "linear"
```
epa095: No, this can be a concrete default argument like `     out_func: str = "linear", ` in the parameter. That is cleaner. The only reason we are not doing that for the encoding/decoding func is because they are lists.

epa095 (commented):
Also, does it make sense to add the same to the lstm-factories?

gordo_components/model/factories/feedforward_autoencoder.py
Natalie-Caruana: ok

Natalie-Caruana:
for the lstm it's already done.

gordo_components/model/factories/feedforward_autoencoder.py
```diff
@@ -18,6 +18,7 @@ def feedforward_model(
     dec_dim: List[int] = None,
     enc_func: List[str] = None,
     dec_func: List[str] = None,
+    out_func: str = None,
```
milesgranger: `out_func` missing in parameter docs.

milesgranger: Also, this should just be set as `"linear"` and not use the `or` concept below.

epa095 (approved):
Rebase them to a single commit and I am happy

PR #357, opened 2019-06-26
Optimize ML server performance

### LINKED ISSUE #200: Upgrade servers to production level serving
The ML server uses the default flask server, we should update these to a more production quality server; preferably one with single process and greenthreads or something similar. Which may also allow us to remove the manual management of Keras tensorflow sessions, as these have a hard time playing well in a multi-threaded environment. 

Gunicorn?

### LINKED ISSUE #356: Optimize ML server performance

milesgranger:
Will close #356


gordo_components/server/views/base.py
```diff
@@ -304,17 +304,41 @@ def multi_lvl_column_dataframe_to_dict(df: pd.DataFrame) -> typing.List[dict]:
         Convert a dataframe which has a pandas.MultiIndex as columns into a dict
         where each key is the top level column name, and the value is the array
         of columns under the top level name.
+
+        Notes
```
epa095: I think this notes section should be for the users of the function, not the future programmers. I.e. i like the comment, but I want it below with the actual code

epa095 (commented):
Does it really close #200?

milesgranger:
@epa095 Not yet.. but I thought while I'm in there optimizing the code, might as well upgrade the WSGI server as well, or do you fancy a separate PR for that?

epa095:
> @epa095 Not yet.. but I thought while I'm in there optimizing the code, might as well upgrade the WSGI server as well, or do you fancy a separate PR for that?

Yeah I think especially the WSGI server thing is another PR

milesgranger:
@epa095 @flikka Shall I remove the `benches` concept?

benches/test_ml_server.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
+
```
epa095: Can you 
1: Write a line or two about how to run this (cause I assume you have to run it with pytest/profile to have the `benchmark.pedantic` fixture)
2: Change the folder to e.g benchmarks
3: Put a single line comment above the "benchmarks" fixture saying its a pytest fixture, otherwise it is just some magic shit which is just there

epa095 (commented):
I like speedups!!

epa095 (approved):
LGTM

PR #362, opened 2019-07-02
Push 'latest' tag for docker images

### LINKED ISSUE #361: Push 'latest' tags for releases and master branch merges.
It is often times convenient to use the _latest_ stable. Instead of looking into the portal to find the current stable master tag, if we push a "latest" tag, we can always simply use that to develop with master.

milesgranger:
Will close #361


flikka (approved):
BRILLIANT

PR #363, opened 2019-07-03
Add Gullfaks A as new asset

choukha:


flikka (approved):
👍

flikka:
@chokha - looks good. Normally we let the creator do the merge, so he/she can time it when he/she wants. So you can merge it yourself. Or, maybe you should use the "squash and merge" option, I think one commit for this change is enough. But not important for me what you do.

PR #367, opened 2019-07-04
Pass kwargs to keras compiling

### LINKED ISSUE #366: Pass kwargs to keras.Model.compile calls
When a user entered `keras.Model.compile` specific parameters in their yaml, ie.
```yaml
gordo_components.model.models.SomeKerasModel:
  metrics:
    - mse
```
`metrics` is a parameter to `keras.Model.compile` but it never reaches that method. `kwargs` should also be passed, which can contain arbitrary parameters.

milesgranger:
Will close #366 

Allows arbitrary parameter passing to keras.Model.compile
which also accepts **kwargs

### DIFF
--- a/gordo_components/model/factories/model_factories_utils.py
+++ b/gordo_components/model/factories/model_factories_utils.py
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
+
 import math
 
 
--- a/gordo_components/model/factories/feedforward_autoencoder.py
+++ b/gordo_components/model/factories/feedforward_autoencoder.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-from typing import List, Tuple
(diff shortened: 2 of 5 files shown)


gordo_components/model/factories/model_factories_utils.py
```diff
@@ -13,3 +18,31 @@ def hourglass_calc_dims(compression_factor, encoding_layers, n_features):
         round(n_features - (i * average_slope)) for i in range(1, encoding_layers + 1)
     ]
     return dims
+
+
+def align_kwargs(ignore: List[str] = []):
```
epa095: I know that its hard to delete written code, but this is no longer needed right? Maybe keep it in your backpocket for the next time it might be needed?

epa095 (commented):
Looks mostly good to me

epa095 (approved):
LGTM

PR #368, opened 2019-07-05
Make autoencoders take a y

### LINKED ISSUE #364: Change our AutoEncoders to take target (Y) that we fit the model to.
Today, we don't take in the target - so we try to fit the neural network to the same as the inputs. In the case of pipelines, this could be scaled data, or even feature reduced or added.

This way - it is not obvious what comes out of the .predict() function.

Suggestion is to supply the X (original data into the pipeline) as the target as well - so that the end result is that what you get from the .predict() is the same dimension (number of targets) and same scale as the X you sent in.

This will make the autoencoder more similar to other "predict" type of models, where we have targets that we fit towards. Scoring will be easier, interpreting results will be easier, and doing things like feature selection in the pipeline will actually work.

Another effect is that then we can treat the "anomaly-calculations" the same way for the autoencoder, as for a multi-target regression. It is simply the error between the predicted output and the real output - the only difference being that the autoencoder has the same input as the target, but the regression will typically split the tags into some used as X to calculate some others, that will become Y.

This task is not small, it affects a lot of things, but it seems very much like the right thing to do.

### LINKED ISSUE #371: Scaling inside a FeatureUnion breaks the inverse_transform
There _seems_ to be a bug somewhere, @jieliu218 showed me an output from a call, both on the /prediction and the /anomaly/prediction endpoint (for the get-versions) where the inverse-transform-model-output seemed to lack the "inverse-tranform"

After some more testing, I can see that the problematic project has the minmax scaler inside a FeatureUnion. I guess this is the reason - that the inverse_transform is done on the FeatureUnion (or not at all, if it doesnt have inversed transform?).

milesgranger:
Will close #364 

Will also close #371 (because it will no longer be relevant)

Allows an 'autoencoder' to take X
via a Pipeline for example, but then train itself to output
the same unscaled values to match y. Models will no longer implicitly train themselves for `X` (which was often scaled via previous steps in the Pipeline) Therefore the Pipeline/
preprocessing steps before the model are part of the encoding
process.

- [X] Make all autoencoders to take a `y`
- [X] Update workflow generator (gordo-infrastructure) to set target_tag_list if not defined
- [x] Update server output (ie. `inverse-transformed-model-output` doesn't make sense anymore)
- [X] Abstract logic for "anomaly" into it's own model. 
- [x] Update readthedocs / documentation
- [x] Update Grafana according to new keys

PRs depending on this
----------------------------
- Update gordo-test-project 
  - https://github.com/equinor/gordo-test-project/pull/14
- Update Grafana
  - https://github.com/equinor/gordo-infrastructure/pull/329

PR depends on:
-------------------
- Auto create `target_tag_list` in workflow generation if not specified 
  - https://github.com/equinor/gordo-infrastructure/pull/326


gordo_components/model/factories/feedforward_autoencoder.py
```diff
@@ -186,25 +197,26 @@ def feedforward_hourglass(
...
-    >>> model = feedforward_hourglass(5)
+    >>> model = feedforward_hourglass(5, 5)
     >>> [model.layers[i].units for i in range(len(model.layers))]
     [4, 4, 3, 3, 4, 4, 5]
-    >>> model = feedforward_hourglass(10, compression_factor=0.2)
+    >>> model = feedforward_hourglass(10, 10, compression_factor=0.2)
     >>> [model.layers[i].units for i in range(len(model.layers))]
     [7, 5, 2, 2, 5, 7, 10]
-    >>> model = feedforward_hourglass(10, encoding_layers=1)
+    >>> model = feedforward_hourglass(10, 10, encoding_layers=1)
```
milesgranger: Now that `n_features_out` defaults to `n_features` this can be removed.

gordo_components/model/factories/lstm_autoencoder.py
```diff
@@ -243,25 +242,26 @@ class (e.x. Adam(lr=0.01,beta_1=0.9, beta_2=0.999)).  If no arguments are
...
-    >>> model = lstm_hourglass(5)
+    >>> model = lstm_hourglass(5, 5)
     >>> [model.layers[i].units for i in range(len(model.layers))]
     [4, 4, 3, 3, 4, 4, 5]
-    >>> model = lstm_hourglass(10, compression_factor=0.2)
+    >>> model = lstm_hourglass(10, 10, compression_factor=0.2)
     >>> [model.layers[i].units for i in range(len(model.layers))]
     [7, 5, 2, 2, 5, 7, 10]
-    >>> model = lstm_hourglass(10, encoding_layers=1)
+    >>> model = lstm_hourglass(10, 10, encoding_layers=1)
```
milesgranger: Same, these second arguments can be removed.

gordo_components/server/views/base.py
```diff
@@ -168,7 +182,59 @@ def get(self):
...
+        Parameters
+        ----------
+        df: Union[dict, list]
+            Any data supported by ``pandas.DataFrame.from_dict``
+        tags: List[str]
+            The expected column names for the dataframe. If data supplied does not
+            have column names, but the length of those columns matches the tags, these
+            tags will be used as column names.
+        name: str
+            How to format the ``flask.Response`` message when referring to the data supplied.
```
milesgranger: add that `flask.Response` only occurs if there was an error parsing the dict into a DataFrame

gordo_components/server/views/base.py
```diff
@@ -241,28 +311,6 @@ def _process_request(
                 f"Calculating model output took "
                 f"{get_model_output_time_s-process_request_start_time_s} s"
             )
-            transformed_model_input = model_io.get_transformed_input(
```
milesgranger, line 244: Can probably remove these methods `model_io.get_transformed_input` & `get_inverse_transformed_input` 

Also should `get_transformed` be added later? Which just applies `transform` on the data? May be useful for folks to see that :man_shrugging:

tests/gordo_components/model/test_lstm_autoencoder.py
```diff
@@ -142,13 +144,13 @@ def test_lstm_hourglass_checks_enc_layers(self):
...
     def test_lstm_hourglass_checks_compression_factor(self):
         """
         Test that lstm_hourglass validates parameter requirements
         """
         with self.assertRaises(ValueError):
-            lstm_hourglass(3, compression_factor=2)
+            lstm_hourglass(3, n_features_out=3, compression_factor=2)
         with self.assertRaises(ValueError):
-            lstm_hourglass(3, compression_factor=-1)
+            lstm_hourglass(3, n_features_out=3, compression_factor=-1)
```
milesgranger: These can be removed as they default to n_features

docs/general/endpoints.rst
```diff
@@ -156,42 +151,29 @@ models.
...
+    - Anomaly per feature/tag calculated from the expected tag input (y) and the model's output for those tags (yhat)
+- ``total-anomaly``:
+    - This is the total anomaly for the given point as calculated by the model.
 
 Sample response:
 
 .. code-block:: python
 
     {'data': [{'end': [None],
-               'error-transformed': [2.746687859183499,
```
epa095, line 176: But the diffbased anomaly detector can calculate errors both the scaled and unscaled right (by using the inverse tranformed), should we do that?

docs/general/endpoints.rst
```diff
@@ -156,42 +151,29 @@ models.
...
-                                0.5305765271186829,
-                                0.7146536707878113],
+               'tag-anomaly': [
+                    2.746687859183499,
+                    2.4497416272485886,
+                    2.508896707372706
+               ],
+               'total-anomaly': [60.4668517758241],
+               'model-output': [
+                    0.6726480722427368,
```
epa095: These are not actually the model outputs now right? Since we are now training the models to output X, and not X-scaled, these should be in the original high scales?

gordo_components/model/anomaly/diff.py
```diff
@@ -0,0 +1,157 @@
...
+from sklearn.base import BaseEstimator, TransformerMixin
+
+from gordo_components import serializer
+from gordo_components.model.base import GordoBase
+
+
+class DiffBasedAnomalyDetector(BaseEstimator, GordoBase):
+    def __init__(
+        self,
+        base_estimator: BaseEstimator = PCA(),
```
epa095: Should we have a default here? Is it not better to demand a base estimator? And if we want a default, is PCA a sensible one?

gordo_components/model/anomaly/diff.py
```diff
@@ -0,0 +1,157 @@
...
+        Treat this model as transparent into base_estimator unless
+        referring to something owned by this object
+        """
+        if item in self.__dict__:
+            return getattr(self, item)
+        else:
+            return getattr(self.base_estimator, item)
+
+    def get_metadata(self):
+        if isinstance(self.base_estimator, GordoBase):
```
epa095, line 60: Is it not better to just check if it has get_metadata?

gordo_components/model/anomaly/diff.py
```diff
@@ -0,0 +1,157 @@
...
+            return self.base_estimator.get_metadata()
+        else:
+            return {}  # TODO: Fill this in.
+
+    def score(
```
epa095, line 68: Do you need this override when you have the `__getattr__` magic?

gordo_components/model/anomaly/diff.py
```diff
@@ -0,0 +1,157 @@
...
+    def fit(self, X: np.ndarray, y: np.ndarray):
+        self.base_estimator.fit(X, y)
+        self.scaler.fit(y)  # Scaler is used for calculating errors in .anomaly()
+        return self
+
+    def anomaly(
+        self, X: pd.DataFrame, y: pd.DataFrame, frequency: Optional[timedelta] = None
+    ) -> pd.DataFrame:
+        """
+        Create an anomaly dataframe from the base provided dataframe.
```
epa095, line 109: Why does this not do self.predict and use the model output from there? It seems clunky for me when I am in the terminal that I first have to run `.predict`, take the output from that and put it into a dataframe and then use the `anomaly` on the same object I just got some output from. 

I understand that this comes from the web-server, but maybe ditch the whole "call basic endpoint first then pass on", and rather just call the `anomaly` method directly?

docs/components/server/anomaly.rst
```diff
@@ -0,0 +1,12 @@
+Anomaly
+-------
+
+The anomaly view into the model. Expects that the model being served
+when accessing this route implements the ``GordoBaseAnomaly`` which
```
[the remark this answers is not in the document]
milesgranger: Ah yes, previous me was expecting later me to implement the base.. good catchy.

docs/general/endpoints.rst
milesgranger: Diffbased anomaly detector doesn't have access to the pipeline, but I guess you're talking about calculating the error _without_ the `scaler` operating on the output/y as well?

milesgranger: if we _do_ that, do we also update grafana? Because if we need to do that, then other anomaly models might not have such a concept, no?

docs/general/endpoints.rst
milesgranger: ya, I guess I'm not really paying attention too much to the _values_ but the _keys_ here.

docs/general/endpoints.rst
epa095: We can, but don't have to, update grafana. Whatever is easiest at this point. But the data will be there, so its easy to add if someone wants to show.

gordo_components/model/anomaly/diff.py
milesgranger: I think we definitely should have a default, as that's the Scikit-Learn way of things.. but agree, maybe PCA isn't the best. Maybe feedforward?

gordo_components/model/anomaly/diff.py
milesgranger: Guess it could go both ways.. checking `GordoBase` ensures we know that it has `get_metadata` as well was _what_ we expect that to return. I'd also advocate that someone reading the code and seeing `hasattr(obj, "get_metadata")` doesn't mean much to them; in the sense we really _only_ mean metadata as it relates to the `GordoBase`  But I'm not strongly attached

docs/general/endpoints.rst
epa095: But yeah, all I expect the diffbased detector to do is to report the errors scaled and unscaled. Although I dont know if it is actually interesting, but I seem to remember that someone wanted that earlier. But maybe what they mostly wanted was the model output in unscaled fashion? And that we get now. 

I think I am fine with not reporting them if we dont think anyone actually wants them.

gordo_components/model/anomaly/diff.py
milesgranger: I agree, but doesn't appear Python realizes this with the inheritance of `GordoBase`.

gordo_components/model/anomaly/diff.py
milesgranger: I'll buy what you're sellin'.

However, by having the base dataframe standardized we ensure at least the default keys `model-output` and `original-input` are _always_ sent in the response and thus Influx/Grafana as well as the format of the dataframe as a whole. Because there is some logic in creating/setting the index from offsets of model output vs input, etc.

If the dataframe is constructed in each Model that implements `.anomaly()` we loose control over it's design/enforcement as more Anomaly detectors are implemented. By requiring it to take the base dataframe and return it with added keys, it allows the model to focus on _anomaly_ calculations and not uniform dataframe creation.

Maybe a way around / compromise it is to force `.anomaly()` to return a dict of dataframes; where each key is some dimension of anomaly calculation, like we do with the `DiffBasedAnomalyDetector`:
```python
{"tag-anomaly": <some-dataframe>, "total-anomaly": <another-dataframe>}
```
Then the server can just append each key as another top level column name to the base dataframe?

milesgranger: Done with https://github.com/equinor/gordo-components/pull/368/commits/49d2de7465a1b61cf25420b8fedbf596fd6c5583

gordo_components/server/views/anomaly.py
```diff
@@ -140,22 +139,21 @@ def _process_base_response(self, start_time: float = None):
...
-        if not any(col in y.columns for col in [t.name for t in self.tags]):
+        if not any(col in g.y.columns for col in [t.name for t in self.tags]):
             message = {"message": "y is not a subset of X, cannot do anomaly detection"}
             return make_response(jsonify(message), 400)
 
         # Now create an anomaly dataframe from the base response dataframe
         try:
-            anomaly_df = current_app.model.anomaly(g.data, y, frequency=self.frequency)
+            anomaly_df = current_app.model.anomaly(g.X, g.y, frequency=self.frequency)
         except AttributeError:
```
milesgranger, line 157: Maybe we check for `AnomalyDetectorBase`; because an attribute error could arise within `.anomaly` and the resulting message wouldn't be true.

gordo_components/server/views/common.py
```diff
@@ -4,9 +4,9 @@
 import functools
 import timeit
 import typing
-import traceback
+from typing import List, Union, Optional
```
milesgranger: General; maybe move `common` into it's own module, ie. `gordo_components.common` since we're now spanning the use of functions like `make_base_dataframe` into the server (by the base view) and `DiffBasedAnomalyDetector.anomaly` .. and maybe in general, we don't have a `shared`/`common` submodule anyway. :man_shrugging:

docs/general/endpoints.rst
milesgranger: Created this issue for it; should we see use in the future: https://github.com/equinor/gordo-components/issues/392

docs/components/server/anomaly.rst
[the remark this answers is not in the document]
milesgranger: Moved that be done in https://github.com/equinor/gordo-components/issues/393

gordo_components/server/views/anomaly.py
milesgranger: Should be addressed in https://github.com/equinor/gordo-components/issues/393

gordo_components/model/factories/lstm_autoencoder.py
```diff
@@ -139,6 +132,8 @@ def lstm_symmetric(
     ----------
     n_features: int
          Number of input and output neurons
+    n_features_out: Optional[int]
```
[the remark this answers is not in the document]
milesgranger: Could rename it so `lstm_symmetric_ish`? :stuck_out_tongue_winking_eye:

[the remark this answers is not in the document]
milesgranger: Added parametrized test in https://github.com/equinor/gordo-components/pull/368/commits/40c2bb349426a0b41a200fc8eecb8db0c68231d7 for such possibilities.

gordo_components/model/anomaly/diff.py
```diff
@@ -0,0 +1,152 @@
...
+from gordo_components.model.base import GordoBase
+from gordo_components.server.views import common
+from gordo_components.model.models import KerasAutoEncoder
+
+
+class DiffBasedAnomalyDetector(BaseEstimator, GordoBase):
+    def __init__(
+        self,
+        base_estimator: BaseEstimator = KerasAutoEncoder(kind="feedforward_hourglass"),
+        scaler: TransformerMixin = MinMaxScaler(),
```
epa095: Should we use something else than MinMaxScaler as default? I don't have a competent opinion really, except the observation that MinMaxScaler is quite sensitive to even a single outlier. Maybe RobustScaler, which seems to be like MinMax, only it does not use the biggest/lowest, but the 25th and 75th quantile?

gordo_components/model/anomaly/diff.py
```diff
@@ -0,0 +1,152 @@
...
+
+        Returns
+        -------
+        pd.DataFrame
+            A superset of the original base dataframe with added anomaly specific
+            features
+        """
+
+        # Get the model output, falling back to transform if 'predict' doesn't exist
+        model_output = self.predict(X) if hasattr(self, "predict") else self.transform
```
epa095: This is definitely nicer than the try-catch, even though the try-catch might be regarded as more [pythonic](https://docs.quantifiedcode.com/python-anti-patterns/readability/asking_for_permission_instead_of_forgiveness_when_working_with_files.html)

gordo_components/model/anomaly/diff.py
```diff
@@ -0,0 +1,152 @@
...
+    def get_params(self, deep=True):
+        return {"base_estimator": self.base_estimator, "scaler": self.scaler}
+
+    def fit(self, X: np.ndarray, y: np.ndarray):
+        self.base_estimator.fit(X, y)
+        self.scaler.fit(y)  # Scaler is used for calculating errors in .anomaly()
+        return self
+
+    def anomaly(
+        self, X: pd.DataFrame, y: pd.DataFrame, frequency: Optional[timedelta] = None
```
epa095, line 106: I dont understand the frequency. Dont we assume that the X includes a timeindex?

gordo_components/model/anomaly/diff.py
```diff
@@ -0,0 +1,152 @@
...
+        model_out_scaled = pd.DataFrame(
+            self.scaler.transform(data["model-output"]),
+            columns=data["model-output"].columns,
+            index=data.index,
+        )
+
+        # Calculate the total anomaly between all tags for the original/untransformed
+        # Ensure to offset the y to match model out, which could be less if it was an LSTM
+        scaled_y = self.scaler.transform(y)
+        tag_anomaly = model_out_scaled - scaled_y[-len(data) :, :]
```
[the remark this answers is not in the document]
milesgranger: Oh yes, good catch, noticed that while making the grafana dashboard, looks a bit weird.

gordo_components/model/anomaly/diff.py
milesgranger: #false, during a POST request, it wouldn't have a frequency.

gordo_components/server/views/common.py
milesgranger: Move to `models`

gordo_components/model/factories/lstm_autoencoder.py
```diff
@@ -262,6 +261,7 @@ class (e.x. Adam(lr=0.01,beta_1=0.9, beta_2=0.999)).  If no arguments are
 
     return lstm_symmetric(
         n_features,
+        n_features_out,
```
epa095, line 264: Here there is definitely kind of an issue. If you run hourglass with 40 features in, 20 out and compression_factor=0.5, how small would you expect the smallest layer to be? Today it will be 20, so the 20 out nodes can be pushed right through. To me it definitely seems reasonable to have the compression-factor multiply the *smallest* of in and out-nodes. But some care must (maybe) be taken in ensuring that it has the right nr of in-neurons as well, and that the slope from 40 to the smallest layer of 10 is reasonable, and that the slope from 10 to 20 is reasonable. It can definitely no longer use `lstm_symmetric` directly:-p
But is this maybe an imidately new PR/issue?

gordo_components/server/views/anomaly.py
```diff
@@ -152,86 +136,34 @@ def _process_base_response(self, response: Response, start_time: float = None):
...
+
         # Now create an anomaly dataframe from the base response dataframe
-        anomaly_df = self.make_anomaly_df(self._data)
+        try:
+            anomaly_df = current_app.model.anomaly(g.X, g.y, frequency=self.frequency)
+        except AttributeError:
+            msg = {
+                "message": f"Model is not an AnomalyDetector, it is of type: {type(current_app.model)}"
+            }
+            return make_response(jsonify(msg), 400)
```
epa095: Some other error code so we can distinguish it? `422 Unprocessable Entity` maybe?

gordo_components/model/factories/lstm_autoencoder.py
milesgranger: https://github.com/equinor/gordo-components/issues/399

gordo_components/server/views/anomaly.py
milesgranger: https://github.com/equinor/gordo-components/issues/400

gordo_components/model/anomaly/diff.py
```diff
@@ -0,0 +1,157 @@
...
+        It trains a ``scaler`` to the target **after** training, purely for
+        error calculations. The underlying ``base_estimator`` is trained
+        with the original, unscaled, ``y``.
+
+        Parameters
+        ----------
+        base_estimator: sklearn.base.BaseEstimator
+            The model to which normal ``.fit``, ``.predict`` methods will be used.
+        scaler: sklearn.base.TransformerMixn
+            Defaults to ``sklearn.preprocessing.MinMaxScaler``
```
epa095: nope it does not

tests/gordo_components/model/anomaly/test_anomaly_detectors.py
```diff
@@ -0,0 +1,109 @@
...
+@pytest.mark.parametrize(
+    "index", (range(10), pd.date_range("2019-01-01", "2019-01-30", periods=10))
+)
+@pytest.mark.parametrize("lookback", (0, 5))
+def test_diff_detector(base_estimator, scaler, index, lookback):
+    """
+    Test the functionality of the DiffBasedAnomalyDetector
+    """
+
+    def simulate_lstm_predict(func):
```
epa095: This is because using our actual lstm-implementation is too slow? Is it actually noticeable slower on this small dataset? If it is not then I think I would prefer using our actual keras lstm.

epa095 (commented):
Getting closer!

tests/gordo_components/model/anomaly/test_anomaly_detectors.py
```diff
@@ -0,0 +1,109 @@
...
+    X, y = (
+        pd.DataFrame(np.random.random((10, 3))),
+        pd.DataFrame(np.random.random((10, 3))),
+    )
+
+    model = DiffBasedAnomalyDetector(base_estimator=base_estimator, scaler=scaler)
+    assert model.get_params() == dict(base_estimator=base_estimator, scaler=scaler)
+
+    model.fit(X, y)
+    scaler.fit(y)
```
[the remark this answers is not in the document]
milesgranger: Correcto, but I'm fitting it separately to manually do the calculations in the test to ensure they are equal to what the anomaly model outputs. 

I could make a note of this, or use the `model.scaler` directly in the calculations?

tests/gordo_components/model/anomaly/test_anomaly_detectors.py
milesgranger: It's more the `model.compile(...)` which takes the time. I can swap it, but it goes from running the 8 tests (parameterized) in ~1 second, to 2 minutes. :laughing:

tests/gordo_components/model/anomaly/test_anomaly_detectors.py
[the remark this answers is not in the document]
epa095: But `model.scaler == scaler` ?

gordo_components/model/anomaly/diff.py
milesgranger: :eagle: :eye:

tests/gordo_components/model/anomaly/test_anomaly_detectors.py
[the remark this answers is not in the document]
milesgranger: Probably. :sweat: `scaler is model.scaler == True`

tests/gordo_components/model/anomaly/test_anomaly_detectors.py
epa095: And that is not just because it is the first time you had to load in tensorflow in that session?

milesgranger: Each test ends up taking 20 some odd seconds. And looking at the slowest tests reports, it's mostly LSTM type tests.

epa095 (approved):
LGTM!!!!

PR #369, opened 2019-07-07
TimeSeriesDataset: support multiple aggregation methods

### LINKED ISSUE #278: Dataset: Add ability to have several resample methods
Today we only resample with mean. I propose that we can not only choose another method (e.g. max), but that we support getting a *list* of resample-methods, and then we use all of them (so returning `n*|tag_list|` columns for `n` methods). That way one can set `resample_methods = ["mean","max","min"]` and one would get out `3*|tag_list|` columns with the different resample methods applied.

epa095:
Now the TimeSeriesDataset supports an optional argument `aggregation_methods`
with default value `mean`. It can be either a single aggregation method or
a list of aggregation methods. Any aggregation method supported by pandas
is supported. If multiple aggregation methods are provided then the resulting
dataframe contains a multi-level column index with the tag-name as the top
level, and the aggregation method as the second level. If a single aggregation
method is provided then only the first level (tag-name) is used.

This closes #278


gordo_components/dataset/base.py
```diff
@@ -111,11 +122,21 @@ def join_timeseries(
             logging.debug(series.head(3))
             logging.debug(series.tail(3))
 
-            resampled = series.resample(resolution, label="left").mean()
+            resampled = series.resample(resolution, label="left").agg(
+                aggregation_methods
+            )
+            # If several aggregation methods are provided, agg gives the column names
+            # like "max" and "mean", so we have to append the original series names to
```
flikka: append should be prepend?

gordo_components/dataset/base.py
```diff
@@ -111,11 +122,21 @@ def join_timeseries(
...
+            if isinstance(resampled, pd.DataFrame):
+                resampled.columns = [
+                    f"{series.name}_{agg_col_name}"
+                    for agg_col_name in resampled.columns
+                ]
             filled = resampled.fillna(method="ffill")
             resampled_series.append(filled)
 
-        new_series = pd.concat(resampled_series, axis=1, join="inner")
+            new_series = pd.concat(resampled_series, axis=1, join="inner")
```
flikka: Oh, did you mean to put this into the for loop???

gordo_components/dataset/base.py
```diff
@@ -111,11 +122,21 @@ def join_timeseries(
...
-            resampled = series.resample(resolution, label="left").mean()
+            resampled = series.resample(resolution, label="left").agg(
+                aggregation_methods
+            )
+            # If several aggregation methods are provided, agg gives the column names
+            # like "max" and "mean", so we have to append the original series names to
+            # the column.
+            if isinstance(resampled, pd.DataFrame):
+                resampled.columns = [
+                    f"{series.name}_{agg_col_name}"
```
flikka: Now I am confused, what happens if you have a single column? Do you not get the aggregation method as the column name? Will the default (old, with mean) path not change the column names?

flikka:
I like the idea, but I think the PR is WIP?

epa095:
I think it is pretty good to go. Except I have not updated the documentation, I will do that as well. 
But, are we good with how it works:
- If there is only a single resample method then the column names are the same as they were given, i.e. "Tag 1", "Tag 2" etc
- If there are several resample methods then the column names becomes "Tag 1_max", "Tag 1_mean", "Tag 2_max", "Tag 2_mean" etc. It is a bit ugly that it is different depending on how many aggregation methods we have, but it is maybe what the user would expect? Also it is backwards-compatible.

epa095:
This PR does also not make sure that it actually works to do this in a workflow. I guess it makes sense to wait for #368 for that. It involves some dirty stuff, especially since the grafana dashboards will probably not work with several aggregation methods.

For now the full workflow works when selecting a single aggregation method (e.g. "max" instead of default "mean"), but the "multiple aggregation methods" is only useful when using the dataset locally.

tests/gordo_components/dataset/test_dataset.py
```diff
@@ -194,6 +225,46 @@ def test_row_filter(self):
...
+        self.assertEqual(
+            list(X.columns),
+            [
+                "Tag 1_mean",
+                "Tag 1_max",
+                "Tag 2_mean",
+                "Tag 2_max",
+                "Tag 3_mean",
+                "Tag 3_max",
+            ],
```
milesgranger, line 237: Does this last part make the first test you wrote redundant?

gordo_components/dataset/datasets.py
```diff
@@ -82,14 +93,31 @@ def get_data(self) -> Tuple[pd.DataFrame, Optional[pd.DataFrame]]:
...
+            aggregation_methods=self.aggregation_methods,
         )
         if self.row_filter:
             data = pandas_filter_rows(data, self.row_filter)
 
         x_tag_names = [tag.name for tag in self.tag_list]
         y_tag_names = [tag.name for tag in self.target_tag_list]
 
+        # If we have multiple aggregation methods then the column-names will be e.g
+        # "tag 1_max", not "tag 1", so we must do some stuff to pick out the right
```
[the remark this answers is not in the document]
epa095: It most definitely does. Or rather, the user must filter on `tag 1_max` instead of `tag 1`.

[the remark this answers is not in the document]
milesgranger: Maybe add to one of the (many) locations of examples? I guess we have a separate repo this example folder and the gordo test project. IMO the [gordo-test-project config](https://github.com/equinor/gordo-test-project/blob/master/kubernetes-deployment/components/config/config.yaml) would be good since that's what we used to "validate" bigger changes and is doing a good job at doubling as an example resource.

tests/gordo_components/dataset/test_dataset.py
epa095: Hmm seems like it. Did I maybe forget that I wrote a test and then basicly rewrite it? IDK, but I will delete the first.  Goodie catchie

milesgranger:
In general, what do you think about using `pd.MultiIndex` for columns so we'd have things like this coming out:
```
                    tag1     tag2    
                    mean sum mean sum
2016-01-01 00:00:00    4  36    4  36
2016-01-01 01:00:00    9   9    9   9
```
or if a single agg:
```
                    tag1 tag2    
                    mean mean
2016-01-01 00:00:00    4    4
2016-01-01 01:00:00    9    9
```

Might result in more complex changes?

epa095:
> In general, what do you think about using `pd.MultiIndex` for columns so we'd have things like this coming out:
> 
> ```
>                     tag1     tag2    
>                     mean sum mean sum
> 2016-01-01 00:00:00    4  36    4  36
> 2016-01-01 01:00:00    9   9    9   9
> ```
> 
> or if a single agg:
> 
> ```
>                     tag1 tag2    
>                     mean mean
> 2016-01-01 00:00:00    4    4
> 2016-01-01 01:00:00    9    9
> ```
> 
> Might result in more complex changes?

I like it. I am a bit uncertain how big of an API change it is. Maybe not that big? At least it has a nicer consistency between the single and multiple aggregation methods.

epa095:
So, now it is as follows: It uses multi-level dataframes when there are several aggregation methods, but the old-style single level ones in case of a single aggregation method. This is to make it backwards compatible, and make it a smaller PR. Two things:
- I must update the notebook (or just ditch that and just add documentation somewhere else. Proposal?)
- Maybe have an explicit tests in e.g. workflow generator which disallows multiple aggregation methods, since it is incompatible with the project at large (grafana particularly)?

milesgranger:
Seems reasonable to me, I've added https://github.com/equinor/gordo-components/issues/382 so we can address using the multi level column dataframes and thus support multiple agg function throughout the workflow later on.

For the example, I think adding it in the https://github.com/equinor/gordo-test-project might be a good place?

gordo_components/dataset/base.py
```diff
@@ -111,7 +122,32 @@ def join_timeseries(
...
+            # For backwards-compatibility we *dont* return a multi-level index
+            # when we have a single resampling method.
+
+            # else:
+            #     resampled_s = resampled.to_frame()
+            #     resampled_s.columns = pd.MultiIndex.from_product(
+            #         [[series.name], [aggregation_methods]],
+            #         names=["tag", "aggregation_method"],
+            #     )
+            #     resampled = resampled_s
```
milesgranger: I'm guessing the PR is implicitly in a WIP state now?

examples/Gordo-Workflow-Semi-Low-Level.ipynb
```diff
@@ -11,7 +11,7 @@
     "---\n",
     "\n",
     "### Import and initialize a Gordo dataset\n",
-    "In this case we shall be using the `DataLakeBackedDataset` where `InfluxBackedDataset` is also available."
+    "In this case we shall be using the `DataLakeProvider` where `InfluxDataProvider` is also available."
```
milesgranger, line 14: Nice house cleaning. :house_with_garden:

gordo_components/dataset/base.py
```diff
@@ -54,11 +55,23 @@ def join_timeseries(
         resolution
             The bucket size for grouping all incoming time data (e.g. "10T")
 
+        aggregation_methods
+            Aggregation method(s) to use for the resampled buckets. If a single
+            resample method is provided then the resulting dataframe will have names
+            identical to the names of the series it got in. If several
+            aggregation-methods are provided then the resulting dataframe will
+            have a multi-level index, with the first level being the series-name
```
milesgranger: Nit-pick alert: :alarm_clock:  I would say `multi-level column index` or something like you say in the `Returns` section below.

gordo_components/dataset/base.py
```diff
@@ -111,7 +124,24 @@ def join_timeseries(
...
+                aggregation_methods
+            )
+            # If several aggregation methods are provided, agg returns a dataframe
+            # instead of a series. In this dataframe the column names are the
+            # aggregation methods, like "max" and "mean", so we have to make a
+            # multi-index with the series-name as the top-level and the
+            # aggregation-method as the lower-level index
+            if isinstance(
+                resampled, pd.DataFrame
+            ):  # Several aggregation methods provided
```
milesgranger, line 139: Nit-pick alert: Could save a couple lines by putting this comment either above or below the `if`

gordo_components/dataset/base.py
```diff
@@ -111,7 +124,24 @@ def join_timeseries(
...
+                resampled.columns = pd.MultiIndex.from_product(
+                    [[series.name], resampled.columns],
+                    names=["tag", "aggregation_method"],
+                )
+            # For backwards-compatibility we *dont* return a multi-level index
+            # when we have a single resampling method.
+
```
milesgranger: A little experiment in the terminal seems if there is a multi-index column dataframe, `pd.concat` will, if done with at least one series, will squash mutli-index names into tuples:
```
        (input, col1)  (input, col2)  (input, col3)  (input, col4)  ...  (transformed, col2)  (transformed, col3)  (transformed, col4)  0
0            0.0            1.0            2.0            3.0  ...                  1.5                  3.0                  4.5  0
1            4.0            5.0            6.0            7.0  ...                  7.5                  9.0                 10.5  1
```

I guess this is ok then? I don't actually know, just clarifying.

However, if all are multi-level which I guess which would typically happen, there is no backwards-compatibility because then it does keep the multi-level structure.

epa095: Hmm. They will all either be single-level or multi-level, since we don't support per-series aggregation method selection, they all either have a single method or all have multiple methods. We could of course could support different aggregation methods, I just did not think about it at all :-p But that can be for the future, just expanding the `aggregation_methods` parameter to also accept a `dict` from tag to resample method(s).

milesgranger: I think it's cool as it is now, just adding the check in infrastructure, then it never actually ends up being the multi-level, and we can tackle that later; but I like the idea that this spawned. :+1:

milesgranger (approved):
Rebase, mahn.

PR #370, opened 2019-07-09
Generalization of fields read by IROC reader

### LINKED ISSUE #169: IROC datasource: support more tags than NINENINE
Currently we only support NINENINE tags, as seen by the regexp and the basepath. But there are other fields in `/raw/plant/uon/cygnet`, e.g. `UON_EF` is in `/raw/plant/uon/cygnet/efd/history/`. A map with regexps to base-paths just as with the NCS stuff would solve it. 

https://github.com/equinor/gordo-components/blob/78203b2db3a79b7d1e96cdc69feb45b810ca5d58
/gordo_components/data_provider/iroc_reader.py#L19

Lezinou:
Duplicated the base_path_to_asset function present in Ncs_reader to the IrocReader.
Introduced new tag : path for eagle ford (uon_ef)
Added some checks in case of multiple/no asset found.

Close #169

### DIFF
--- a/gordo_components/data_provider/iroc_reader.py
+++ b/gordo_components/data_provider/iroc_reader.py
@@ -17,10 +17,13 @@ logger = logging.getLogger(__name__)
 
 
 class IrocReader(GordoBaseDataProvider):
-    ASSET_TO_PATH = {"ninenine": "/raw/plant/uon/cygnet/ninenine/history"}
+    ASSET_TO_PATH = {
+        "ninenine": "/raw/plant/uon/cygnet/ninenine/history",
+        "uon_ef": "/raw/plant/uon/cygnet/efd/history",
+    }
 
-    def can_handle_tag(self, tag):
-        return tag.asset in IrocReader.ASSET_TO_PATH
+    def can_handle_tag(self, tag: SensorTag):
+        return IrocReader.base_path_from_asset(tag.asset) is not None
 
     def __init__(self, client: core.AzureDLFileSystem, threads: int = 50, **kwargs):
         """
@@ -38,7 +41,6 @@ class IrocReader(GordoBaseDataProvider):
         to_ts: datetime,
         tag_list: List[SensorTag],
         dry_run: Optional[bool] = False,
-        base_path="raw/plant/uon/cygnet/ninenine/history",
     ):
         """
         See GordoBaseDataProvider for documentation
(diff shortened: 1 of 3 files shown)


flikka:
Thanks - I will look at it! Tests are failing, maybe you'd want too look at that, but I'll have a look myself. A point: It is preferred to make PRs from your own fork, and not create new branches on the upstream repository.

gordo_components/data_provider/iroc_reader.py
```diff
@@ -53,7 +55,21 @@ def load_series(
...
+                "Iroc reader found more than one asset from the tag list provided, returning none."
+            )
+            return
+        elif len(base_paths_from_assets) != len(tag_list):
+            logger.warning(
+                "Iroc reader could not associate some tags to an asset, returning none."
+            )
+            return
+
+        base_path = base_paths_from_assets[0]
```
flikka, line 72: You also need to handle the "None" case - where the resolved paths are "None". That would go through as it is now.

tests/gordo_components/data_provider/test_data_provider_iroc.py
```diff
@@ -105,6 +122,32 @@ def test_load_series_no_tag_list(self):
         )
         self.assertEqual([], res)
 
+    def test_load_series_many_assets(self, _mocked_method):
```
flikka, line 125: The _mocked_method require the use of mocking fixtures. I don't think you need it here, however, as these tests should cause an early return, before the reading of files are started.

tests/gordo_components/data_provider/test_data_provider_iroc.py
```diff
@@ -105,6 +122,32 @@ def test_load_series_no_tag_list(self):
...
+        res = list(
+            iroc_reader.load_series(
+                from_ts=isoparse("2018-05-02T01:56:00+00:00"),
+                to_ts=isoparse("2018-05-03T01:56:00+00:00"),
+                tag_list=IROC_MANY_ASSETS_TAG_LIST,
+            )
+        )
+        self.assertEqual([], res)
+
+    def test_load_series_no_asset_found(self, _mocked_method):
```
flikka, line 138: Same as above

tests/gordo_components/data_provider/test_data_provider_iroc.py
```diff
@@ -105,6 +122,32 @@ def test_load_series_no_tag_list(self):
...
+        iroc_reader = IrocReader(client=None, threads=1)
+        res = list(
+            iroc_reader.load_series(
+                from_ts=isoparse("2018-05-02T01:56:00+00:00"),
+                to_ts=isoparse("2018-05-03T01:56:00+00:00"),
+                tag_list=IROC_NO_ASSET_TAG_LIST,
+            )
+        )
+        self.assertEqual([], res)
+
```
flikka, line 150: My preference would actually be that these tests caused an exception to be raised - since it is non-supported input, and we might as well fail hard.

flikka (changes requested):
Good thinking here, fix up the things that don't work, and we can be ready to go! Also, we should add an IROC example to the gordo-test-project perhaps? https://github.com/equinor/gordo-test-project/blob/master/kubernetes-deployment/components/config/config.yaml 

But this we can do later - but nice to have some running example of this.

tests/gordo_components/data_provider/test_data_provider_iroc.py
```diff
@@ -16,6 +16,23 @@
 HAPPY_FROM_TS = isoparse("2018-05-02T01:56:00+00:00")
 HAPPY_TO_TS = isoparse("2018-05-03T01:56:00+00:00")
 
+IROC_MANY_ASSETS_TAG_LIST = [
```
flikka, line 19: Okey! So something more came to my attention here. The testing you do here, sends in tags with asset=None. That will always fail in your code (as in never provide a path). If you look in the test_ncs_reader.py you see that we send in assets explitly, or use the "normalize_sensor_tags" in the sensor_tag.py file. This must be done here too, to make sense.

flikka (changes requested):
So, some more candy here: We must (should) test the "can_handle_tag" function too, with several examples, with missing asset, with unknown asset, and with known assets.

PR #375, opened 2019-07-11
Generalize IROC reader (many fields)

Lezinou:


gordo_components/data_provider/iroc_reader.py
```diff
@@ -59,14 +59,12 @@ def load_series(
             map(lambda tag: IrocReader.base_path_from_asset(tag.asset), tag_list)
         )
         if len(set(base_paths_from_assets)) > 1:
-            logger.warning(
-                "Iroc reader found more than one asset from the tag list provided, returning none."
+            raise ValueError(
```
flikka, line 62: Also log error, before raising the error

gordo_components/data_provider/iroc_reader.py
```diff
@@ -59,14 +59,12 @@ def load_series(
...
+                "Iroc reader found more than one asset from the tag list provided"
             )
             return
```
flikka: The return can never be reached, remove it

gordo_components/data_provider/iroc_reader.py
```diff
@@ -59,14 +59,12 @@ def load_series(
...
-        elif len(base_paths_from_assets) != len(tag_list):
-            logger.warning(
-                "Iroc reader could not associate some tags to an asset, returning none."
-            )
+        elif any([path is None for path in base_paths_from_assets]):
+            raise ValueError("Iroc reader could not associate some tags to an asset.")
             return
```
flikka, line 67: Same as above

flikka (dismissed):
Some comments added
In addition:
* You have added a notebook, I assume that is by mistake
* The title of the PR and the commit message is BAD :-)

flikka:
The failing test is for the notebook, just remove this from the PR

gordo_components/data_provider/iroc_reader.py
```diff
@@ -59,14 +59,12 @@ def load_series(
...
-                "Iroc reader found more than one asset from the tag list provided, returning none."
+            raise ValueError(
+                "Iroc reader found more than one asset from the tag list provided"
             )
             return
-        elif len(base_paths_from_assets) != len(tag_list):
-            logger.warning(
-                "Iroc reader could not associate some tags to an asset, returning none."
-            )
+        elif any([path is None for path in base_paths_from_assets]):
```
flikka: Here you can rather write `elif None in base_paths_from_assets:`

gordo_components/data_provider/iroc_reader.py
```diff
@@ -59,14 +59,12 @@ def load_series(
             map(lambda tag: IrocReader.base_path_from_asset(tag.asset), tag_list)
         )
         if len(set(base_paths_from_assets)) > 1:
```
flikka: Safer to check if the len != 1 perhaps? If the len of the set is 0 (zero) is that ok?

Lezinou: For me, the case len = 0 does not exist because the base_path_from_asset will always return something (won't it?), so this check is only for the case where more than 1 asset is found. The next check is supposed to handle when no asset has been found for some tags.
I might be mistaking of course :)

gordo_components/data_provider/iroc_reader.py
Lezinou: Not sure I understand, I did the same as for the time check (line 53-56)!

gordo_components/data_provider/iroc_reader.py
Lezinou: Done!

gordo_components/data_provider/iroc_reader.py
Lezinou: Done!

gordo_components/data_provider/iroc_reader.py
flikka: I see. Just thought that it was nice to do a log message first, and then raise. But now am not sure what is the best practice. Just leave it for now, no problem.

gordo_components/data_provider/iroc_reader.py
flikka: In a way yes - but if for example the tag list was empty, I guess base_path_from_asset would return an empty list? And the next test (testing equal lenghts) would also pass. 

In the code today, we have already tested that the tag list is not empty, but that is a bit risky to assume, someone might change the lines above. I don't see what allowing for a lenght==0 gives here, it seems to me that the most precise thing to test in this "if" is is if the set is exactly lenght==1.

flikka:
Better and better :-) Still - the notebook change don't belong in this PR as far as I see. I tried to look at the diff, and I can't see that there are any relevant changes in the diff for the notebook. Could you please remove the notebook change from the PR? Merci!

gordo_components/data_provider/iroc_reader.py
Lezinou: Understood, I was taking for granted that the rest of the checks would not be changed but the world is a nasty place, so this is a fair point :)

flikka (approved):
Looks good. But please use the "Squash option" and change to one commit, with commit message that describes the "main purpose".

PR #377, opened 2019-07-22
Add EV, R2, MSE and MAE to metadata 

### LINKED ISSUE #360: Support other metrics for hyperparameter tuning in Katib
Currently, in the yaml file used to generate a Katib workflow, we support
- explained-variance_mean as the _objectivevaluename_  and _metricsnames_
- explained-variance_std, explained-variance_max and explained-variance_mean as _metricsnames_

These metrics are aggregates over all sensors and validation splits.

1.  
It would be useful to support other metrics such as:
- mean squared error
- R2 score
- mean absolute error

2. 
It would also be useful to output metrics for each sensor and each validation split in addition to aggregated values.

Natalie-Caruana:
This will close #360 and #401 

Add EV, R2, MSE and MAE 
 - for each sensor and each validation split
-  aggregated (mean) scores across sensors for each validation split
-  aggregated scores across sensors and validation splits

to metadata.

### DIFF
--- a/gordo_components/builder/build_model.py
+++ b/gordo_components/builder/build_model.py
@@ -12,8 +12,14 @@ import pandas as pd
 import numpy as np
 
 from sklearn.base import BaseEstimator
-from sklearn.model_selection import cross_val_score, TimeSeriesSplit
-from sklearn.metrics import explained_variance_score, make_scorer
+from sklearn.metrics import (
+    explained_variance_score,
+    make_scorer,
+    r2_score,
+    mean_squared_error,
+    mean_absolute_error,
+)
+from sklearn.model_selection import cross_validate, TimeSeriesSplit
 from sklearn.pipeline import Pipeline
 
 from gordo_components.util import disk_registry
@@ -22,7 +28,6 @@ from gordo_components.dataset.dataset import _get_dataset
 from gordo_components.dataset.base import GordoBaseDataset
 from gordo_components.model.base import GordoBase
 from gordo_components.model.utils import metric_wrapper
-
 
 logger = logging.getLogger(__name__)
 
@@ -85,34 +90,45 @@ def build_model(
     cv_duration_sec = None
 
     if cv_mode.lower() in ("cross_val_only", "full_build"):
-
+        metrics_list = [
+            explained_variance_score,
+            r2_score,
+            mean_squared_error,
+            mean_absolute_error,
+        ]
         # Cross validate
         logger.debug("Starting cross validation")
         start = time.time()
-
-        scores: Dict[str, Any]
-
+        scores: Dict[str, Any] = dict()
         if hasattr(model, "predict"):
 
-            cv_scores = cross_val_score(
+            metrics_dict = get_metrics_dict(metrics_list, y)
+
+            cv = cross_validate(
                 model,
                 X,
                 y,
-                scoring=make_scorer(metric_wrapper(explained_variance_score)),
+                scoring=metrics_dict,
+                return_estimator=True,
                 cv=TimeSeriesSplit(n_splits=3),
             )
-
-            scores = {
-                "explained-variance": {
-                    "mean": cv_scores.mean(),
-                    "std": cv_scores.std(),
-                    "max": cv_scores.max(),
-                    "min": cv_scores.min(),
-                    "raw-scores": cv_scores.tolist(),
+            for metric, test_metric in map(lambda k: (k, f"test_{k}"), metrics_dict):
+                val = {
+                    "fold-mean": cv[test_metric].mean(),
+                    "fold-std": cv[test_metric].std(),
+                    "fold-max": cv[test_metric].max(),
+                    "fold-min": cv[test_metric].min(),
                 }
-            }
+                val.update(
+                    {
+                        f"fold-{i + 1}": raw_value
+                        for i, raw_value in enumerate(cv[test_metric].tolist())
+                    }
+                )
+                scores.update({metric: val})
+
         else:
-            logger.debug("Unable to score model, has no attribute 'score'.")
+            logger.debug("Unable to score model, has no attribute 'predict'.")
             scores = dict()
 
         cv_duration_sec = time.time() - start
@@ -150,8 +166,61 @@ def build_model(
     }
 
     metadata["model"].update(_get_metadata(model))
-
     return model, metadata
+
+
+def get_metrics_dict(metrics_list: list, y: pd.DataFrame) -> dict:
+
+    """
+    Given a list of metrics that accept a true_y and pred_y as inputs this returns
+    a dictionary with keys in the form '{score}-{tag_name}' for each given target tag
+    and '{score}' for the average score across all target tags and folds,
+    and values being the callable make_scorer(metric_wrapper(score)).
+    Note: score in {score}-{tag_name} is a sklearn's score function name with '_' replaced by '-' and tag_name
+    corresponds to given target tag name with ' ' replaced by '-'.
+
+    Parameters
+    ----------
+    metrics: list
+        List of sklearn score functions
+    y: pd.DataFrame
+        Target data
+
+
+    Returns
+    -------
+        dict
+    """
+
+    def _score_factory(metric=r2_score, index=0):
+        def _score_per_tag(y_true, y_pred):
+            # This function extracts the score for each given target_tag to
+            # use as scoring argument in sklearn cross_validate, as the scoring
+            # must return a single value.
+            if hasattr(y_true, "values"):
+                y_true = y_true.values
+            if hasattr(y_pred, "values"):
+                y_pred = y_pred.values
+
+            return metric(y_true[:, index], y_pred[:, index])
+
+        return _score_per_tag
+
+    metrics_dict = {}
+    for metric in metrics_list:
+        for index, col in enumerate(y.columns):
+            metric_str = metric.__name__.replace("_", "-")
+            metrics_dict.update(
+                {
+                    metric_str
+                    + f'-{col.replace(" ", "-")}': make_scorer(
+                        metric_wrapper(_score_factory(metric=metric, index=index))
+                    )
+                }
+            )
+
+        metrics_dict.update({metric_str: make_scorer(metric_wrapper(metric))})
+    return metrics_dict
 
 
 def _determine_offset(model: BaseEstimator, X: Union[np.ndarray, pd.DataFrame]) -> int:
--- a/gordo_components/cli/cli.py
+++ b/gordo_components/cli/cli.py
@@ -329,8 +329,22 @@ def expand_model(model_config: str, model_parameters: dict):
 
 def get_all_score_strings(metadata):
     """Given metadata from the model builder this function returns a list of
-    strings of the format {metric_name}_{score_name}={score_val} for katib
-    to pick up. Replaces all spaces with `-`.
+    strings of the following format:
+    {metric_name}-{tag_name}_{fold-fold-number} = {score_val}.  This computes the score for the given tag and
+    cross validation split.
+    {metric_name}-{tag_name}_{fold-aggregation} = {score_val}. This computes the score for the given tag and aggregates
+    the score over all cross validation splits (aggregations currently used are mean, std, min and max)
+    {metric_name}_{fold-fold-number} = {score_val}.  This computes the score aggregate across all tags (uses sklearn's default
+    aggregation method) for a given cross validation split.
+    {metric_name}_{fold-aggregation} = {score_val}.  This computes the score aggregate across all tags (uses sklearn's default
+    aggregation method) and cross validation splits (aggregations currently used are mean, std, min and max).
+
+    for katib to pick up.
+
+    Current metric names supported are sklearn score functions: 'r2_score', 'explained_variance_score',
+    'mean_squared_error' and 'mean_absolute_error'.  The underscores in such score names are replaced by '-'.
+
+    All spaces in the tag name are also replaced by '-'.
 
     Parameters
     ----------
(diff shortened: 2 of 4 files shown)


gordo_components/builder/build_model.py
```diff
@@ -75,16 +77,37 @@ def build_model(
...
+                "min": cv["test_score"].min(),
+                "raw-scores": cv["test_score"].tolist(),
             }
         }
+
+        # adding loss function to scores if Gordo model is a KerasBaseEstimator
+        gordo_model = _get_final_gordo_base_step(model)
+        if issubclass(gordo_model.__class__, KerasBaseEstimator):
+            val_losses = np.array([])
+            for mod in cv["estimator"]:
```
epa095: So this loops over each of the trained models? You want the loss for all of them? And does it make sense to have the mean/std, etc? If those values are not meaningful for anything really then it is better to not add them.

gordo_components/builder/build_model.py
```diff
@@ -75,16 +77,37 @@ def build_model(
...
+                gordo_model_cv = _get_final_gordo_base_step(mod)
+                val_loss = gordo_model_cv.get_metadata()["history"]["loss"][0]
```
epa095: What does the `[0]` index into? The epochs?`If so, do you only want the loss for the first epoch?

gordo_components/builder/build_model.py
```diff
@@ -75,16 +77,37 @@ def build_model(
...
+                val_losses = np.append(val_losses, val_loss)
+            scores["cv-loss"] = {
```
epa095: Seems like this assignment should be after the for-loop.

gordo_components/builder/build_model.py
```diff
@@ -75,16 +77,37 @@ def build_model(
...
+                "mean": val_losses.mean(),
+                "std": val_losses.std(),
+                "max": val_losses.max(),
+                "min": val_losses.min(),
+                "raw-scores": val_losses.tolist(),
```
epa095: This can become quite big, no?

gordo_components/builder/build_model.py
Natalie-Caruana: That's a mistake...should read [-1] so the loss from last epoch is extracted.

gordo_components/builder/build_model.py
Natalie-Caruana: i think it makes sense to have the mean/std/min/max especially if we decide to increase the number of splits as it would give us an idea of the variability of the validation losses for the different splits.

ryanjdillon:
Is this resolving #401?

gordo_components/builder/build_model.py
```diff
@@ -89,30 +94,66 @@ def build_model(
...
         logger.debug("Starting cross validation")
         start = time.time()
-
-        scores: Dict[str, Any]
-
+        scores: Dict[str, Any] = dict()
         if hasattr(model, "predict"):
-
-            cv_scores = cross_val_score(
+            metrics_list = [
```
epa095: Can maybe this whole thing be extracted to a function?

gordo_components/builder/build_model.py
```diff
@@ -89,30 +94,66 @@ def build_model(
...
+                explained_variance_score,
+                r2_score,
+                mean_squared_error,
+                mean_absolute_error,
+            ]
+            metrics_dict = {}
+            for metric in metrics_list:
+                for index, col in enumerate(y.columns):
+
+                    def my_custom_score(y_true, y_pred, metric=metric, index=index):
```
epa095: And this thing, can it be moved out, and given a better name? And maybe even a docstring. But it can be hidden (starting with a underscore)

Natalie-Caruana: I tried to remove this function out of the for loops and create a wrapper, something like this:

```            def my_custom_score(y_true, y_pred, metric=r2_score, index=0):
                score = metric(y_true.values[:, index], y_pred[:, index])
                return score
```
and then in the for loop use a lambda function:
```    for metric in metrics_list:
        for index, col in enumerate(y.columns):
            metric_str = metric.__name__
            metrics_dict.update(
                {
                    metric_str.replace("_", "-")
                    + f'-{col.replace(" ", "-")}': make_scorer(
                        metric_wrapper(lambda y_true,y_pred: my_custom_score(y_true,y_pred,metric=metric,index=index)
                    )
                }
            )
```
but im getting the same results for different metrics/indexes...it looks like (since the lambda function depends only on y_true and y_pred that the metric and index are not being iterated...

gordo_components/cli/cli.py
```diff
@@ -345,16 +345,21 @@ def get_all_score_strings(metadata):
...
     ...     "model": {
     ...         "cross-validation": {
-    ...             "scores": {"explained variance": {"min": 0, "max": 2}}
-    ...         }
+    ...             "scores": {"explained variance tag 0": {"fold_1": 0, "fold_2": 0.9, "fold_3": 0.1,"min": 0.1, "max": 0.9, "mean": 1/3},
+    ...                        "explained variance tag 1": {"fold_1": 0.2, "fold_2": 0.3, "fold_3": 0.6, "min": 0.2, "max": 0.6, "mean": 1.1/3},
+    ...                        "explained variance mean" : {"min": 0.2/2, "max": 1.2/2, "mean": (1/3)*(0.2/2 + 1.2/2 + 0.7/2)},
+    ...                        "r2 score tag 0" : {"fold_1": 0.8, "fold_2": 0.5, "fold_3": 0.6, "min": 0.5, "max": 0.8, "mean": 1.9/3},
+    ...                        "r2 score tag 1" : {"fold_1": 0.4, "fold_2": 0.3, "fold_3": 0.5, "min": 0.3, "max": 0.5, "mean": 1.2/3},
+    ...                        "r2 score mean"  : {"min": 0.8/2,"max": 1.2/2, "mean": (1/3)*(1.2/2 + 0.8/2 + 1.1/2)}
```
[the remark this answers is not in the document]
Natalie-Caruana: As discussed my aim was to show what it was actually being computed...that it basically averages both over folds and tags.  But i will explain this in a comment as suggested

gordo_components/builder/build_model.py
```diff
@@ -85,34 +90,48 @@ def build_model(
...
+                    "fold-mean": cv[f"test_{key}"].mean(),
+                    "fold-std": cv[f"test_{key}"].std(),
+                    "fold-max": cv[f"test_{key}"].max(),
+                    "fold-min": cv[f"test_{key}"].min(),
                 }
-            }
+                val.update(
+                    {
+                        f"fold-{i + 1}": raw_value
+                        for i, raw_value in enumerate(cv[f"test_{key}"].tolist())
```
milesgranger: instead of all these `f"test_{key}"` , maybe create a variable, or update `key` in a map during the for loop. ie.
```python
for key in map(lambda k: f"test_{k}", metrics_dict):
```

gordo_components/builder/build_model.py
```diff
@@ -85,34 +90,48 @@ def build_model(
...
+                    }
+                )
+                scores.update({key: val})
+
         else:
-            logger.debug("Unable to score model, has no attribute 'score'.")
+            logger.debug("Unable to score model, has no attribute 'predict'.")
```
milesgranger, line 131: Nice catchy! :+1:

gordo_components/builder/build_model.py
```diff
@@ -150,10 +169,62 @@ def build_model(
...
 
     metadata["model"].update(_get_metadata(model))
-
     return model, metadata
 
 
+def get_metrics_dict(metrics_list, y):
+    """
+        Given a list of metrics that accept a true_y and pred_y as inputs this returns
+        a dictionary with keys in the form '{score}-{tag_name}' for each given target tag
```
milesgranger: _very_ mild, but you have an extra indent here in the docs.

Also, as we talked, maybe it is more generic to call things `feature(s)` rather than tags? I'm not terribly concerned about this though...

gordo_components/builder/build_model.py
```diff
@@ -150,10 +169,62 @@ def build_model(
...
+            Target data
+
+
+        Returns
+        -------
+            dict
+        """
+
+    def _score_factory(metric=r2_score, index=0):
+        def _score_per_tag(y_true, y_pred):
```
milesgranger, line 196: Nice and clean!

gordo_components/builder/build_model.py
```diff
@@ -150,10 +169,62 @@ def build_model(
...
+            # This function extracts the score for each given target_tag to
+            # use as scoring argument in sklearn cross_validate, as the scoring
+            # must return a single value.
+            return metric(y_true.values[:, index], y_pred[:, index])
```
milesgranger: Seems like we're making some fairly inflexible assumptions here.
1. That `y_true` _will_ be a dataframe
2. That `y_pred` _will_ be a numpy array

Perhaps we can accept either for both and do some `hasattr(..., 'values')` or something similar to be a bit more flexible?

gordo_components/builder/build_model.py
```diff
@@ -150,10 +169,62 @@ def build_model(
...
+                    + f'-{col.replace(" ", "-")}': make_scorer(
+                        metric_wrapper(_score_factory(metric=metric, index=index))
+                    )
+                }
+            )
+
+        metrics_dict.update(
+            {
+                metric_str.replace("_", "-")
+                + f"-tag-mean": make_scorer(metric_wrapper(metric))
```
milesgranger: here i think we talked about removing 'tag' maybe?

tests/gordo_components/builder/test_builder.py
```diff
@@ -50,7 +50,39 @@ def metadata_check(metadata, check_history):
...
+            for tag in metadata["dataset"]["tag_list"]
+        ]
+        assert (
+            "explained-variance-score-tag-mean"
+            in metadata["model"]["cross-validation"]["scores"]
+        )
+        assert [
+            f"mean-squared-error-{tag.name.replace(' ','-')}"
+            in metadata["model"]["cross-validation"]["scores"]
+            for tag in metadata["dataset"]["tag_list"]
```
milesgranger: seems like lots of `metadata["dataset"]["tag_list"]` followed by `tag.name.replace(...)` maybe at the start of this conditional, do the preprocessing into `tag_list` and probably `scores` from `metadata["model"]["cross-validation"]["scores"]` 

That might actually save you a bunch of lines from the formatting.

tests/gordo_components/builder/test_builder.py
```diff
@@ -262,6 +294,174 @@ def test_get_metadata_helper(model: BaseEstimator, expect_empty_dict: bool):
...
+    model3, metadata3 = build_model(
+        name="model-name",
+        model_config=model_config3,
+        data_config=data_config,
+        metadata={},
+    )
+
+    metadata_check(metadata1, False)
+    metadata_check(metadata2, False)
+    metadata_check(metadata3, False)
```
milesgranger: This test, specifically the `raw_model` configs, should be parameterized `pytest.mark.parameterize`

tests/gordo_components/builder/test_builder.py
```diff
@@ -262,6 +294,174 @@ def test_get_metadata_helper(model: BaseEstimator, expect_empty_dict: bool):
...
+
+    assert (
+        metadata["model"]["cross-validation"]["scores"]["mean-absolute-error-Tag-1"][
+            "fold-mean"
+        ]
+        + metadata["model"]["cross-validation"]["scores"]["mean-absolute-error-Tag-2"][
+            "fold-mean"
+        ]
+    ) / 2 == pytest.approx(
+        metadata["model"]["cross-validation"]["scores"]["mean-absolute-error-tag-mean"][
```
milesgranger: again for readability, creating a variable `scores = metadata["model"]["cross-validation"]["scores"]` would save on readability in this test.

milesgranger (commented):
Sorry for the comments, mostly small things I think. Looks nice in general. :+1:

gordo_components/builder/build_model.py
```diff
@@ -217,10 +219,7 @@ def _score_per_tag(y_true, y_pred):
             )
 
         metrics_dict.update(
-            {
-                metric_str.replace("_", "-")
-                + f"-tag-mean": make_scorer(metric_wrapper(metric))
-            }
+            {metric_str.replace("_", "-"): make_scorer(metric_wrapper(metric))}
```
milesgranger: I think everywhere you use `metric_str` you do this `.replace(...)` maybe just do that at the assignment 
`metric_str = metric.__name__.replace('_', '-')`

tests/gordo_components/builder/test_builder.py
```diff
@@ -50,39 +50,22 @@ def metadata_check(metadata, check_history):
...
-        assert [
-            f"mean-absolute-error-{tag.name.replace(' ','-')}"
-            in metadata["model"]["cross-validation"]["scores"]
-            for tag in metadata["dataset"]["tag_list"]
-        ]
-        assert (
-            "mean-absolute-error-tag-mean"
-            in metadata["model"]["cross-validation"]["scores"]
-        )
+        all_scores_list = [
```
milesgranger, line 62: Nice! :100:

gordo_components/builder/build_model.py
```diff
@@ -150,10 +166,64 @@ def build_model(
     }
 
     metadata["model"].update(_get_metadata(model))
-
     return model, metadata
 
 
+def get_metrics_dict(metrics_list, y):
```
milesgranger: type defs

gordo_components/builder/build_model.py
```diff
@@ -150,10 +166,63 @@ def build_model(
...
+
+    def _score_factory(metric=r2_score, index=0):
+        def _score_per_tag(y_true, y_pred):
+            # This function extracts the score for each given target_tag to
+            # use as scoring argument in sklearn cross_validate, as the scoring
+            # must return a single value.
+            if hasattr(y_true, "values"):
+                y_true = y_true.values
+            if hasattr(y_pred, "values"):
+                y_true = y_pred.values
```
milesgranger: Here's a fun bug...

Natalie-Caruana: grrrrr :angry:

Natalie-Caruana: angry at me hehe

milesgranger:
@Natalie-Caruana 
I re-ran the workflow you submitted earlier, as all the models were cached because their config didn't actually change. Anyhow, I can confirm it works! Nice job!

image

PR #380, opened 2019-07-31
Fix implicit ordering of columns in server base

### LINKED ISSUE #379: Tags are being reordered
This began after optimizing the server's conversion of a multi-level column dataframe into a JSON serializable `dict` object in https://github.com/equinor/gordo-components/commit/3da81734a0b0bc9234dd67a4e86cd3a2e1fce590

Essentially, calling `.stack()` on such a dataframe will order the second level columns. Therefore the resulting records do not match the ordering of the original features/tags for the top level feature name (such as `model-input`)

milesgranger:
Will close #379 

- [X] Fix reordering problem
- [x] Change the tags used in tests not to be alphabetical (the reason existing tests didn't catch this)

### DIFF
--- a/gordo_components/server/views/base.py
+++ b/gordo_components/server/views/base.py
@@ -320,8 +320,18 @@ class BaseModelView(Resource):
             # Stack the dataframe so second level column names become second level indexs
             df.stack()
             # For each column now, unstack the previous second level names (which are now the indexes of the series)
-            # back into a dataframe with those names, and convert to list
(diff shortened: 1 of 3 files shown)


epa095 (commented):
What about a test which would have catched the issue?

milesgranger:
That was added to/discovered with the existing `test_ml_server_dataframe_to_dict_and_back` test.

epa095 (approved):
LGTM!

PR #394, opened 2019-08-09
Use yaml.safe_load for json like parsing in CLI

### LINKED ISSUE #390: Use yaml/json instead of `literal_eval`
In `gordo-components/gordo_components/cli/cli.py` it seems a bit random what is parsed as json and what is parsed with `literal_eval`. Maybe just use yaml for all (since json is a subset of yaml, yaml should be the most accepting one)? What U think @milesgranger ?

milesgranger:
Will close #390

### DIFF
--- a/gordo_components/cli/cli.py
+++ b/gordo_components/cli/cli.py
@@ -5,9 +5,7 @@ CLI interfaces
 """
 
 import logging
-import ast
 import os
-from ast import literal_eval
 
 import jinja2
 import yaml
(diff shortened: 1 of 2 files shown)


epa095 (commented):
LGTM, is it even tested on a real workflow or are we trusting our tests?

milesgranger:
It is tested on some inputs, and some terminal experimentation shows identical functionality with `literal_eval`

Also, I'd guess we'll run it through the gauntlet before the next release anyhow.

epa095:
Yeah. I see that in the argo workflow we do .to_dict, and not exporting to json/yaml. But that is an issue in infrastructure which should then be fixed before we make a release there with a new components. Feel free to merge.

epa095 (approved):

milesgranger:
Can you refer to the line/issue number. I'm not sure what/why calling `to_dict` would be an issue, That ends up being _almost_ valid json (if not using double quotes that is) but will work fine with `yaml`

PR #395, opened 2019-08-09
Model builder - Support a pure CV/HP mode

### LINKED ISSUE #391: Model builder: Support a pure CV/HP mode.
When doing hyperparameter tuning today we are using `gordo-components build` which outputs CV scores, and then builds a final model it stores on disk. But for HP-tuning the final model is not used at all, only the CV-scores. And training the final model takes significant amount of time, since it uses the whole dataset. I propose we make a toggle to `build` which makes it just do the CV and print it, but not build (and store) the final model.

EinarJohnsen:
Added support to specify a mode for getting cv_scores without building the entire model. Will close #391

### DIFF
--- a/gordo_components/builder/build_model.py
+++ b/gordo_components/builder/build_model.py
@@ -6,7 +6,7 @@ import logging
 import os
 import time
 from pathlib import Path
-from typing import Union, Optional, Dict, Any
+from typing import Union, Optional, Dict, Any, Tuple
 
 import pandas as pd
 import numpy as np
@@ -32,7 +32,8 @@ def build_model(
     model_config: dict,
     data_config: Union[GordoBaseDataset, dict],
     metadata: dict,
-):
+    cv_mode: str = "full_build",
+) -> Tuple[Union[BaseEstimator, None], dict]:
     """
     Build a model and serialize to a directory for later serving.
 
@@ -51,10 +52,15 @@ def build_model(
         Mapping of the Dataset to initialize, following the same logic as model_config.
     metadata: dict
         Mapping of arbitrary metadata data.
+    cv_mode: str
+        String which enables three different modes:
+        * cross_val_only: Only perform cross validation
+        * build_only: Skip cross validation and only build the model
+        * full_build: Cross validation and full build of the model, default value
 
     Returns
     -------
-        Tuple[sklearn.base.BaseEstimator, dict]
+        Tuple[Optional[sklearn.base.BaseEstimator], dict]
     """
     # Get the dataset from config
     logger.debug(f"Initializing Dataset with config {data_config}")
@@ -76,34 +82,53 @@ def build_model(
     logger.debug(f"Initializing Model with config: {model_config}")
     model = serializer.pipeline_from_definition(model_config)
 
-    # Cross validate
-    logger.debug(f"Starting to do cross validation")
-    start = time.time()
+    cv_duration_sec = None
 
-    scores: Dict[str, Any]
-    if hasattr(model, "predict"):
-        cv_scores = cross_val_score(
-            model,
-            X,
-            y,
-            scoring=make_scorer(metric_wrapper(explained_variance_score)),
-            cv=TimeSeriesSplit(n_splits=3),
-        )
-        scores = {
-            "explained-variance": {
-                "mean": cv_scores.mean(),
-                "std": cv_scores.std(),
-                "max": cv_scores.max(),
-                "min": cv_scores.min(),
-                "raw-scores": cv_scores.tolist(),
+    if cv_mode.lower() in ("cross_val_only", "full_build"):
+
+        # Cross validate
+        logger.debug("Starting cross validation")
+        start = time.time()
+
+        scores: Dict[str, Any]
+
+        if hasattr(model, "predict"):
+
+            cv_scores = cross_val_score(
+                model,
+                X,
+                y,
+                scoring=make_scorer(metric_wrapper(explained_variance_score)),
+                cv=TimeSeriesSplit(n_splits=3),
+            )
+
+            scores = {
+                "explained-variance": {
+                    "mean": cv_scores.mean(),
+                    "std": cv_scores.std(),
+                    "max": cv_scores.max(),
+                    "min": cv_scores.min(),
+                    "raw-scores": cv_scores.tolist(),
+                }
             }
-        }
+        else:
+            logger.debug("Unable to score model, has no attribute 'score'.")
+            scores = dict()
+
+        cv_duration_sec = time.time() - start
+
+        # If cross_val_only, return the cv_scores and empty model.
+        if cv_mode == "cross_val_only":
+            metadata["model"] = {
+                "cross-validation": {
+                    "cv-duration-sec": cv_duration_sec,
+                    "scores": scores,
+                }
+            }
+            return None, metadata
     else:
-        logger.debug("Unable to score model, has no attribute 'score'.")
+        # Setting cv scores to zero when not used.
         scores = dict()
-
-    cv_duration_sec = time.time() - start
-
     # Train
     logger.debug("Starting to train model.")
     start = time.time()
@@ -295,6 +320,44 @@ def calculate_model_key(
     return hashlib.sha3_512(json_rep.encode("ascii")).hexdigest()
 
 
+def check_cache(model_register_dir: Union[os.PathLike, str], cache_key: str):
+    """
+    Checks if a model is cached, and returns its path if it exists.
+
+    Parameters
+    ----------
+    model_register_dir: [os.PathLike, None]
+        The register dir where the model lies.
+    cache_key: str
+        A 512 byte hex value as a string based on the content of the parameters.
+
+     Returns
+    -------
+    Union[os.PathLike, None]:
+        The path to the cached model, or None if it does not exist.
+    """
+    existing_model_location = disk_registry.get_value(model_register_dir, cache_key)
+
+    # Check that the model is actually there
+    if existing_model_location and Path(existing_model_location).exists():
+        logger.debug(
+            f"Found existing model at path {existing_model_location}, returning it"
+        )
+        return existing_model_location
+    elif existing_model_location:
+        logger.warning(
+            f"Found that the model-path {existing_model_location} stored in the "
+            f"registry did not exist."
+        )
+        return None
+    else:
+        logger.info(
+            f"Did not find the model with key {cache_key} in the register at "
+            f"{model_register_dir}."
+        )
+        return None
+
+
 def provide_saved_model(
     name: str,
     model_config: dict,
@@ -303,6 +366,7 @@ def provide_saved_model(
     output_dir: Union[os.PathLike, str],
     model_register_dir: Union[os.PathLike, str] = None,
     replace_cache=False,
+    cv_mode: str = "full_build",
 ) -> Union[os.PathLike, str]:
     """
     Ensures that the desired model exists on disk, and returns the path to it.
@@ -334,6 +398,11 @@ def provide_saved_model(
     replace_cache: bool
         Forces a rebuild of the model, and replaces the entry in the cache with the new
         model.
+    cv_mode: str
+        String which enables three different modes:
+        * cross_val_only: Only perform cross validation
+        * build_only: Skip cross validation and only build the model
+        * full_build: Cross validation and full build of the model, default value
 
     Returns
     -------
@@ -353,27 +422,17 @@ def provide_saved_model(
             )
             disk_registry.delete_value(model_register_dir, cache_key)
 
-        existing_model_location = disk_registry.get_value(model_register_dir, cache_key)
+        cached_model_location = check_cache(model_register_dir, cache_key)
 
-        # Check that the model is actually there
-        if existing_model_location and Path(existing_model_location).exists():
-            logger.debug(
-                f"Found existing model at path {existing_model_location}, returning it"
-            )
+        if cached_model_location:
+            return cached_model_location
 
-            return existing_model_location
-        elif existing_model_location:
-            logger.warning(
-                f"Found that the model-path {existing_model_location} stored in the "
-                f"registry did not exist."
-            )
-        else:
-            logger.info(
-                f"Did not find the model with key {cache_key} in the register at "
-                f"{model_register_dir}."
-            )
     model, metadata = build_model(
-        name=name, model_config=model_config, data_config=data_config, metadata=metadata
+        name=name,
+        model_config=model_config,
+        data_config=data_config,
+        metadata=metadata,
+        cv_mode=cv_mode,
     )
     model_location = _save_model_for_workflow(
         model=model, metadata=metadata, output_dir=output_dir
--- a/gordo_components/cli/cli.py
+++ b/gordo_components/cli/cli.py
@@ -12,7 +12,13 @@ import os
 import jinja2
 import yaml
 import click
-from gordo_components.builder.build_model import provide_saved_model
+
+from gordo_components.builder.build_model import (
+    provide_saved_model,
+    check_cache,
+    calculate_model_key,
+    build_model,
+)
 from gordo_components.data_provider.providers import InfluxDataProvider
 from gordo_components.serializer import (
     load_metadata,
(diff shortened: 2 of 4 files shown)


gordo_components/builder/build_model.py
```diff
@@ -45,10 +46,13 @@ def build_model(
         Mapping of the Dataset to initialize, following the same logic as model_config.
     metadata: dict
         Mapping of arbitrary metadata data.
+    cv_mode:
```
milesgranger: `cv_mode: str` 
But also wondering if maybe this is a place where we should be using an `Enum`? 

```python
CvMode = Enum('CvMode', 'FULL PARTIAL SKIP')

# Probably done in the CLI portion:
cv_mode = getattr(CvMode, cv_mode_from_cl_or_click_type_param, CvMode.FULL)

# Then this parameter's signature is:
cv_mode: CvMode
```

Because even in the documentation, without looking at the code, I don't know what the possible values of `cv_mode` are; having it as an `Enum` keeps it pretty explicit about the options, and can even auto update it in `click` as an option with `CvMode.__members__`
```

File: gordo_components/builder/build_model.py [hunk 2]
```diff
@@ -45,10 +46,13 @@ def build_model(
         Mapping of the Dataset to initialize, following the same logic as model_config.
     metadata: dict
         Mapping of arbitrary metadata data.
+    cv_mode: 
+        String which enables three different modes: Only printing cv_scores after cross validation, build the model i.e.
+        no cross validation or do a full build. Default is full build with cross validation.
```
milesgranger: you explain the different modes, but not what values they should be for each. If we don't go with the `Enum` option, we should at least list each option out here with bullet points or something to that effect.

gordo_components/builder/build_model.py
```diff
@@ -45,10 +46,13 @@ def build_model(
...
 
     Returns
     -------
-        Tuple[sklearn.base.BaseEstimator, dict]
+        Tuple[Optional[sklearn.base.BaseEstimator], dict]
```
milesgranger, line 63: I know it wasn't updated before, but just made me notice that we didn't type this return in the actual function.
`-> Tuple[...blah blah]`

gordo_components/builder/build_model.py
```diff
@@ -241,6 +261,33 @@ def calculate_model_key(
     return hashlib.sha3_512(json_rep.encode("ascii")).hexdigest()
 
 
+def check_cache(model_register_dir, cache_key):
```
milesgranger: type declarations plz.

gordo_components/builder/build_model.py
```diff
@@ -241,6 +261,33 @@ def calculate_model_key(
...
+    """
+    Check if the cache exists.
+    Returns None if not.
```
milesgranger: Update docstring to match the numpy style we're using.

gordo_components/builder/build_model.py
milesgranger: @epa095 probably has input on this. :moneybag:

gordo_components/builder/build_model.py
```diff
@@ -45,10 +46,16 @@ def build_model(
         Mapping of the Dataset to initialize, following the same logic as model_config.
     metadata: dict
         Mapping of arbitrary metadata data.
+    cv_mode: 
+        String which enables three different modes:
+        * cross_val_only: Only printing cv_scores after cross validation, build the model i.e.
+        no cross validation or
+        * build only: Skips cross validation and only builds the model
```
milesgranger: I guess you mean `build_only`

gordo_components/builder/build_model.py
```diff
@@ -70,28 +77,44 @@ def build_model(
...
-        scores = {
-            "explained-variance": {
-                "mean": cv_scores.mean(),
-                "std": cv_scores.std(),
-                "max": cv_scores.max(),
-                "min": cv_scores.min(),
-                "raw-scores": cv_scores.tolist(),
+    cv_duration_sec = None
+
+    if cv_mode == "cross_val_only" or cv_mode == "full_build":
```
milesgranger: minor, `if cv_mode.lower() in ("cross_val_only", "full_build"):` ever so slightly easier to read; also `.lower()` maybe help the frustrated user. :sweat_smile:

gordo_components/builder/build_model.py
```diff
@@ -241,6 +264,44 @@ def calculate_model_key(
     return hashlib.sha3_512(json_rep.encode("ascii")).hexdigest()
 
 
+def check_cache(model_register_dir: Union[os.PathLike, str], cache_key: str):
+    """
+    Checks if a model is cached, and returns it path if it exists.
```
milesgranger: `its`

gordo_components/builder/build_model.py
```diff
@@ -249,6 +310,7 @@ def provide_saved_model(
     output_dir: Union[os.PathLike, str],
     model_register_dir: Union[os.PathLike, str] = None,
     replace_cache=False,
+    cv_mode: str = "full",
```
milesgranger: This isn't a valid option `full`

...if only there was just some structure we could use, which had such available choices predefined. :rofl:

gordo_components/builder/build_model.py
```diff
@@ -280,6 +342,9 @@ def provide_saved_model(
     replace_cache: bool
         Forces a rebuild of the model, and replaces the entry in the cache with the new
         model.
+    cv_mode:
```
milesgranger: `cv_mode: str`

gordo_components/cli/cli.py
```diff
@@ -62,6 +67,8 @@
 logger = logging.getLogger(__name__)
 logging.getLogger("azure.datalake").setLevel(azure_log_level)
 
+cv_mode_input = {"cross_val_only", "full_build", "build_only"}
```
milesgranger: So we'll have to update both the documentation for available choices, _and_ remember to update this choice list, which lives in a completely separate module?

gordo_components/cli/cli.py
```diff
@@ -167,8 +184,21 @@ def build(
         config wherever there is a jinja variable with the key.
     model_location_file: str/path
         Path to a file to open and write the location of the serialized model to.
+    data_provider_threads: int
+        Number of threads to use for the data provider when fetching data.
+    cv_mode: str
+        String which enables three different modes:
+        * cross_val_only: Only printing cv_scores after cross validation, build the model i.e.
+        no cross validation or
+        * build only: Skips cross validation and only builds the model
```
milesgranger: `build_only`

gordo_components/cli/cli.py
```diff
@@ -204,15 +234,53 @@ def build(
...
+        model_location = provide_saved_model(
+            name,
+            model_config,
+            data_config,
+            metadata,
+            output_dir,
+            model_register_dir,
+            cv_mode=cv_mode,
+        )
+        metadata = load_metadata(model_location)
```
milesgranger, line 259: Are these two last conditionals the same code?

tests/gordo_components/builder/test_builder.py
```diff
@@ -278,3 +278,36 @@ def test_provide_saved_model_caching(
...
+    """
+    Test checks that the model is None if cross_val_only is used as the cv_mode.
+    If the default mode ('full') is used, the model should not be None.
+
+    Parameters
+    ----------
+    should_be_equal: bool
+        Refers to whether or not the cv_mode should be equal to full (default) or cross_val only.
+    cv_mode: str
+        The mode which is tested, is either full or cross_val_only
```
milesgranger, line 378: We don't normally give our tests this thorough documentation, but I like it. :+1:

tests/gordo_components/cli/test_cli.py
```diff
@@ -202,3 +203,124 @@ def test_expand_model_complains_on_missing_vars(self):
...
+            MODEL_CONFIG=json.dumps(MODEL_CONFIG),
+        ):
+
+            location_file = f"{os.path.join(tmpdir, 'location.txt')}"
+            metadata_file = f"{os.path.join(tmpdir, 'metadata.json')}"
+            runner.invoke(
+                cli.gordo,
+                [
+                    "build",
+                    f"--cv-mode=build_only",
```
milesgranger: Don't really need the string formatting here, unless you were aiming to parameterize this?

milesgranger (commented):
Looking nice, sorry for the tardy review.

gordo_components/cli/cli.py
EinarJohnsen: They are called with different ```cv_mode```, so as ```provide_saved_model``` calls ```build_model```, it will behave differently.

gordo_components/cli/cli.py
```diff
@@ -228,15 +282,17 @@ def build(
...
             all_scores = get_all_score_strings(saved_metadata)
 
         for score in all_scores:
             print(score)
 
-    # Write out the model location to this file.
-    model_location_file.write(model_location)
+    # Write out the model location to this file if it exists and cv_mode is not equal to cross_val_only.
+    if cv_mode != "cross_val_only":
+        model_location_file.write(model_location)
```
milesgranger, line 292: What do you think about doing a simple `return` after you've done what you need to do in `cross_val_only` condition, and then remove this `if`.
Alternatively, still remove the `if` but indent this into the `else` block.

tests/gordo_components/cli/test_cli.py
```diff
@@ -202,3 +207,124 @@ def test_expand_model_complains_on_missing_vars(self):
...
+                ],
+            )
+            # Checks that the file is empty or not depending on the mode.
+            # Also checks the output contains 'explained-variance_raw-scores'
+            if should_be_equal:
+                assert os.stat(location_file).st_size != 0
+                assert "explained-variance_raw-scores=" in result.output
+            else:
+                assert os.stat(location_file).st_size == 0
+                assert "explained-variance_raw-scores=" in result.output
```
milesgranger: Seems this `assert "explained-variance_raw-scores=" in result.output` does not depend on the `should_be_equal` and thus probably doesn't need to be duplicated but pulled out of this condition check, no?

tests/gordo_components/cli/test_cli.py
```diff
@@ -202,3 +207,124 @@ def test_expand_model_complains_on_missing_vars(self):
...
+            MODEL_REGISTER_DIR=model_register_dir,
+        ):
+
+            if should_be_equal:
+                runner.invoke(cli.gordo, ["build", f"--cv-mode={cv_mode_1}"])
+                runner.invoke(cli.gordo, ["build", f"--cv-mode={cv_mode_2}"])
+                assert os.path.exists(model_register_dir)
+            else:
+                runner.invoke(cli.gordo, ["build", f"--cv-mode={cv_mode_1}"])
+                runner.invoke(cli.gordo, ["build", f"--cv-mode={cv_mode_2}"])
```
milesgranger: These
```python
runner.invoke(cli.gordo, ["build", f"--cv-mode={cv_mode_1}"])
runner.invoke(cli.gordo, ["build", f"--cv-mode={cv_mode_2}"])
```
Seem like they can be moved out and before the ` if should_be_equal:`

tests/gordo_components/cli/test_cli.py
```diff
@@ -202,3 +207,124 @@ def test_expand_model_complains_on_missing_vars(self):
...
+    """
+    Checks that cv_scores uses cache if runned after a full build. Loads the same model, and can 
+    print the cv_scores from them. 
+    """
+
+    runner = CliRunner()
+
+    logger.info(f"MODEL_CONFIG={json.dumps(MODEL_CONFIG)}")
+
+    with tempfile.TemporaryDirectory() as tmpdir:
```
milesgranger: Could save yourself a few lines in these tests by using the `tmp_dir` fixture if you like.
https://github.com/equinor/gordo-components/blob/63c53ef9eade25732f683f757adb9a3ee5ce193b/tests/conftest.py#L49

gordo_components/builder/build_model.py
```diff
@@ -48,10 +49,16 @@ def build_model(
         Mapping of the Dataset to initialize, following the same logic as model_config.
     metadata: dict
         Mapping of arbitrary metadata data.
+    cv_mode: str
+        String which enables three different modes:
+        * cross_val_only: Only printing cv_scores after cross validation, build the model i.e.
+        no cross validation or
+        * build_only: Skips cross validation and only builds the model
```
epa095: Imperative>descriptive. So `Skip cross validation and only build the model` instead of "Skips cross validation and only builds the model". And similar for the other ones.

gordo_components/builder/build_model.py
```diff
@@ -73,34 +80,53 @@ def build_model(
...
-            "explained-variance": {
-                "mean": cv_scores.mean(),
-                "std": cv_scores.std(),
-                "max": cv_scores.max(),
-                "min": cv_scores.min(),
-                "raw-scores": cv_scores.tolist(),
+    if cv_mode.lower() in ("cross_val_only", "full_build"):
+
+        # Cross validate
+        logger.debug(f"Starting to do cross validation")
```
epa095: `Starting cross validation` sounds more correct to me, but I be not native freedom-speaker. @milesgranger?

gordo_components/builder/build_model.py
```diff
@@ -308,6 +373,9 @@ def provide_saved_model(
     replace_cache: bool
         Forces a rebuild of the model, and replaces the entry in the cache with the new
         model.
+    cv_mode: str
```
epa095, line 401: Should have the 3 possible values. You can either copy-paste the doc from build_model, or write something like pandas/numpy does: 
```
cv_mode: {'full_build', 'cross_val_only', 'build_only'}
    Do a full model build and calculation of cross-validation (cv) scores, only calculate  cv scores, or only do a model build without cv.
```

gordo_components/cli/cli.py
```diff
@@ -124,6 +130,19 @@ def gordo():
     type=click.File(mode="w", lazy=False),
     default="/tmp/model-location.txt",
 )
+@click.option(
+    "--data-provider-threads",
```
epa095: Some merge issues here I think, since this should be gone. But maybe you have not rebased on top of master?

gordo_components/cli/cli.py
```diff
@@ -167,6 +188,14 @@ def build(
         config wherever there is a jinja variable with the key.
     model_location_file: str/path
         Path to a file to open and write the location of the serialized model to.
+    data_provider_threads: int
```
epa095: same here

gordo_components/cli/cli.py
```diff
@@ -167,6 +188,14 @@ def build(
...
+        Number of threads to use for the data provider when fetching data.
+    cv_mode: str
+        String which enables three different modes:
+        * cross_val_only: Only printing cv_scores after cross validation, build the model i.e.
+        no cross validation or
+        * build_only: Skips cross validation and only builds the model
```
epa095: imperative

gordo_components/builder/build_model.py
```diff
@@ -48,10 +49,16 @@ def build_model(
         Mapping of the Dataset to initialize, following the same logic as model_config.
     metadata: dict
         Mapping of arbitrary metadata data.
+    cv_mode: str
```
epa095, line 55: Numpy usually writes the 3 explicit choices, like this: https://github.com/numpy/numpy/blob/d9b1e32cb8ef90d6b4a47853241db2a28146a57d/numpy/core/fromnumeric.py#L509

epa095: But I am also fine with the `str`, especially if the typechecker `mypy` finds it easier.

tests/gordo_components/cli/test_cli.py
```diff
@@ -202,3 +207,118 @@ def test_expand_model_complains_on_missing_vars(self):
...
+    with temp_env_vars(
+        MODEL_NAME="model-name",
+        OUTPUT_DIR=tmp_dir.name,
+        DATA_CONFIG=DATA_CONFIG,
+        MODEL_CONFIG=json.dumps(MODEL_CONFIG),
+    ):
+
+        location_file = f"{os.path.join(tmp_dir.name, 'location.txt')}"
+        metadata_file = f"{os.path.join(tmp_dir.name, 'metadata.json')}"
+        print(metadata_file)
```
epa095: dd

epa095 (commented):
Starting to look good:-)

gordo_components/cli/cli.py
EinarJohnsen: Yes, I have rebased on top of master recently, but obviously have made a mistake at some point. Have removed these now.

epa095 (approved):
LGTM:-)

PR #405, opened 2019-08-12
[WIP] Support better dataframe to JSON and Msgpack

### LINKED ISSUE #403: Simplify sending of data from /prediction endpoints
Context
----------
We use `pandas.MultiIndex` in our columns, which works nicely during the server prediction process

Problem
----------
`pandas.read_json(df.to_json())` doesn't actually work on such a `MultiIndex` column dataframe
so we use custom serializer to send and reconstruct those dataframes between the server and the client.

Potential Solution
----------------------
Simplify the custom serialization (if possible) and/or just use `df.to_msgpack()` and `pandas.read_msgpack()`, these keep the `MultiIndex` format and handles `DatetimeIndex` issues as well. Definitely/maybe use `msgpack` between server and client; perhaps with a url serialization parameter?

milesgranger:
Will close #403 

Use msgpack for client and server interactions. 

Support accepting `pandas.DataFrame.from_dict()` formats for making predictions, and return msgpack or JSON; depending on `Accept` header being `application/x-msgpack` or `application/msgpack` [based on this](https://github.com/msgpack/msgpack/issues/194) or JSON if `application/json`

### DIFF
--- a/gordo_components/server/utils.py
+++ b/gordo_components/server/utils.py
@@ -3,6 +3,7 @@
 import logging
 import functools
 import timeit
+import traceback
 import typing
 import dateutil
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


milesgranger:
Closing in favor of #410

PR #409, opened 2019-08-13
Fix threading

### LINKED ISSUE #408: IROC workflows defaults to 1 thread
IROC dataprovider has always used 50 threads by default, but with the introduction of #344 the default thread count in the model-builder and client was set to `1` for backwards-compability with the old NCS behaviour. This has the unintended effect that IROC also downloads with a single thread, making it much slower than it was. 

There are a few fixes. 
- Change the default from `1`thread to `None`, which leaves it up to the dataprovider to choose. But then we should probably also change the default for NCS-reader to not be `5*CPU-count` to some constant, since the CPU-count can change from cloud-machine to cloud-machine, and does not take into account the memory it is allocated anyway. 

- Allow a `threads` keyword in the dataset definition. This is already used by several, but it has no effect today(?).  But do we want to mix data_provider parameters in with dataset parameters?

- Add a data-provider key in the config, and pass this on to the builder and the client. The client already takes in a data-provider, but it is hard-coded in the yaml, not something passed down from the config file.

epa095:
This closes #408

### DIFF
--- a/gordo_components/data_provider/ncs_reader.py
+++ b/gordo_components/data_provider/ncs_reader.py
@@ -29,7 +29,7 @@ class NcsReader(GordoBaseDataProvider):
     }
 
     def __init__(
-        self, client: core.AzureDLFileSystem, threads: Optional[int] = None, **kwargs
+        self, client: core.AzureDLFileSystem, threads: Optional[int] = 1, **kwargs
     ):
         """
         Creates a reader for tags from the Norwegian Continental Shelf. Currently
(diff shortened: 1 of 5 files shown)


gordo_components/data_provider/ncs_reader.py
```diff
@@ -38,12 +38,15 @@ def __init__(
...
         ----------
         threads : Optional[int]
-            Number of threads to use. If None then use # of CPU's * 5
+            Number of threads to use. If None then use 1 thread
 
         """
         super().__init__(**kwargs)
         self.client = client
         self.threads = threads
+        if not self.threads:
```
milesgranger: Should we not simply set the default in `__init__` then?

milesgranger (commented):
Looks cool, I guess there will be a PR in infrastructure to compliment this?

epa095:
> Looks cool, I guess there will be a PR in infrastructure to compliment this?

Yeah, https://github.com/equinor/gordo-infrastructure/issues/334

flikka (approved):
Good stuff.

PR #410, opened 2019-08-13
Simplify data serialization

### LINKED ISSUE #403: Simplify sending of data from /prediction endpoints

milesgranger:
Will close #403 

Approach
------------
Server uses a method to serialized a `MultiIndex` column df into a json serializable `dict`, implemented a complimentary inverse of such a function for the client to use. This simplifies the (de)serialization of data between the client and the server. 

Did not pursue `msgpack` yet, as this seems to satisfy our issue anyhow, so maybe that's not needed anyway.

### DIFF
--- a/gordo_components/server/utils.py
+++ b/gordo_components/server/utils.py
@@ -3,9 +3,9 @@
 import logging
 import functools
 import timeit
-import typing
 import dateutil
 from datetime import datetime
+from typing import Union, List
 
 import pandas as pd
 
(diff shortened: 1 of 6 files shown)


gordo_components/server/utils.py
```diff
@@ -168,6 +132,53 @@ def parse_iso_datetime(datetime_str: str) -> datetime:
     return parsed_date
 
 
+def _df_from_dict_or_respond(
+    data: Union[dict, list], expected_columns: List[str]
+) -> Union[Response, pd.DataFrame]:
+    """
+    Convert a dict or list of dicts into a :class:`pandas.core.DataFrame`.
+    Setting the column names to ``expected_columns`` if not already labeled and
+    the length of the columns match the length of the epxected columns.
```
epa095: `epxected`. Also, does the docstring reflect what it does?

epa095 (commented):
What about a simple test which tests
`multi_lvl_column_dataframe_from_dict(multi_lvl_column_dataframe_to_dict(df)) == df`
and
`multi_lvl_column_dataframe_to_dict(multi_lvl_column_dataframe_from_dict(d))==d`
?

gordo_components/server/utils.py
```diff
@@ -168,6 +132,53 @@ def parse_iso_datetime(datetime_str: str) -> datetime:
     return parsed_date
 
 
+def _df_from_dict_or_respond(
```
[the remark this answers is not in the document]
milesgranger: Ya, I was wavering on the naming.. basically, _"look man, I'll try a build you a dataframe from this dict here, but if not, I'll instead build you a response reporting the error"_
 it was `response` maybe that is better?

gordo_components/server/utils.py
```diff
@@ -168,6 +132,53 @@ def parse_iso_datetime(datetime_str: str) -> datetime:
...
+    if not all(col in df.columns for col in expected_columns):
+
+        # If the length doesn't mach, then we can't reliably determine what data we have hre.
+        if len(df.columns) != len(expected_columns):
+            msg = dict(
+                message=f"Unexpected features:"
+                f" was expecting {expected_columns} length of {len(expected_columns)}, "
+                f"but got {df.columns} length of {len(df.columns)}"
+            )
+            return make_response((jsonify(msg), 400))
```
epa095, line 170: I hate this! Is it not better to raise e.g. ValueError, and then let the caller make the stupid response object if it wants?

epa095 (commented):
I mostly loooove it.

gordo_components/server/utils.py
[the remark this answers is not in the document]
epa095: Ahh, but then should it not be `_df_or_response_from_dict` ?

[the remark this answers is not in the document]
epa095: Of course, I prefer to avoid this whole thing by *not* having it return a stupid response object.

gordo_components/server/utils.py
milesgranger: Hate is a strong word. 

Basically this is only taking place in `extract_X_y` decorator, so if it cannot do that, it will return an error before the decorated method has the opportunity to execute, because it simply _can't_ without such data being extracted. 

with that said, I'm ok raising an error in this function, but in `extract_X_y` returning the response. Heck, even could return a fancy `enum`if you like. :+1:

epa095: > with that said, I'm ok raising an error in this function, but in extract_X_y returning the response. 

What a lovely idea!
> Heck, even could return a fancy enumif you like. 

No thanks:-)

milesgranger: Although, the message won't be as informative as it won't have access to the previously built dataframe occuring in this function. 

```rust
let df = match _df_or_resp_from_dict(data) {
    Dataframe(df) => df,
    Response(resp) => return resp
}
```

epa095: You can just put the message in the ValueError and use it in the response.

epa095 (approved):
LGTM

PR #413, opened 2019-08-15
Add exponential retries to influx forwarder

### LINKED ISSUE #343: Client: Retry pushes to influx
Today there is some retry-logic for failing POST requests to the server, but there is no retry-logic if a write to influx fails, and then the whole client crashes. We could port the same logic to where we write data to influx

### LINKED ISSUE #411: Exponential retries in client
The client supports retrying requests for predictions, but it does the requests right after another. Maybe better to do exponential backof? I found the python library [tenacity](https://tenacity.readthedocs.io/en/latest/) which is supposed to make retries easy and nice, and it includes exponential retries as one method. Maybe use it instead of our own little loop? And then maybe also do #343 at the same time?

epa095:
This closes #343

### DIFF
--- a/gordo_components/client/forwarders.py
+++ b/gordo_components/client/forwarders.py
@@ -1,7 +1,9 @@
 # -*- coding: utf-8 -*-
 
 import asyncio
+import itertools
 import logging
+import time
 import typing
 from typing import Optional
 from concurrent.futures import ThreadPoolExecutor
@@ -59,6 +61,7 @@ class ForwardPredictionsIntoInflux(PredictionForwarder):
         destination_influx_uri: Optional[str] = None,
         destination_influx_api_key: Optional[str] = None,
         destination_influx_recreate: bool = False,
+        n_retries=5,
     ):
         """
         Create an instance which, when called, is a coroutine capable of
@@ -75,6 +78,7 @@ class ForwardPredictionsIntoInflux(PredictionForwarder):
             Drop the database before filling it with data?
         """
         # Create df client if provided
+        self.n_retries = n_retries
         self.dataframe_client = (
             influx_client_from_uri(
                 destination_influx_uri,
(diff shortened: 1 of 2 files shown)


flikka (commented):
Nice stuff, would be cool with some testing. Not too easy perhaps, but maybe setting n_retries to 1 and 2, and do something?

epa095:
> Nice stuff, would be cool with some testing. Not too easy perhaps, but maybe setting n_retries to 1 and 2, and do something?

Testing is cool. But how?

flikka (approved):
Let's be pragmatic, I agree it is not easy to test, also with the sleep stuff it becomes a bit weird.

If it works - maybe testing it, go for it.

gordo_components/client/forwarders.py
```diff
@@ -152,39 +156,58 @@ def __init__(
...
+                    if current_attempt <= self.n_retries:
+                        # Sleep at most 5 min
+                        time_to_sleep = min(2 ^ (current_attempt + 2), 300)
+                        logger.warning(
+                            f"Failed to forward data to influx on attempt "
+                            f"{current_attempt} out of {self.n_retries} attempts. "
+                            f"Error: {exc}."
+                            f"Sleeping {time_to_sleep} seconds and trying again"
+                        )
+                        time.sleep(time_to_sleep)  # Not async on purpose.
```
milesgranger, line 190: I missed this review entirely, but this will end up holding up _all_ other coroutines in the current batch, no? Why was async sleep not an option?

gordo_components/client/forwarders.py
```diff
@@ -152,39 +156,58 @@ def __init__(
...
-                    self.dataframe_client.write_points,
-                    dataframe=sub_df,
-                    measurement=top_lvl_name,
-                    tags=tags,
-                    batch_size=10000,
-                )
-                await asyncio.wrap_future(future)
+            await self._write_to_influx_with_retries(sub_df, tags, top_lvl_name)
+
+    async def _write_to_influx_with_retries(self, df, tags, measurement):
```
milesgranger, line 161: No type defs?

milesgranger (commented):
This review is tardy and holds no merit, but I'm doing it anyway. :1234:  :tophat:

PR #414, opened 2019-08-15
Created a dev-requirements.txt

### LINKED ISSUE #412: Create a dev-requirements.txt
Problem
----------
We have a lot of a unnecessary install requirements in our project.

Such as:
https://github.com/equinor/gordo-components/blob/522feaec8c1d50bc62f4b79e6735d62ac3e5ea4b/requirements.in#L22

This ends up having the end user install these packages whilst installing gordo-components.

Solution?
-----------
Create a `dev_requirements.txt` and move these unneeded requirements into that, which would then be installed in Travis/local dev but would not end up requiring the end user to install these.

Patil2099:
Fixes #412


milesgranger (changes requested):
First, thank you for taking time to work on this issue, we really appreciate it! :100: 

I've made a comment on additional dependencies you can move out of `requirements.in`. 

Then you will need to do the following:
- `pipe-compile --output-file=dev_requirements.txt dev_requirements.in`
   this will produce the `dev_requirements.txt` as you would expect.

- Since you've modified `requirements.in` we'll need to run the same command but for that file:

  `pip-compile --output-file=requirements.txt requirements.in` 

  to compute the updated `requirements.txt`

- Finally, you'll just need to add `pip install -r dev_requirements.txt` after this [line here](https://github.com/jazzband/pip-tools) in the `.travis.yml` file. This is because the `Make docs` is actually the only portion of the CI which needs these dependencies. 

---

After that, I think we'll be good to go. If you don't have `pip-compile` command, that is brought by [pip-tools](https://github.com/jazzband/pip-tools)

Patil2099:
> First, thank you for taking time to work on this issue, we really appreciate it! 💯
> 
> I've made a comment on additional dependencies you can move out of `requirements.in`.
> 
> Then you will need to do the following:
> 
> * `pipe-compile --output-file=dev_requirements.txt dev_requirements.in`
>   this will produce the `dev_requirements.txt` as you would expect.
> * Since you've modified `requirements.in` we'll need to run the same command but for that file:
>   `pip-compile --output-file=requirements.txt requirements.in`
>   to compute the updated `requirements.txt`
> * Finally, you'll just need to add `pip install -r dev_requirements.txt` after this [line here](https://github.com/jazzband/pip-tools) in the `.travis.yml` file. This is because the `Make docs` is actually the only portion of the CI which needs these dependencies.
> 
> After that, I think we'll be good to go. If you don't have `pip-compile` command, that is brought by [pip-tools](https://github.com/jazzband/pip-tools)

Which line in .travis.yml ?

.travis.yml
```diff
@@ -31,6 +31,7 @@ jobs:
   - name: "make docs"
     install:
       - pip install -r requirements.txt
+      -pip install -r dev_requirements.txt
```
milesgranger: _Very_ small nit-pick here, could you add a space between the `-` and `pip` so it's `- pip install -r dev_requirements.txt`

milesgranger:
ah, ok..pytest is trying to be smart and collect from `docs`, so we'll need to add ` --ignore=docs` to the `pytest.ini` at this line: 

https://github.com/equinor/gordo-components/blob/0be52ac0eb93c9e2b5391ac545d5c73e5a2e22cd/pytest.ini#L2

So it'll start like this:
`addopts = --ignore=docs ...`

Patil2099:
@milesgranger  is it okay now ?

milesgranger (approved):
Thank you, sir! :+1:

PR #417, opened 2019-08-16
Return None if register_dir is None in disk_registry

### LINKED ISSUE #407: Return None if registry_dir in disk_registry is None.
Path(registry_dir).joinpath(key) fails if registry_dir is None. Should return None if this is the case, as the documentation says.

EinarJohnsen:
This fixes #407


epa095 (approved):
LGTM!

PR #419, opened 2019-08-16
Fix logging-issue stemming from absl

epa095:
Tensorflow imports absl, and it breaks the standard logging module in
python by setting a handler on the root-level logger. The result is
that the log-level does not get set, and the format is all wrong.
This patch introduces an ugly fix I found on the internet to handle it.
Hopefully we can delete it in the future.

References:
https://github.com/tensorflow/tensorflow/issues/29842
https://github.com/tensorflow/tensorflow/issues/26691
https://github.com/abseil/abseil-py/issues/99

### DIFF
--- a/gordo_components/cli/cli.py
+++ b/gordo_components/cli/cli.py
@@ -5,6 +5,8 @@ CLI interfaces
 """
 
 import logging
+import warnings
+import traceback
 import os
 
 import jinja2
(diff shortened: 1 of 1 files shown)


gordo_components/cli/cli.py
```diff
@@ -25,6 +25,29 @@
...
+    # The code below fixes this double-logging.  FMI see:
+    #   https://github.com/tensorflow/tensorflow/issues/26691#issuecomment-500369493
+
+    import absl.logging
+
+    logging.root.removeHandler(absl.logging._absl_handler)
+    absl.logging._warn_preinit_stderr = False
+
+except Exception as e:
+    print("Failed to fix absl logging bug", e)
```
milesgranger: `warning.warn(f"....{traceback.format_exc()})"` or something other than `print`?

gordo_components/cli/cli.py
```diff
@@ -33,6 +56,7 @@
     level=getattr(logging, log_level),
     format="[%(asctime)s] %(levelname)s [%(name)s.%(funcName)s:%(lineno)d] %(message)s",
 )
+
```
milesgranger, line 61: nice catch.

milesgranger (approved):

PR #420, opened 2019-08-16
Extract model metadata from the depths

### LINKED ISSUE #418: Transform X and y using same scaler when training autoencoders
When training an autoencoder, if the features are scaled, we need to scale y in the same way during fitting the model, so that the input to the model is equal to the output.  One way to do so is by using scikit learn TransformedTargetRegressor. This can be implemented in gordo by using the config string
```yaml
sklearn.compose.TransformedTargetRegressor:
    regressor: 
        sklearn.pipeline.Pipeline:
            steps: 
            - sklearn.preprocessing.data.MinMaxScaler
            - gordo_components.model.models.KerasAutoEncoder:
                kind: feedforward_hourglass
                compression_factor: 0.5
                encoding_layers: 2
                func: tanh
                out_func: linear
                epochs: 3
    transformer: sklearn.preprocessing.data.MinMaxScaler
```
By doing so the pipeline becomes an argument to the _TransformedTargetRegressor_, and so the model is no longer a scikit learn pipeline. This breaks parts of the model building, for example the function __get_final_gordo_base_step_ used in _build_model_.

milesgranger:
Will close #418 

Context
-----------
Models such as our `Keras` models generate some metadata which we like to have; typically with `.get_metadata()`

Problem
-----------
When a model is tucked away inside a pipeline, we use to dig into it, finding the last step and ran this `.get_metadata()`

Now as the model is in a pipeline, which is then in another estimator which is passed as a parameter, and an arbitrary amount of nested models later, we basically don't know where the _final_ model is at.

Solution
-----------
Recursively check attributes of the model for sub-models which are `GordoBase` instances.

### DIFF
--- a/gordo_components/builder/build_model.py
+++ b/gordo_components/builder/build_model.py
@@ -111,9 +111,7 @@ def build_model(
         "cross-validation": {"cv-duration-sec": cv_duration_sec, "scores": scores},
     }
 
-    gordobase_final_step = _get_final_gordo_base_step(model)
-    if gordobase_final_step:
-        metadata["model"].update(gordobase_final_step.get_metadata())
+    metadata["model"].update(_get_metadata(model))
 
     return model, metadata
 
@@ -141,6 +139,50 @@ def _save_model_for_workflow(
     os.makedirs(output_dir, exist_ok=True)  # Ok if some dirs exist
     serializer.dump(model, output_dir, metadata=metadata)
     return output_dir
+
+
+def _get_metadata(model: BaseEstimator, metadata: dict = dict()) -> dict:
+    """
+    Recursively check for :class:`gordo_components.model.base.GordoBase` in a
+    given ``model``. If such a model exists buried inside of a
+    :class:`sklearn.pipelilne.Pipeline` which is then part of another
+    :class:`sklearn.base.BaseEstimator`, this function will return its metadata.
+
+    Parameters
+    ----------
+    model: BaseEstimator
+    metadata: dict
+        Any initial starting metadata, but is mainly meant to be used during
+        the recursive calls to accumulate any multiple :class:`gordo_components.model.base.GordoBase`
+        models found in this model
+
+    Notes
+    -----
+    If there is a ``GordoBase`` model inside of a ``Pipeline`` which is not the final
+    step, this function will not find it.
+
+    Returns
+    -------
+    dict
+        Dictionary representing accumulated calls to :meth:`gordo_components.model.base.GordoBase.get_metadata`
+    """
+    metadata = metadata.copy()
+
+    if not isinstance(model, GordoBase):
+
+        for attr, val in model.__dict__.items():
+            if isinstance(val, Pipeline):
+                final_step = _get_final_gordo_base_step(val)
+                if final_step is not None:
+                    metadata.update(_get_metadata(final_step))
+            elif isinstance(val, GordoBase):
+                metadata.update(val.get_metadata())
+            elif isinstance(val, BaseEstimator):
+                metadata.update(_get_metadata(val))
+    else:
+        metadata.update(model.get_metadata())
+
+    return metadata
 
 
 def _get_final_gordo_base_step(model: BaseEstimator):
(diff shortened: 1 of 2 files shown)


tests/gordo_components/builder/test_builder.py
```diff
@@ -27,182 +26,129 @@ def get_random_data():
...
-    """
-    Test functionality of the builder processes
-    """
-
-    def test_output_dir(self):
-        """
-        Test building of model will create subdirectories for model saving if needed.
-        """
-        from gordo_components.builder import build_model
+def metadata_check(metadata, check_history):
```
epa095, line 36: Nice moving away from unittest:-)

gordo_components/builder/build_model.py
```diff
@@ -143,6 +141,50 @@ def _save_model_for_workflow(
...
+    dict
+        Dictionary representing accumulated calls to :meth:`gordo_components.model.base.GordoBase.get_metadata`
+    """
+    metadata = metadata.copy()
+
+    if not isinstance(model, GordoBase):
+
+        for attr, val in model.__dict__.items():
+            if isinstance(val, Pipeline):
+                final_step = _get_final_gordo_base_step(val)
```
flikka: Oh, my head explodes. This _get_final_gordo_base_step is also recursive.

gordo_components/builder/build_model.py
```diff
@@ -143,6 +141,50 @@ def _save_model_for_workflow(
...
+    step, this function will not find it.
+
+    Returns
+    -------
+    dict
+        Dictionary representing accumulated calls to :meth:`gordo_components.model.base.GordoBase.get_metadata`
+    """
+    metadata = metadata.copy()
+
+    if not isinstance(model, GordoBase):
```
flikka: Maybe we need an equivalent test for Pipeline, to get into the _get_final_gordo_base_step function

gordo_components/builder/build_model.py
```diff
@@ -143,6 +141,50 @@ def _save_model_for_workflow(
...
+
+        for attr, val in model.__dict__.items():
```
flikka: Could do with a small comment here, to explain why this is needed (ie. for the special target transform regressor thing, and for the theoretical case where a gordo base model has a gordo model in it.)

gordo_components/builder/build_model.py
```diff
@@ -143,6 +141,50 @@ def _save_model_for_workflow(
...
+                metadata.update(val.get_metadata())
+            elif isinstance(val, BaseEstimator):
+                metadata.update(_get_metadata(val))
+    else:
+        metadata.update(model.get_metadata())
+
+    return metadata
+
+
 def _get_final_gordo_base_step(model: BaseEstimator):
```
flikka: I suggest we rename this to "get_gordo_model_from_pipeline"

flikka: Maybe we should even integrate this into the recursive _get_metadata() function. I think that might make it a bit simpler to read, overall?

flikka (changes requested):
Tried to comment a bit, but we also talked. Overall, I think I get a bit confused by the various paths into and through this recursive stuff.

gordo_components/builder/build_model.py
milesgranger: Removed the function entirely, because, as you pointed out, the final step should actually be a BaseEstimator which in turn took a `GordoBase` as an estimator. :+1:

gordo_components/builder/build_model.py
milesgranger: Went with your second suggestion. :1st_place_medal:

gordo_components/builder/build_model.py
```diff
@@ -143,6 +141,50 @@ def _save_model_for_workflow(
     return output_dir
 
 
+def _get_metadata(model: BaseEstimator, metadata: dict = dict()) -> dict:
+    """
+    Recursively check for :class:`gordo_components.model.base.GordoBase` in a
+    given ``model``. If such a model exists buried inside of a
+    :class:`sklearn.pipelilne.Pipeline` which is then part of another
```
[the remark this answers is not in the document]
milesgranger: Good catch!

flikka (approved):

flikka:
Good stuff. We'll make a new issue #422 and maybe come up with a better solution.

PR #421, opened 2019-08-16
[WIP] Add deployer

milesgranger:


epa095:
Lovely! When you get an 'changed' event, is it possible to see what part of the object changed (i.e the paths in the tree)? Because if workflow generator is going to start the workflow, and then inject the workflow id into the Gordo yaml again, then that will fire a change event which we don't want to have starting another job. But it someone applies a changed config with changed machines then we do want a redeploy.

milesgranger:
Updated to see what type of event occured within `gordo` resources, albeit, I think I'd need to discuss more with you the process flow in general as I don't think I _fully_ understand how it's suppose to go. In my mind this should do the following:

1. If a new gordo resource is added, then we'll always want to start a worfkow generator job. (implemented here)
2. If one is modified we'll want to check if it was modified by the workflow generator job and then _not_ do anything, but if it was updated by another `kubectl apply -f my-gordo-config.yml` _then_ we'll want do do something. Not implemented yet, but suppose this would fall under the `WatchEvent::Modified` case.

milesgranger:
Closing in favor of https://github.com/equinor/gordo-infrastructure/pull/347

PR #423, opened 2019-08-20
Filter bad data (code 0) from the datalake.

### LINKED ISSUE #365: Filter "bad" data (code 0) from the datalake
The csv files we read here in the Equinor has a status column, which is either 192 (Good) or 0 (Bad).

We should filter out bad data on reading.

EinarJohnsen:
Removes rows with code 0 from the data in the ncs_reader. Fixes #365


epa095 (commented):
Definitely looking good. But do you remove the "status" column again, or is also returned now?

gordo_components/data_provider/ncs_reader.py
```diff
@@ -58,6 +58,7 @@ def load_series(
         to_ts: datetime,
         tag_list: List[SensorTag],
         dry_run: Optional[bool] = False,
+        remove_bad_values: Optional[bool] = True,
```
milesgranger: As we have lots of 'Status codes', perhaps we should change this to `filter_status_codes = [0]` and this way we can support any future changes in status codes or other requirements without hard-coding.

gordo_components/data_provider/ncs_reader.py
```diff
@@ -139,13 +144,14 @@ def read_tag_files(
...
-                    usecols=[tag.name, "Timestamp"],
+                    usecols=[tag.name, "Timestamp", "Status"],
                     dtype={tag.name: np.float32},
                     parse_dates=["Timestamp"],
                     date_parser=lambda col: pd.to_datetime(col, utc=True),
                     index_col="Timestamp",
                 )
-
+                if remove_bad_values:
+                    df = df[df["Status"] != 0]
```
milesgranger: And then here, it would be a one liner; `df = df[df["Status"].isin(filter_status_codes)]`

tests/gordo_components/data_provider/test_ncs_reader.py
```diff
@@ -109,3 +109,25 @@ def test_load_series_dry_run(dates, ncs_reader):
...
+    {
+        "1776-troc": os.path.join(
+            os.path.dirname(os.path.realpath(__file__)), "data", "datalake"
+        )
+    },
+)
+@pytest.mark.parametrize("remove_bad_values", [True, False])
+def test_load_series_with_filter_bad_data(dates, ncs_reader, remove_bad_values):
+    valid_tag_list = normalize_sensor_tags(["TRC-322"])
+    frame = ncs_reader.load_series(
```
milesgranger: This isn't a frame, but rather a series generator, no?

tests/gordo_components/data_provider/test_ncs_reader.py
```diff
@@ -109,3 +109,25 @@ def test_load_series_dry_run(dates, ncs_reader):
...
+        dates[0], dates[1], valid_tag_list, remove_bad_values=remove_bad_values
+    )
+    # Checks if the bad data from the files under tests/gordo_components/data_provider/data/datalake/TRC-322
+    # are filtered out. 20 rows exists, 5 of then have the value 0.
+    if remove_bad_values:
+        assert len([value for pdf in frame for value in pdf]) == 15
+    else:
+        assert len([value for pdf in frame for value in pdf]) == 20
```
milesgranger: Can we instead do something like this:
```python
n_expected = 15 if remove_bad_values else 20
assert all(len(series) == n_expected for series in series_gen)
```

tests/gordo_components/data_provider/test_ncs_reader.py
EinarJohnsen: This is true, it is a series generator.

gordo_components/data_provider/ncs_reader.py
EinarJohnsen: Would ```remove_status_codes = [0]``` be a better name? ```filter_status_code``` is kind of ambiguous, you have to look at the code to see if the status codes are kept or removed. @milesgranger

gordo_components/data_provider/ncs_reader.py
```diff
@@ -112,7 +112,7 @@ def read_tag_files(
             range object providing years to include
         dry_run: bool
             if True, don't download data, just check info, log, and return
-        remove_bad_values: bool
+        remove_status_codes: bool
             If True, removes data with code 0 in the Status field.
```
milesgranger: Will need to update the parameter docs and its type.

milesgranger (commented):
At least from my end, just had one last comment. After that, you've got my approval I think.

milesgranger (approved):
:100: :tophat: :100:

PR #425, opened 2019-08-21
Make client auto choose prediction endpoint

### LINKED ISSUE #400: Update client to auto-choose endpoint
Problem
----------
Model is more-or-less hard coded to `/anomaly/prediction`

Context
----------
#368 sets the return code from `/anomaly/prediction` as `422 Unprocessable Entity` if the model is not capable of doing anomaly detection. 

Potential solution(s)
-------------------------
Should now be able to set the `client` to always try `/anomaly/prediction` first and then switch endpoints to `/prediction` should the return code be `422`

Otherwise we'll need to update the model's metadata to reflect what type of model it is and then decide in the client which endpoint to connect to based on that.

milesgranger:
Will close #400 

Problem
-----------
We use to have to manually specify `prediction_path` to `Client` for either `/prediction` or `/anomaly/prediction`

Solution
-----------
Now that we are returning `422` status code if the model being served is not an AnomalyDetector, the client can default to `/anomaly/prediction` and when it gets a `422` can update its `prediction_path` to `/prediction` instead for all future requests.

### DIFF
--- a/gordo_components/client/client.py
+++ b/gordo_components/client/client.py
@@ -19,6 +19,7 @@ from werkzeug.exceptions import BadRequest
 
 from gordo_components import serializer
 from gordo_components.client import io as gordo_io
+from gordo_components.client.io import HttpUnprocessableEntity
 from gordo_components.client.forwarders import PredictionForwarder
 from gordo_components.client.utils import EndpointMetadata, PredictionResult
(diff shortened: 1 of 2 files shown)


epa095 (approved):
LGTM, but remember to change commit message! 
And then you are allowed to use your mighty admin privileges to ignore the test-checks.

PR #426, opened 2019-08-22
Add **kwargs to sub_providers (iroc and ncs_reader)

### LINKED ISSUE #387: DataLakeProvider: Pass kwargs to sub-dataproviders
They take in `**kwargs` in their init, so it is safe to pass them arbitrary parameters.

EinarJohnsen:
The test: ```test_load_series_with_filter_bad_data``` broke because it used a parameter (```remove_bad_values```) which is now passed from kwargs in the DataLakeProvider (so it can be used by end-users). Therefore that test is fixed by passing the parameter directly into the NcsReader's __init__.  Fixes #387

### DIFF
--- a/gordo_components/data_provider/ncs_reader.py
+++ b/gordo_components/data_provider/ncs_reader.py
@@ -29,7 +29,11 @@ class NcsReader(GordoBaseDataProvider):
     }
 
     def __init__(
-        self, client: core.AzureDLFileSystem, threads: Optional[int] = 1, **kwargs
+        self,
+        client: core.AzureDLFileSystem,
+        threads: Optional[int] = 1,
+        remove_status_codes: Optional[list] = [0],
+        **kwargs,
     ):
         """
         Creates a reader for tags from the Norwegian Continental Shelf. Currently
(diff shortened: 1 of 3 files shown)


gordo_components/data_provider/ncs_reader.py
```diff
@@ -39,11 +43,15 @@ def __init__(
         ----------
         threads : Optional[int]
             Number of threads to use. If None then use 1 thread
+        remove_status_codes: Optional[list]
+            removes data with Status code(s) in the list. By default it removes data
```
milesgranger: Nit pick here, but could you capitalize the start of this sentence?

milesgranger (approved):
Approved, with one very small comment.

gordo_components/data_provider/ncs_reader.py
```diff
@@ -39,11 +43,15 @@ def __init__(
         ----------
         threads : Optional[int]
             Number of threads to use. If None then use 1 thread
+        remove_status_codes: Optional[list]
+            Removes data with Status code(s) in the list. By default it removes data
+            with Status code 0. Can be overwritten from kwargs.
```
epa095: Ditch the `Can be overwritten from kwargs.` From the perspective of this class it is overwritten by the explicit parameter your are documenting, not kwargs.

epa095 (approved):
LGTM, except minor documentation comment. I would also try to get `DataLakeProvider` into the git commit message somehow, since it is only that one which is touched. Remember it is alowed to write a larget git message below the headline ;-)

PR #427, opened 2019-08-22
Allow serialization of hidden Gordo models & wrapper for metric functions

milesgranger:
Meant as a temp fix, until we can get `GordoBase` models to be pickle-able (namely, Keras models because we save the tensorflow graph as part of the model).

Problem :warning: 
-----------
If a Keras model is hidden in an attribute of another estimator, the serializer won't see it and attempt to pickle it; which won't work.

Solution :bulb: 
-----------
Iterate over the attributes of the estimator, and if it's a GordoBase, save the model to a sub dir where this model will be saved, and update the original attribute to hold metadata about how to load back that attribute.

Bonus! :tada: 
---------
Added wrapper for scoring functions to ensure their input is aligned. ie. If X, y was given to an LSTM, it's output is less than `y`, the wrapper will match the offset of `yhat` so the scoring function operates without problem.

### DIFF
--- a/gordo_components/builder/build_model.py
+++ b/gordo_components/builder/build_model.py
@@ -10,6 +10,7 @@ from typing import Union, Optional, Dict, Any
 
 from sklearn.base import BaseEstimator
 from sklearn.model_selection import cross_val_score, TimeSeriesSplit
+from sklearn.metrics import explained_variance_score, make_scorer
 from sklearn.pipeline import Pipeline
 
 from gordo_components.util import disk_registry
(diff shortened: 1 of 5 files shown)


milesgranger:
Ran with the PR in gordo-test-project (https://github.com/equinor/gordo-test-project/pull/14) which uses the `DiffBasedAnomalyDetector` along with `TransformedTargetRegressor`

image

**Disclaimer** Changed some code, between that screen shot and now.. but should work(?) :man_shrugging:

epa095 (approved):
Lgtm

PR #428, opened 2019-08-23
Change interactive logic

### LINKED ISSUE #389: DataLakeProvider: Make interactive mean interactive
Today the logic behind `interactive` is a bit weird (to me at least):
- If the env-variable `DL_SERVICE_AUTH_STR` exists then use it independently of whether `interactive` is set or not. 
- If the env-variable is not set AND `interactive` is False, then raise `ValueError`. 
- If the env-variable is not set AND `interactive` is True then try to do interactive login.

This means that the only way to force `interactive` login is to clear the env-variable. 
To me this seems like a more logical setup: 
- If `interactive` is True then do interactive login.
- If `interactive` is False AND the env-variable is set then use it.
- If `interactive` is False AND the env-variable is NOT set then raise `ValueError`. 

What U think @flikka ? I think you made it so you get to have opinion:-p Also you are old and wise.

EinarJohnsen:
Interactive login is prioritized if set to True. Fixes #389

### DIFF
--- a/gordo_components/data_provider/azure_utils.py
+++ b/gordo_components/data_provider/azure_utils.py
@@ -11,7 +11,7 @@ logger = logging.getLogger(__name__)
 
 
 def get_datalake_token(
-    interactive=True, dl_service_auth_str=None
+    interactive: bool = False, dl_service_auth_str: str = None
 ) -> lib.DataLakeCredential:
     """
     Provides a token for azure datalake, either by parsing a datalake service
@@ -19,10 +19,9 @@ def get_datalake_token(
 
     Parameters
     ----------
-    interactive
-        If true then fall back to interactive authentication in case
-        `dl_service_auth_str` is empty
-    dl_service_auth_str
+    interactive: bool
+        If true then use interactive authentication
+    dl_service_auth_str: str
         String on the format tenant:client_id:client_secret
 
     Returns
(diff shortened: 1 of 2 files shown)


EinarJohnsen:
As I understand it, the codecov coverage from earlier was very similar to this. Coverage is down by -1.22%, the 50% diff seems to partly be because of the ```walk_azure``` function. It could be fixed by adding a test for this, but I think it should be in it's own PR as it is related to the iroc_reader and not the interactive thing. But please correct me if I am wrong. @epa095 @milesgranger

gordo_components/data_provider/azure_utils.py
```diff
@@ -30,8 +29,11 @@ def get_datalake_token(
     lib.DataLakeCredential
         A lib.DataLakeCredential which can be used to authenticate towards the datalake
     """
-    logger.info("Looking for ways to authenticate with data lake")
-    if dl_service_auth_str:
+
+    if interactive:
+        logger.info("Attempting to use interactive azure authentication")
+        return lib.auth()
```
milesgranger, line 35: I think adding a simple test with patching here an asserting this is called would be sufficient. https://pypi.org/project/pytest-mock/

gordo_components/data_provider/azure_utils.py
```diff
@@ -20,8 +20,7 @@ def get_datalake_token(
     Parameters
     ----------
     interactive
-        If true then fall back to interactive authentication in case
-        `dl_service_auth_str` is empty
+        If true then use interactive authentication
```
milesgranger, line 23: While you're in here,could you add type defs to the docs? `interactive: bool` for example

gordo_components/data_provider/azure_utils.py
EinarJohnsen: Fixed

gordo_components/data_provider/azure_utils.py
EinarJohnsen: and Fixed

flikka: I guess also type hint to the actual method? like 
`def get_datalake_token(
    interactive:bool =False, dl_service_auth_str:str = None
) -> lib.DataLakeCredential:`

EinarJohnsen:
Should we have a discussion about this issue? @epa095

epa095 (approved):
Lovely!

milesgranger:
@EinarJohnsen When you rebase before merging, be sure to reword your commit to less than 50 chars.

Issue #283, opened 2019-05-27
Align feedforward_autoencoder and lstm_autoencoder scripts


Natalie-Caruana:
Currently there are some differences between the scripts feedforward_autoencoder.py and lstm_autoencoder.py
1. for example the feedforward_autoencoder uses enc_dim as argument for the encoding layers whereas the lstm_autoencoder
uses encoding_dim (similarly for dec_dim, enc_func, dec_func).  
Need to harmonize such argument names.
2. the encoding dimension is a list in feedforward_autoencoder.py and a tuple in lstm_autoencoder.py (same for dec_dim, enc_func, dec_func).  Need to change the lists to tuples.

It would also be useful
1. to raise an error if the encoding_dim is a string (same for dec_dim, enc_func, dec_func) instead of tuple.
2. to add some logging to warn the user of inputting the incorrect argument names (as these will be dumped in **kwargs and ignored, hence the default values will be used instead)

Check that example notebooks are aligned with the changes.

ryanjdillon:
>1. to raise an error if the encoding_dim is a string (same for dec_dim, enc_func, dec_func) instead of tuple.

This seems like it could be tied into [Erik's comment](https://github.com/equinor/gordo-components/issues/130#issuecomment-467005835) over on #130, maybe in its own issue

ryanjdillon:
We will check that passing an unknown keras optimizer kwarg in `optimizer_kwargs` in the kwargs won't break anything. The goal being to start passing `decay` in `optimizer_kwargs`, and perhaps other algo args not in the gordo API.

[Most, but not all](https://github.com/keras-team/keras/blob/master/keras/optimizers.py), optimizers accept `decay` as an arg.

ryanjdillon:
> 1. to raise an error if the encoding_dim is a string (same for dec_dim, enc_func, dec_func) instead of tuple.
> 2. to add some logging to warn the user of inputting the incorrect argument names (as these will be dumped in **kwargs and ignored, hence the default values will be used instead)

I didn't close this before, because I did not implement these things. Is this still something we would like to do @Natalie-Caruana?

PR #429, opened 2019-08-23
Expose `optimizer` and `optimizer_kwargs` in feedforward_model factory

ryanjdillon:
This just copies over the implementation in `lstm_model()` from `lstm_autoencoder.py` for `feedforward_autoencoder.py`. 

As such, a default optimizer is set to `adam`, with any configuration in `optimizer_kwargs` being passed to `Optim(**optimizer_kwargs)` at initialization.

This is part of the needed work to harmonize the api for these two functions, as mentioned in #283.

### DIFF
--- a/gordo_components/model/factories/feedforward_autoencoder.py
+++ b/gordo_components/model/factories/feedforward_autoencoder.py
@@ -1,7 +1,8 @@
 # -*- coding: utf-8 -*-
 
-from typing import Tuple, Dict, Any
+from typing import Tuple, Dict, Any, Union
 
+from keras.optimizers import Optimizer
 from keras import regularizers
 from keras.layers import Dense
 import keras
@@ -23,6 +24,8 @@ def feedforward_model(
     decoding_dim: Tuple[int, ...] = (64, 128, 256),
     decoding_func: Tuple[str, ...] = ("tanh", "tanh", "tanh"),
     out_func: str = "linear",
+    optimizer: Union[str, Optimizer] = "adam",
+    optimizer_kwargs: Dict[str, Any] = dict(),
     compile_kwargs: Dict[str, Any] = dict(),
     **kwargs,
 ) -> keras.models.Sequential:
@@ -44,7 +47,16 @@ def feedforward_model(
     decoding_func: tuple
         Activation functions for the decoder part.
     out_func: str
-        Activation function for the output layer.
+        Activation function for the output layer
+    optimizer: Union[str, Optimizer]
+        If str then the name of the optimizer must be provided (e.x. "adam").
+        The arguments of the optimizer can be supplied in optimize_kwargs.
+        If a Keras optimizer call the instance of the respective
+        class (e.x. Adam(lr=0.01,beta_1=0.9, beta_2=0.999)).  If no arguments are
+        provided Keras default values will be set.
+    optimizer_kwargs: Dict[str, Any]
+        The arguments for the chosen optimizer. If not provided Keras'
+        default values will be used.
     compile_kwargs: Dict[str, Any]
         Parameters to pass to ``keras.Model.compile``.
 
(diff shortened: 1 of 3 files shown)


gordo_components/model/factories/feedforward_autoencoder.py
```diff
@@ -42,6 +45,15 @@ def feedforward_model(
         Activation functions for the decoder part
     out_func: str
         Activation function for the output layer
+    optimizer: Union[str, Optimizer]
+        If str then the name of the optimizer must be provided (e.x. "adam").
+        The arguments of the optimizer can be supplied in optimization_kwargs.
```
milesgranger: `optimizer_kwargs`

gordo_components/model/factories/feedforward_autoencoder.py
```diff
@@ -120,6 +138,15 @@ def feedforward_symmetric(
          Must have len > 0
     funcs: List[str]
         Activation functions for the internal layers
+    optimizer: Union[str, Optimizer]
+        If str then the name of the optimizer must be provided (e.x. "adam").
+        The arguments of the optimizer can be supplied in optimization_kwargs.
```
milesgranger, line 137: `optimizer_kwargs`

milesgranger (commented):
Think it looks pretty ok; but is there not a lot of duplicate code now between these two modules?

gordo_components/model/factories/feedforward_autoencoder.py
ryanjdillon: Yes indeed. I'll be back!

gordo_components/model/factories/feedforward_autoencoder.py
```diff
@@ -86,10 +98,14 @@ def feedforward_model(
         model.add(Dense(units=units, activation=activation))
 
     # Final output layer
```
milesgranger: This comment doesn't seem relevant to this line, should probably move it below this conditional.

gordo_components/model/factories/feedforward_autoencoder.py
```diff
@@ -86,10 +98,14 @@ def feedforward_model(
...
+    if isinstance(optimizer, str):
+        Optim = getattr(keras.optimizers, optimizer)
+        optimizer = Optim(**optimizer_kwargs)
+
     model.add(Dense(n_features_out, activation=out_func))
 
     # Set some pre-determined default compile kwargs.
-    compile_kwargs.setdefault("optimizer", "adam")
+    compile_kwargs.update({"optimizer": optimizer})
     compile_kwargs.setdefault("loss", "mean_squared_error")
```
milesgranger, line 103: Know you didn't write this, but maybe we should default to whatever `Keras` is (ie, not actually set a default), because that's probably where we'll refer people to documentation about optimizer and compile kwargs. If they see the default is something else, that's what I would expect it to be.

milesgranger (commented):
Very nice, feel free to rebase/squash into the desired commits after addressing my small comments and can probably merge after that. :+1:

gordo_components/model/factories/feedforward_autoencoder.py
ryanjdillon: That sounds proper to me. Should I just go ahead and do that or get a consensus from folks?

epa095: From [this](https://keras.io/models/sequential/) doc-string it seems like `optimizer` is a non-optional argument for model.compile, i.e. it does not have a default value. But I think we should have a default, so then adam is maybe as good a default as any?

ryanjdillon: @natalie-caruana @flikka Does "adam" sound good to you all here?

milesgranger: Sorry, I was referring to `loss` specifically; but the consensus of `optimizer` remains. :+1:

ryanjdillon: It looks like these are [required args](https://keras.io/losses/) for `model.compile`.

Natalie-Caruana: Adam optimizer as default should be fine and for the loss function mse looks good to me too.

gordo_components/model/factories/feedforward_autoencoder.py
milesgranger: `optimizer_kwargs`

milesgranger (approved):

Issue #130, opened 2019-02-22
Change feedforward_symetric/feedforward_model to not have as many optional parameters


epa095:
Today it has the following optional parameters
```
 enc_dim: List[int] = None,
 dec_dim: List[int] = None,
 enc_func: List[str] = None,
 dec_func: List[str] = None,
```
But when #128 is merged we will have builders which are more natural to use when you don't know the dimensions you want, so then it is maybe better to make these parameters non-optional. Will maybe break some stuff, but who cares. 

Relates to #127

epa095:
This should be discussed a tiny bit before starting it. The error validation is a no-brainer, so that is good, but maybe not the non-optionality. I no longer see any strong reasons for making them compolsory.

ryanjdillon:
I can pick up all of these API dealies. Maybe we could have said discussion i dag and roll them all up into one issue.

PR #431, opened 2019-08-27
Make autoencoder API parameters match and add some input validation

ryanjdillon:
In this PR the following was done:
* enc_* and dec_* parameters in the feedforward factory were renamed to match the lstm factory arguments encoding_*, decoding_*
* Extra validation of input was added in factory utils to give a warning if only one of these layer configuraton parameters is passed.
* Defaults are now set after initial checking the passing of a single layer configuration parameter.
* Other validation statements in factories are now moved to utils.
* Tests were added to check these


gordo_components/model/factories/feedforward_autoencoder.py
```diff
@@ -1,24 +1,27 @@
...
+from typing import Tuple, Dict, Any
 
 from keras import regularizers
 from keras.layers import Dense
 import keras
 
 from keras.models import Sequential as KerasSequential
 from gordo_components.model.register import register_model_builder
-from gordo_components.model.factories.model_factories_utils import hourglass_calc_dims
+from gordo_components.model.factories.model_factories_utils import (
```
milesgranger: To follow the same pattern, you can just name this `utils`; as that's what we have in most of our component modules.

gordo_components/model/factories/lstm_autoencoder.py
```diff
@@ -71,17 +74,13 @@ class (e.x. Adam(lr=0.01,beta_1=0.9, beta_2=0.999)).  If no arguments are
     """
     n_features_out = n_features_out or n_features
 
-    if len(encoding_dim) != len(encoding_func):
-        raise ValueError(
-            f"Number of layers ({len(encoding_dim)}) and number of functions ({len(encoding_func)}) "
-            f"must be equal for the encoder."
-        )
+    encoding_dim, encoding_func = set_dim_func(
+        encoding_dim, encoding_func, (256, 128, 64), ("tanh", "tanh", "tanh")
```
milesgranger: Rather than having the defaults down here, why not have them set in the function signature?

gordo_components/model/factories/model_factories_utils.py
```diff
@@ -15,3 +16,36 @@ def hourglass_calc_dims(compression_factor, encoding_layers, n_features):
         round(n_features - (i * average_slope)) for i in range(1, encoding_layers + 1)
     ]
     return dims
+
+
+def _check_if_only_dim_or_func(dim, func):
```
milesgranger: This PR isn't marked with `WIP` so I'll go ahead and say that for all these functions, we'll need type hints and numpy style docs for them.

tests/gordo_components/model/test_model_factories_utils.py
```diff
@@ -1,25 +1,55 @@
...
+
+def test_check_dim_func_len():
+    # Raise a ValueError if the len of dim and func are not equal
+    with pytest.raises(ValueError):
+        _check_dim_func_len(dim=(256,128), func=("tanh", "tanh", "tanh"))
+
+
+@pytest.mark.parametrize("dim", ((256, 128, 64), (), (256, 128, 64)))
+@pytest.mark.parametrize("func", (("tanh", "tanh", "tanh"), ("tanh", "tanh", "tanh"), ()))
+def test_set_dim_func(dim, func):
```
milesgranger: Short human readable description of the test goal, pretty please.

tests/gordo_components/model/test_model_factories_utils.py
```diff
@@ -1,25 +1,55 @@
...
+    dims = hourglass_calc_dims(0.5, 3, 10)
+    assert dims == [8, 7, 5]
+    dims = hourglass_calc_dims(0.5, 3, 3)
+    assert dims == [3, 2, 2]
+    dims = hourglass_calc_dims(0.3, 3, 10)
+    assert dims == [8, 5, 3]
+    dims = hourglass_calc_dims(1, 3, 10)
+    assert dims == [10, 10, 10]
+    dims = hourglass_calc_dims(0, 3, 100000)
+    assert dims == [66667, 33334, 1]
```
milesgranger: Nice, but I think this is a prime candidate for `input, expected` parameterization.

milesgranger (commented):
Looking pretty good.

You'll need to reword all your commits to be in the imperative style we use. [good reference here](https://chris.beams.io/posts/git-commit/)

gordo_components/model/factories/lstm_autoencoder.py
ryanjdillon: Yeah, this didn't feel great. This was done to address the comment mentioned [here](https://github.com/equinor/gordo-components/issues/130#issuecomment-467005835). TL;DR Passing only dim or func with a different number of layers from the default will break. 

My solution here was to first check this and throw a warning if only one is passed, set the default to the one that wasn't passed, and then allow it to break as before if they (the one passed and the default used for the other) are of unequal length.

tests/gordo_components/model/test_model_factories_utils.py
ryanjdillon: This was the one test method for the pre-existing utils module that I put my routines in. Should I rewrite things like this as I come across them and include them in my PR, or maybe create a new issue/PR with things I uncover as I work on something else?

gordo_components/model/factories/feedforward_autoencoder.py
ryanjdillon: This goes along with the last comment, but this was already there. Is it best for me to rename it as I come across it and include it in my PR, create a new issue/PR, etc?

gordo_components/model/factories/lstm_autoencoder.py
ryanjdillon: Going to roll this back and keep them in the signature with a less descriptive error if one not passed. :+1:

tests/gordo_components/model/test_model_factories_utils.py
ryanjdillon: I've included this in this PR, but just let me know if I should pull it out to a different PR.

milesgranger: ya, just because it's a pre-existing condition doesn't mean we should repeat it. ;-)

gordo_components/model/factories/lstm_autoencoder.py
```diff
@@ -41,16 +44,16 @@ def lstm_model(
...
+    encoding_dim: tuple
+        Tuple of numbers with the number of neurons in the encoding part.
+    decoding_dim: tuple
+        Tuple of numbers with the number of neurons in the decoding part.
+    encoding_func: tuple
         Activation functions for the encoder part.
-    decoding_func: list
+    decoding_func: tuple
         Activation functions for the decoder part.
     func_output: str
```
milesgranger: I know you didn't change this, but just noticed it now; this parameter doesn't exist. Should be renamed to `out_func` I suppose.

gordo_components/model/factories/lstm_autoencoder.py
```diff
@@ -117,8 +111,8 @@ def lstm_symmetric(
     n_features: int,
     n_features_out: int = None,
     lookback_window: int = 1,
-    dims: Tuple[int, ...] = (256, 128, 64),
-    funcs: Tuple[str, ...] = ("tanh", "tanh", "tanh"),
+    dims: Tuple[int, ...] = (),
+    funcs: Tuple[str, ...] = (),
```
milesgranger: Should have the default values here. 
Right now, all of our `Keras..` models can be initialized and _"just work"_ by supplying the `kind` parameter (and we'll eventually have a default for that as well, to be more scikit-learn players) 

However, removing defaults here will cause the construction of the model to fail because of the first check in this very function.

gordo_components/model/factories/utils.py
```diff
@@ -15,3 +16,13 @@ def hourglass_calc_dims(compression_factor, encoding_layers, n_features):
         round(n_features - (i * average_slope)) for i in range(1, encoding_layers + 1)
     ]
     return dims
+
+
+def check_dim_func_len(prefix: str, dim: Tuple[int, ...], func: Tuple[str, ...]):
```
milesgranger, line 45: docs plz

gordo_components/model/factories/utils.py
```diff
@@ -15,3 +16,13 @@ def hourglass_calc_dims(compression_factor, encoding_layers, n_features):
...
+    if len(dim) != len(func):
+        raise ValueError(
+            f"The length (i.e. the number of network layers) of {prefix}_dim "
+            f"({len(dim)}) and {prefix}_func ({len(func)}) must be equal. If only "
+            f"{prefix}_dim or {prefix}_func was passed, ensure that its length matches "
+            "the that of the {prefix} parameter not passed."
```
milesgranger: "the that of the"

tests/gordo_components/model/test_lstm_autoencoder.py
```diff
@@ -18,8 +19,13 @@ def test_lstm_symmetric_checks_dims(self):
         """
         Test that lstm_symmetric validates parameter requirements
         """
+        # Raise an error if empty dimension and function layers defined.
         with self.assertRaises(ValueError):
-            lstm_symmetric(4, dims=[], funcs=[])
+            lstm_symmetric(4, dims=(), funcs=())
+
```
milesgranger, line 44: Additionally, you might also guarantee that something like `lstm_symmetric(4)` would work. As mentioned in a previous comment, a bit strange that the default args would illicit a `ValueError` in this function.

gordo_components/model/factories/lstm_autoencoder.py
ryanjdillon: It seems like `out_func` is used in the derivative functions where one only designates the function to be used and the dimensions are calculated from the inputs. 

Here it seems like `decoding_func` is being used [here](https://github.com/ryanjdillon/gordo-components/blob/dimension_function_error_checking/gordo_components/model/factories/lstm_autoencoder.py#L89-L92), just getting pulled out as `activation`.

```python
 89     # decoding layers
 90     for i, (n_neurons, activation) in enumerate(zip(decoding_dim, decoding_func)):
 91         return_seq = i != len(decoding_dim) - 1  # Don't return sequences on 2nd to last
 92         model.add(LSTM(n_neurons, activation=activation, return_sequences=return_seq))
```

gordo_components/model/factories/lstm_autoencoder.py
ryanjdillon: Thanks Miles, I missed this one. Strange the tests passed... I'll take a look there and see if I can add something that would make that fail next time.

gordo_components/model/factories/lstm_autoencoder.py
```diff
@@ -41,16 +44,16 @@ def lstm_model(
         One timestep = current observation in the sample.
         Two timesteps = current observation + previous observation in the sample.
         ...
-    encoding_func: list
+    encoding_dim: tuple
+        Tuple of numbers with the number of neurons in the encoding part.
```
EinarJohnsen, line 48: Minor detail, but here the documentation is using period at the end and in the ```gordo_components/model/factories/feedforward_autoencoder.py```, the documentation for encoding/decoding does not use a period at the end. Maybe we should change to one or the other?

ryanjdillon: Grammatically speaking, since these aren't complete sentences, there shouldn't be a period; though, I am not sure what is considered standard practice here.

tests/gordo_components/model/test_lstm_autoencoder.py
ryanjdillon: I ended up landing on comparing the each layer of the different LSTM models' (with deafult parameters) configurations via the `layer.get_config()` method. 

This then ensure that the models pass default config muster and also that they are the same with default config.

gordo_components/model/factories/lstm_autoencoder.py
milesgranger: We've talked about this a bit, and I agree we haven't settled and it's a bit _lazy_ at times. However, we strive to follow the numpy style of documentation, which tends to be written in sentence format. But I think we're all guilty of slipping here and there.

ryanjdillon: Sounds good to me :+1: I love the numpy docstring format, and if complete sentences are sanctioned by them then I think that is more then enough reason to do that.

ryanjdillon: It doesn't seem most of these are written as complete sentences. I'll just give it a whack and see if you all agree with my rephrasing of things.

ryanjdillon: In taking a look at things, it seems like numpy does the same as we are doing here.

[Theirs](https://github.com/numpy/numpy/blob/master/numpy/core/numeric.py#L123):
```
out : ndarray
    Array of zeros with the same shape and type as `a`.
```

Ours:
```
encoding_dim: tuple
    Tuple of numbers with the number of neurons in the encoding part.
```

So here they are using an incomplete sentence with a period. Erik likes the period as it indicated the end of the definition, which I agree is a useful thing. 

In this case, I'll leave the definitions all the same, and just add `.`s where missing in those files I've handled here in this PR.

gordo_components/model/factories/lstm_autoencoder.py
```diff
@@ -71,17 +74,8 @@ class (e.x. Adam(lr=0.01,beta_1=0.9, beta_2=0.999)).  If no arguments are
...
-            f"must be equal for the encoder."
-        )
-
-    if len(decoding_dim) != len(decoding_func):
-        raise ValueError(
-            f"Number of layers ({len(decoding_dim)}) and number of functions ({len(decoding_func)}) "
-            f"must be equal for the decoder."
-        )
+    check_dim_func_len("encoding", encoding_dim, encoding_func)
+    check_dim_func_len("decoding", decoding_dim, decoding_func)
```
milesgranger, line 78: Nice. :+1:

gordo_components/model/factories/lstm_autoencoder.py
```diff
@@ -166,18 +160,19 @@ class (e.x. Adam(lr=0.01,beta_1=0.9, beta_2=0.999)).  If no arguments are
...
+        n_features_out=n_features_out,
+        lookback_window=lookback_window,
+        encoding_dim=dims,
+        decoding_dim=dims[::-1],
+        encoding_func=funcs,
+        decoding_func=funcs[::-1],
+        out_func=out_func,
+        optimizer=optimizer,
+        optimizer_kwargs=optimizer_kwargs,
+        compile_kwargs=compile_kwargs,
```
milesgranger, line 175: I like the use of the keyword arguments. :1st_place_medal:

gordo_components/model/factories/utils.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
+
+import math
+from typing import Tuple
+
+
+def hourglass_calc_dims(compression_factor, encoding_layers, n_features):
```
milesgranger: Need type hints, return type and numpy docs plz

gordo_components/model/factories/utils.py
milesgranger: Numpy style docs.

tests/gordo_components/model/test_factories_utils.py
```diff
@@ -0,0 +1,38 @@
...
+        ((1, 3, 10), [10, 10, 10]),
+        ((0, 3, 100000), [66667, 33334, 1]),
+    ],
+)
+def test_hourglass_calc_dims_check_dims(test_input, test_expected):
+    """
+    Test that hourglass_calc_dims implements the correct dimensions
+    """
+    dims = hourglass_calc_dims(*test_input)
+    assert dims == test_expected
```
milesgranger, line 27: LoooOoOoove it! :heart:

milesgranger (commented):
Very nice sir, I think after you adjust the few comments I have on docs, you can rebase/squash this into 2-3 commits?

ryanjdillon:
> Very nice sir, I think after you adjust the few comments I have on docs, you can rebase/squash this into 2-3 commits?

All doc'ed and squash'ed :+1:

gordo_components/model/factories/utils.py
```diff
@@ -0,0 +1,64 @@
...
+
+def check_dim_func_len(prefix: str, dim: Tuple[int, ...], func: Tuple[str, ...]):
+    """
+    Check that the number of layer dimensions and layer functions are equal
+
+    Parameters
+    ----------
+    prefix: str
+        Parameter name prefix for error message generation (Options: "encoding" or "decoding").
+    dim: tuple of int
```
milesgranger, line 53: `Tuple[int, ...]`

milesgranger:
You're also going to have to reword your first two commits so they aren't clipped. <~50 chars I think.

milesgranger (approved):
LGTM

PR #433, opened 2019-08-29
Remove explicit threads=None argument from DataLakeProvider __init__

### LINKED ISSUE #432: NCS reader should default to 1 thread if None is sent
Currently, if threads argument to ncs_reader is explicitly set to None, we get a bunch of threads, determined by the TreadPoolExecutor. Set it to 1 thread, if None is sent in.

flikka:
This fixes a bug where the NcsReader thread default was never used,
it came in as None, resulting in an unknown number of threads being used.

Close #432


milesgranger (approved):
VGPOTFTIAI!

PR #435, opened 2019-08-30
Fix overwrite of y in KerasLSTMAutoEncoder.fit

milesgranger:
Problem :alarm_clock: 
-----------
After refactoring the models to take a `y`, a mistake was made by me, in which during the fitting, the original `y` was overwritten with what was suppose to be the primer to create the underlying model.

Solution :bulb: 
-----------
Give that `y` a different name, explicit that it is _not_ the original `X` and `y` it should be fitting on.


epa095:
> ... a mistake was made by me, ...

By **us** miles, us.

epa095 (approved):
Loovely

PR #437, opened 2019-08-30
Support single level dataframes in serialization funcs

milesgranger:
Problem :alarm_clock: 
-----------
Sending a multi-level column dataframe to the server, while serving a model with `ColumnTransformer` will cause it to fail.

Solution :checkered_flag: 
-----------
Trying to _only_ accept single level dataframes, causes problems when `.to_dict()` on a dataframe with `datatime` index, is not JSON serializable.

Fortunately, this problem is already addressed in the serialization functions of multi-level dataframes. Therefore, updating those functions to serialize and deserialize single level dataframes solves both issues. And thus have renamed those functions to `dataframe_from/to_dict` 

Bonus :gift: 
--------
While running the gordo test project, the `RandomForestRegressor` as a `DiffBasedAnomalyDetector` failed with 
```json
{"message": "y is not a subset of X, cannot do anomaly detection"}
```
But this is no longer valid; for example, our `DiffBasedAnomalyDetector` only compares the `y` against the model output. So that check in the `AnomalyView` was removed.

### DIFF
--- a/gordo_components/client/client.py
+++ b/gordo_components/client/client.py
@@ -386,8 +386,8 @@ class Client:
         """
 
         json = {
-            "X": server_utils.multi_lvl_column_dataframe_to_dict(X.iloc[chunk]),
-            "y": server_utils.multi_lvl_column_dataframe_to_dict(y.iloc[chunk])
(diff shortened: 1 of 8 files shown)


epa095 (approved):
Lovely

PR #442, opened 2019-09-06
Switch to CircleCI and push to Dockerhub

milesgranger:


milesgranger:
@epa095 @flikka About ready, but perhaps we should create a temp `rc` branch (and modify this config to also do the deploy workflows on branch `master` and `rc`).  After we ensure it works there when tagging a dummy release, might be more comfortable to merge that into `master`?

.circleci/config.yml
```diff
@@ -0,0 +1,314 @@
...
+            - python-docs-cache-{{ checksum "requirements.txt" }}-{{ checksum "dev_requirements.txt"}}
+      - run:
+          name: Install Dependencies
+          command: |
+            python -m venv ~/venv
+            source ~/venv/bin/activate
+            pip install -r requirements.txt
+            pip install -r dev_requirements.txt
+      - save_cache:
+          key: python-docs-cache-{{ checksum "requirements.txt" }}-{{ checksum "dev_requirements.txt"}}
```
epa095, line 121: I don't know if it will matter, but you have several caches which it seems like should be possible to optimize a bit:
1.  `python-docs-cache` with ` ~/venv`
2. `v1-with-test-eggs-cache` with `~/code/venv` and `~/code/.eggs`
3. `v1-cache` with `~/code/venv`
It probably possible to restore 2 caches, so then point 2 above can be converted to just `~/code/.eggs` maybe? And can `~/code/venv` and ` ~/venv` be the same?

.circleci/config.yml
```diff
@@ -0,0 +1,314 @@
+
+version: 2.1
+
+executors:
+  python-executor:
+    working_directory: ~/code
+    machine:
+      image: "circleci/classic:201808-01"
```
epa095, line 9: `      docker_layer_caching: true`?

milesgranger:
@epa095 Tested on the other branch and we're now pushing to Dockerhub and PyPi seems like it'll work out without trouble. Authenticates but failed with invalid version (when attempting to push a non-tagged version)

And reduced the Travis config to just make and push images.

So I think it's good for any more comments.

flikka (commented):
Looks good to me, nice job. But I am but a mere mortal, not a fully qualified master of the API driven cloud based continuous integration and deployment universe.

.circleci/config.yml
```diff
@@ -0,0 +1,310 @@
...
+            - install
+      - test-allelse:
+          requires:
+            - install
+      - make-images:
+          filters:
+            branches:
+              only: /.*/
+            tags:
+              only: /.*/
```
epa095, line 281: Dont we want to try making images always, but only push them on tags?

docker_push.sh
```diff
@@ -58,14 +58,14 @@ else
...
 
-docker tag $DOCKER_IMAGE $DOCKER_REGISTRY/$DOCKER_NAME:$git_sha$suffix
-docker push $DOCKER_REGISTRY/$DOCKER_NAME:$git_sha$suffix
+docker tag $DOCKER_IMAGE $DOCKER_REPO/$DOCKER_NAME:$git_sha$suffix
+docker push $DOCKER_REPO/$DOCKER_NAME:$git_sha$suffix
 
 git tag --points-at HEAD | while read -r tag ; do
-    docker tag $DOCKER_IMAGE $DOCKER_REGISTRY/$DOCKER_NAME:$tag$suffix
-    docker push $DOCKER_REGISTRY/$DOCKER_NAME:$tag$suffix
+    docker tag $DOCKER_IMAGE $DOCKER_REPO/$DOCKER_NAME:$tag$suffix
```
epa095, line 69: This looks a bit weird to me. Before with DOCKER_REGISTRY=auroradevacr.azurecr.io and DOCKER_NAME=gordo-components/gordo-model-builder this would expand to `auroradevacr.azurecr.io/gordo-components/gordo-model-builder`, but now it expands to `gordo-components/gordo-model-builder`. I see the variable `DOCKER_REGISTRY` set in travis/circleci, but I dont see how it is used.

epa095 (approved):
Minor stuff, but in generall LGTM. I propose you merge it when you are at work, and not the moment you land at the airport :-)

.circleci/config.yml
milesgranger: Ya it looks a bit awkward, but for any steps which has `filters` such as `push-images` any step it depends on must also specify an equal-to-or-looser equivalent filter. Slightly strange quirk.

docker_push.sh
milesgranger: When doing `docker login` in at the start to `DOCKER_REGISTRY` it will automatically push to that one. It is also set in the `push-images` step here for circleci

Issue #443, opened 2019-09-10
Update client


milesgranger:
Problem :fire: 
------------------
We serve models with an output offset (such as LSTMs). This means that when `N` samples are sent to it for prediction, `N - offset/lookback` are returned. This causes troubles for the client when asking for predictions in a given time range. Because in fact, it'll be a subset of that time-range after taking into account the offset returned by these models.

Steps to success :man_playing_water_polo: 
-----------------------

- [x] During model building, determine what offset the model has. (https://github.com/equinor/gordo-components/pull/444)
- [x] When running the client, move the start time to account for any offset the model might have.(https://github.com/equinor/gordo-components/pull/445)

PR #444, opened 2019-09-10
Add model offset to metadata

milesgranger:
Part of #443 

Problem :radioactive: 
-----------
When serving a model such as LSTMs, we don't have a clear idea that it will return less data than it got in; less looking in depth at the model metadata in an non-deterministic way.

Solution :racehorse: 
-----------
During model building, determine what the offset of the model is, and put that in a predetermined place in the model's metadata. `model-offset`

### DIFF
--- a/gordo_components/builder/build_model.py
+++ b/gordo_components/builder/build_model.py
@@ -8,6 +8,9 @@ import time
 from pathlib import Path
 from typing import Union, Optional, Dict, Any
 
+import pandas as pd
+import numpy as np
+
 from sklearn.base import BaseEstimator
 from sklearn.model_selection import cross_val_score, TimeSeriesSplit
 from sklearn.metrics import explained_variance_score, make_scorer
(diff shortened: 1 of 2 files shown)


gordo_components/builder/build_model.py
```diff
@@ -125,6 +129,28 @@ def build_model(
     return model, metadata
 
 
+def _determine_offset(model: BaseEstimator, X: Union[np.ndarray, pd.DataFrame]) -> int:
```
epa095, line 132: Nice, it did not even cross my mind that we could just do this and calculate the offset instead of having the model tell us about it!

tests/gordo_components/builder/test_builder.py
```diff
@@ -50,6 +58,25 @@ def metadata_check(metadata, check_history):
         )
 
 
+@pytest.mark.parametrize(
```
epa095, line 61: I like it

epa095 (approved):
:100:

gordo_components/builder/build_model.py
```diff
@@ -125,6 +129,28 @@ def build_model(
...
+    """
+    Determine the model's offset. How much does the output of the model differ
+    from its input?
+
+    Parameters
+    ----------
+    model: sklearn.base.BaseEstimator
+        Trained model with either ``predict`` or ``transform`` method, preference
+        given to ``predict``
```
EinarJohnsen: Nitpick, but this could have a punctuation at the end, or remove the one below, depending on whether you think this is a complete sentence or not.

tests/gordo_components/builder/test_builder.py
```diff
@@ -11,15 +11,21 @@
...
+
 import gordo_components
 from gordo_components.builder.build_model import (
     _save_model_for_workflow,
     provide_saved_model,
     _get_metadata,
+    _determine_offset,
 )
 from gordo_components.builder import build_model
 from gordo_components.dataset.sensor_tag import SensorTag
```
EinarJohnsen, line 26: Is there a extra line here? And if so, is it necessary?

milesgranger: Line 26? Nope, it would be green if it was added. :thinking:

EinarJohnsen: Yeah, I would assume, but the ```from gordo_components.model import models``` is added at line 27, no? And by extra line I should have said "gap"

EinarJohnsen (approved):
:ship: :it:

PR #445, opened 2019-09-10
Client adjust for offset

### LINKED ISSUE #443: Update client

milesgranger:
Will close #443 

Problem :question: 
-----------
When the client asks the ML server for predictions, it doesn't know about any offset the model might have and thus doesn't now (or care) that the returned length may be different.

Solution :hammer: 
-----------
#444 Adds `model-offset` to the metadata in a predefined location. This has the client look at that value and 'backup' the initial requested start date by the `resolution` * `model-offset` so that returned predictions will contain the initially requested prediction time range.

Dependencies :handshake: 
-------------------
- #444

### DIFF
--- a/gordo_components/client/client.py
+++ b/gordo_components/client/client.py
@@ -203,6 +203,9 @@ class Client:
                 resolution=data["endpoint-metadata"]["metadata"]["dataset"][
                     "resolution"
                 ],
+                model_offset=data["endpoint-metadata"]["metadata"]["model"].get(
+                    "model-offset", 0
+                ),
             )
             if data["healthy"]
             else EndpointMetadata(
@@ -212,6 +215,7 @@ class Client:
                 tag_list=None,
                 target_tag_list=None,
                 resolution=None,
+                model_offset=None,
             )
             for data in resp.json()["endpoints"]
         ]
@@ -618,6 +622,15 @@ class Client:
             Dataframe of required tags and index reflecting the datetime point
         """
 
+        # We want to adjust for any model offset. If the model outputs less than it got in, it requires
+        # extra data than what we're being asked to get predictions for.
+        # just to give us some buffer zone.
+        start = self._adjust_for_offset(
+            dt=start,
+            resolution=endpoint.resolution,
+            n_intervals=endpoint.model_offset + 5,
+        )
+
         dataset = TimeSeriesDataset(  # type: ignore
             data_provider=self.data_provider,
             from_ts=start,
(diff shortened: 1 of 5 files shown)


gordo_components/client/client.py
```diff
@@ -631,6 +644,37 @@ def predict(
...
+            Initial datetime to adjust
+        resolution: str
+            A string code capable of being parsed by :meth::`pandas.Timedelta`
+        n_intervals: int
+            Number of resolution steps to take earlier than the given date
+
+        Returns
+        -------
+        datetime
+            The new offset datetime object.
```
EinarJohnsen, line 666: Nitpick: Punctuation at the return, but not at the parameters.

milesgranger: Added punctuation to the prior params. :capital_abcd:

gordo_components/client/client.py
```diff
@@ -203,6 +203,9 @@ def _endpoints_from_watchman(self, endpoint: str) -> typing.List[EndpointMetadat
                 resolution=data["endpoint-metadata"]["metadata"]["dataset"][
                     "resolution"
                 ],
+                model_offset=data["endpoint-metadata"]["metadata"]["model"][
+                    "model-offset"
```
epa095: As a mild backwards-compability should we maybe have it at 0 if this is not set?

gordo_components/client/client.py
```diff
@@ -618,6 +622,15 @@ def predict(
             Dataframe of required tags and index reflecting the datetime point
         """
 
+        # We want to adjust for any model offset. If the model outputs less than it got in, it requires
+        # extra data than what we're being asked to get predictions for. Double the expected model offset
```
epa095: Why double compared to e.g. "+ 5"? Why should no offset mean 0 buffer, but 144 offset should mean 288 buffer?

milesgranger: Would `+ 5` be more acceptable? I did double of the offset for no particular reason, but can see with an offset of 0, it isn't an optimal choice indeed.

gordo_components/client/client.py
milesgranger: Good catchy!

gordo_components/client/client.py
epa095: I think some kind of constant offset is more reasonable yeah. But when we change the dataset so it goes back and gets data so it can deliver data from the actual start-date then I guess we dont need this anymore?

epa095 (approved):
Lgtm

PR #450, opened 2019-09-18
Use explicit docker registry

milesgranger:
Problem :face_with_head_bandage: 
-----------
Pushing to `docker.io` is the default, but even after logging into a different registry, it will still attempt to push to `docker.io`

Solution :checkered_flag: 
-----------
Go back to using explicit `DOCKER_REGISTRY` when tagging and pushing.

@epa095 was right all along.. :open_mouth:

### DIFF
--- a/docker_push.sh
+++ b/docker_push.sh
@@ -29,6 +29,11 @@
 #                  Defaults to false
 # DOCKER_REPO: The docker repository of concern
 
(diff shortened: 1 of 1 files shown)


epa095 (approved):
LGTM, except compared to before there is no longer a default. Maybe add a check if "DOCKER_REGISTRY" is defined, and if not, exit?

epa095 (approved):
LGTM

PR #451, opened 2019-09-18
Clean up tests / CI

milesgranger:
Problems :skull_and_crossbones: 
-----------

- Uploading code coverage on each tests cost some time, maybe ~1-2min per in some bad cases.
- Client's get/post json tests use httpbin dockerfile, which we don't like
- Model tests by far take the longest, ~5-7mins

Solution :nail_care: 
-----------

- [x] Collect the reports into a `workspace` and then upload all at once
- [x] Remove those client post/fetch json tests
~Optimize the model component testing~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,9 +51,26 @@ commands:
             source ~/code/venv/bin/activate
             python setup.py test<< parameters.for >>
       - run:
-          name: Upload Test Coverage
-          command: bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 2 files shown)


milesgranger:
@epa095 I'll actually sit out the last bit, trying to shorten the `testmodel` tests; think that might end up in a bigger change that I would like for this PR.

Feel free to comment on what I have here.

epa095 (approved):
Lgtm

PR #455, opened 2019-09-20
Upgrade pip before install starts

milesgranger:
Problem :night_with_stars: 
---
image

Solution :sunrise: 
---
Add step before install begins to upgrade pip, which results in substantially faster `Requirement already satisfied...` checking. Down to ~8sec over 20 some odd seconds per install and test requirements.


epa095 (approved):
Lgtm

PR #456, opened 2019-09-20
Move workflow_generator and config_elements into components

milesgranger:
Problem :carrot: 
-----
We've talked about it before, workflow generator, generally fitting as a component itself to the "gordo" workflow. It's also a cost to explain to data scientists / power users, that if they want to trial gordo-components, they'd have to update multiple versions here and there, vs only concerning themselves with the version of gordo-components they're working with.

Solution :cake: 
---
Move `workflow_generator` and `config_elements` into components.

### DIFF
--- a/examples/config.yaml
+++ b/examples/config.yaml
@@ -1,0 +1,65 @@
+machines:
+
+  - name: ct-23-0001 #1st machine
+    dataset:
+      tags: #list of tags for 1st machine
+        - TAG1
+        - TAG2
+        - TAG3
+      train_start_date: 2016-11-07T09:11:30+01:00
+      train_end_date: 2018-09-15T03:01:00+01:00
+    runtime:
+      server:
+        resources:
+          requests: # The resources we require reserved
+            memory: 1700 # Request 1.7G of memory
+            cpu: 2000 # Request two cores, while e.g. 250 would mean a quarter core
+          limits: # The resources we are limited to using.
+            memory: 2000 #Kill the server if it uses more than 2GB
+            cpu: 2000 #Throttle the server to never using more than 2 cores
+
+
+  - name: ct-23-0002 #2nd machine
+    dataset:
+      resolution: 2T
+      tags: #list of tags for 2nd machine
+        - TAG1
+        - TAG2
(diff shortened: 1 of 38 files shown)


examples/config.yaml
```diff
@@ -0,0 +1,65 @@
+machines:
+
+  - name: ct-23-0001 #1st machine
+    dataset:
+      tags: #list of tags for 1st machine
+        - GRA-TE  -23-0733.PV
```
flikka: Unfortunately, we cant put tag names out there (yet). I know we may have some already, but they are not open.

examples/ioc_config_example.csv
```diff
@@ -0,0 +1,6 @@
+Hierarchy_L1|Hierarchy_L2|Hierarchy_L3|Hierarchy_L4|Hierarchy_L5|Tag_Type|Tag_Source_Address
```
flikka: What is this file doing here?

resources/grafana/dashboards/how_to_modify_dashboard.md
```diff
@@ -0,0 +1,5 @@
+To change the dashboard:
```
flikka, line 1: Should this file be here?

resources/secrets_templates/gordo-config-secrets.yaml
```diff
@@ -0,0 +1,27 @@
+# Data Lake secret
```
flikka: Should this file be here?

tests/gordo_components/config_elements/test_config_elements.py
```diff
@@ -0,0 +1,179 @@
...
+    def test_from_config(self):
+        """
+        Test ability to create a Dataset from a config element
+        """
+        element_str = """
+            name: ct-23-0002
+            dataset:
+              resolution: 2T
+              tags:
+                - GRA-YE  -23-0751X.PV
```
flikka: Tag names are not open

flikka (commented):
Overall. INteresting. But why are there so other many files?

PR #457, opened 2019-09-23
Add more explicit docs about resolution strings

milesgranger:
Problem :alarm_clock: 
---
Our current documentation about `resolution` is pretty obscure; what is the format of these magical strings?

Solution :unicorn: 
---
Add link to where/what strings are valid.


epa095 (approved):
goodie doccie

PR #458, opened 2019-09-23
Expose evaluation from config in model builder

EinarJohnsen:
Let the user specify the evaluation key in the config file to change ```cv_mode``` and have the possibility to add more parameters.

### DIFF
--- a/gordo_components/builder/build_model.py
+++ b/gordo_components/builder/build_model.py
@@ -38,7 +38,7 @@ def build_model(
     model_config: dict,
     data_config: Union[GordoBaseDataset, dict],
     metadata: dict,
-    cv_mode: str = "full_build",
+    evaluation_config: dict = {"cv_mode": "full_build"},
 ) -> Tuple[Union[BaseEstimator, None], dict]:
     """
     Build a model and serialize to a directory for later serving.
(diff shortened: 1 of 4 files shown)


gordo_components/builder/build_model.py
```diff
@@ -467,11 +469,13 @@ def provide_saved_model(
...
-        * cross_val_only: Only perform cross validation
-        * build_only: Skip cross validation and only build the model
-        * full_build: Cross validation and full build of the model, default value
+    evaluation_config: dict
+        Dict of parameters which are exposed to build_model.
+            - cv_mode: str
+                String which enables three different modes:
+                * cross_val_only: Only perform cross validation
+                * build_only: Skip cross validation and only build the model
+                * full_build: Cross validation and full build of the model, default value
```
milesgranger: I think perhaps instead we should refer to the docs in `build_model`as done for the `data_config` and `model_config` parameters.

gordo_components/cli/cli.py
```diff
@@ -188,11 +188,13 @@ def build(
...
-        * cross_val_only: Only perform cross validation
-        * build_only: Skip cross validation and only build the model
-        * full_build: Cross validation and full build of the model, default value
+    evaluation_config: dict
+        Dict of parameters which are exposed to build_model.
+            - cv_mode: str
+                String which enables three different modes:
+                * cross_val_only: Only perform cross validation
+                * build_only: Skip cross validation and only build the model
+                * full_build: Cross validation and full build of the model, default value
```
milesgranger, line 189: Maybe here too, refer to the 'master' doc of this parameter in `build_model` only that it would be in its string representation form?

milesgranger (commented):
Looking promising!

gordo_components/cli/cli.py
EinarJohnsen: I am a bit uncertain about having the 'master' doc in the ```build_model```, as the ```cli``` can be used "without" knowing about the ```build_model```, and for a user to have to "dig" into the code to find it might not be the best. Thoughts? @milesgranger

milesgranger (approved):
Awesome possum! :squid: 

Maybe open a PR into `gordo-test-project` with an example (which works, `full_build` I guess)?

PR #460, opened 2019-09-23
Make ML Server serve multiple models

### LINKED ISSUE #453: RFC: Redesing model server for scale
We want to be able to scale to serve 9000 models. The access patterns will be of two types(?):
1. Continuous: Over 30 min (that is guesswork, but I will use that for the rest of this issue) we want to update all the 9000 models, and we do this at a relatively constant rate of 5 models per second (`5 * 60 * 30 = 9000`).
1. Backfilling: One or more models will be applied on data for e.g. 2 months. Because of batching from the client this might end up just being a few requests (there are roughly 44000 minutes in a month).

# Different models: 
### A single deployment (with possible replicas) per model (today's model). 
Advantage: 
- Fast response time
- Possible to scale individual models which are used much.
- Conceptually simple

Disadvantages: 
- High memory usage. For 1.5GB per model, 9000 models imply 13500 GB, or roughly 220 machines with 64Gb memory. Even if we managed to get LSTM memory usage down to insignificant, or only use simple models, we still need ~400Mb for python, giving roughly 60 machines to keep the models alive. Each model will spend 99% of its time unused. 

### Pack multiple predefined models into each deployment
So we have server-1..server-n, where each server loads e.g. 20 models and keeps them in memory for fast serving.
This gives good memory reduction when we use simple models, so most of the memory usage is python. Gives little memory reduction for e.g. LSTM. 

Disadvantages:
- Small/insignificant benefit as long as we use LSTM/keras/tensorflow witout being able to reduce its memory usage.
- A bit of book-keeping arranging the different deployments and assigning models to them. Especially if we have models of different types, we will have to somehow guess how much memory they need, so if we pack 10 LSTM-models together we must reserve 15Gb memory for it. 

### Single deployment loading models as needed
Here we have a single deployment for the model server per project/project_version. This deployment can be scaled to many many replicas. When it gets a request for model1 it will try to load model1, satisfy the request, and them possibly release model1 from memory. 

If we assume that when we request/limit it to 1 CPU it takes on average 5 seconds to load a model and satisfy the request, then we find that we will need 25 replicas to be able to handle the continuous 9000-model scenario. That is 3.1 8-core nodes. 

For the backfilling case there is a "simple"(?) optimization to cache one/a few of the previous models in memory, and then activate some "sticky routing" in istio (if possible), so it will be default try to serve requests to the same model by using the same actual pod several times. In python there is something called a "weak reference" which lets you keep stuff in memory but allow the GC to clear it if it wants to. The cached models can be kept like that maybe.  This is not useful at all for the continuous case, since after having used a model it is roughly 30 min until you will use that model again.

This model is inspired by "functions as a service" things, except loading the whole docker-image/code on a request it will load only the model. 

The simplest is probably to keep having a single thread, and the request/limit to exactly 1 CPU, and rely on HPA to scale up. That might need some tuning to have the HPA be sensitive enough.

Advantage: 
- Nr of pods scale only with the requests per second, not `nr of models` * `request per second`, so it becomes much cheaper for the 9k case. 
- Pretty easy, we deploy a single deployment per project, and all it needs to know is the folder where it find the models. 

Disadvantage: 
- Each response will be answered much slower than today since it needs to load the model.

## Question:
Whould we go for, or just try out, the "Single deployment loading models as needed" way?

milesgranger:
Will closes #453 

Problem :pill: 
---
Scaling to thousands of models will be quite taxing to have all those loaded models served by individual servers.

Solution/Approach :milk_glass: 
---
Will take the gordo-name as a parameter of the url to load the model by request.
ie. `/gordo/v0/gordo-project/gordo-name/predict` would load the model 'gordo-name' and process the request.

Depends on :cow:
---
- https://github.com/equinor/gordo-infrastructure/pull/389

### DIFF
--- a/gordo_components/server/server.py
+++ b/gordo_components/server/server.py
@@ -1,14 +1,10 @@
 # -*- coding: utf-8 -*-
-import os
 import logging
 import timeit
 import typing
 from functools import wraps
 
 from flask import Flask, g
-from sklearn.base import BaseEstimator
-
-from gordo_components import serializer
 from gordo_components.data_provider.base import GordoBaseDataProvider
 from gordo_components.server import views
 
@@ -19,6 +15,7 @@ class Config:
     """Server config"""
 
     MODEL_LOCATION_ENV_VAR = "MODEL_LOCATION"
+    MODEL_COLLECTION_DIR_ENV_VAR = "MODEL_COLLECTION_DIR"
 
 
 def adapt_proxy_deployment(wsgi_app: typing.Callable) -> typing.Callable:
@@ -97,38 +94,6 @@ def adapt_proxy_deployment(wsgi_app: typing.Callable) -> typing.Callable:
     return wrapper
 
 
-def load_model_and_metadata(
-    model_dir_env_var: str
-) -> typing.Tuple[BaseEstimator, dict]:
-    """
-    Loads a model and metadata from the path found in ``model_dir_env_var``
-    environment variable
-
-    Parameters
-    ----------
-    model_dir_env_var: str
-        The name of the environment variable which stores the location of the model
-
-    Returns
-    -------
-    BaseEstimator, dict
-        Tuple where the 0th element is the model, and the 1st element is the metadata
-        associated with the model
-    """
-    logger.debug("Determining model location...")
-    model_location = os.getenv(model_dir_env_var)
-    if model_location is None:
-        raise ValueError(f'Environment variable "{model_dir_env_var}" not set!')
-    if not os.path.isdir(model_location):
-        raise NotADirectoryError(
-            f'The supplied directory: "{model_location}" does not exist!'
-        )
-
-    model = serializer.load(model_location)
-    metadata = serializer.load_metadata(model_location)
-    return model, metadata
-
-
 def build_app(data_provider: typing.Optional[GordoBaseDataProvider] = None):
     """
     Build app and any associated routes
@@ -156,11 +121,6 @@ def build_app(data_provider: typing.Optional[GordoBaseDataProvider] = None):
         logger.debug(f"Total runtime for request: {runtime_s}s")
         response.headers["Server-Timing"] = f"request_walltime_s;dur={runtime_s}"
         return response
-
-    with app.app_context():
-        app.model, app.metadata = load_model_and_metadata(
-            app.config["MODEL_LOCATION_ENV_VAR"]
-        )
 
     return app
 
--- a/gordo_components/server/server.py
+++ b/gordo_components/server/server.py
@@ -14,7 +14,6 @@ logger = logging.getLogger(__name__)
 class Config:
     """Server config"""
 
-    MODEL_LOCATION_ENV_VAR = "MODEL_LOCATION"
     MODEL_COLLECTION_DIR_ENV_VAR = "MODEL_COLLECTION_DIR"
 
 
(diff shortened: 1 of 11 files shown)


gordo_components/server/server.py
```diff
@@ -19,6 +15,7 @@ class Config:
     """Server config"""
 
     MODEL_LOCATION_ENV_VAR = "MODEL_LOCATION"
+    MODEL_COLLECTION_DIR_ENV_VAR = "MODEL_COLLECTION_DIR"
```
milesgranger, line 17: `MODEL_LOCATION_ENV_VAR` can go.

gordo_components/server/utils.py
```diff
@@ -316,3 +319,45 @@ def wrapper_method(self, *args, **kwargs):
...
+    """
+    path = os.path.join(directory, name)
+    model = serializer.load(path)
+    metadata = serializer.load_metadata(path)
+    return model, metadata
+
+
+def model_required(f):
+    """
+    Decorate a view which has ``gordo_name`` as a url parameter and will
```
milesgranger, line 383: We accept `gordo_project` as well now with the full path.

gordo_components/server/utils.py
```diff
@@ -316,3 +319,70 @@ def wrapper_method(self, *args, **kwargs):
         return method(self, *args, **kwargs)
 
     return wrapper_method
+
+
+@lru_cache(maxsize=int(os.getenv("SERVER_N_MODELS", 2)))
```
epa095: Should the env variable not contain the word "cache"?

gordo_components/server/views/base.py
```diff
@@ -237,6 +236,7 @@ class DownloadModel(Resource):
     @api.doc(
         description="Download model, loadable via gordo_components.serializer.loads"
     )
+    @server_utils.model_required
```
epa095, line 239: Ok for now I guess, but this will cause the server to first load the model, then write it back again to a temp directory, which it again reads using the tar-stuff and sends over the network, instead of just taring the folder-structure it already has?

gordo_components/server/views/base.py
```diff
@@ -215,15 +213,16 @@ class MetaDataView(Resource):
     Serve model / server metadata
     """
 
+    @server_utils.metadata_required
```
epa095, line 216: I must admit that I definitely dont see the prettyness in having these as decorators which puts data inside the magic variable `current_app`, vs just calling a function inside this get.

milesgranger: This is, in effect, exactly how we had it before. Server would load it, but serialize the loaded model upon request. But alas, your wish is my command.

gordo_components/server/views/base.py
milesgranger: This is, in effect, exactly how we had it before. Server would load it, but serialize the loaded model upon request. And this endpoint is called so little, the effectual difference would go unnoticed IMO, but alas, your wish is my command.

milesgranger: I take that back, unless I misunderstand, to 'tar the directly it already has' would be to duplicate the logic we have for what _"serializer.loads"_ requires. If in the future we change `serializer.dumps` to be pickle-able, we'll _also_ have to remember that we need to update the server's `/download` logic. By depending on `serializer.dumps`  we know the two are compatible.

flikka (approved):
Sweet potato, exciting to see how awesome it will work!

PR #461, opened 2019-09-23
Builder: always save model to output_dir

### LINKED ISSUE #447: Model builder: Always deposit model in output-dir (?)
Today the logic of where model-builder deposits the model is a bit odd. 
If a `model-register-dir` is provided then it looks in there to see if it finds a location of a previously built model, and if it does then it does not build the model. It then writes a txt-file to `model-location-file` (default `/tmp/model-location.txt`) with the location of where the model resides. If no `model-register-dir` is provided, or if it does not contain a model, then model-builder *will* save the model to `output-dir`, and write that path in the `model-location-file`.

I propose the following change: 
Always write the model to `output-dir`, possibly *copying* it from the previous location found in `model-register-dir` if provided. No magic `model-location-file`.

Why? 
I think it would make the code nicer.  And the models are tiny, so it should not pose a problem. Also, once when we ran a large gordo, argo workflow failed at copying the path from `model-location-file` to the server, so the server never started. With this change I propose there is no need to communicate the model-location between the builder and the server, the workflow will know beforehand where the model is if the model-builder started.

epa095:
This closes #447

### DIFF
--- a/gordo_components/builder/build_model.py
+++ b/gordo_components/builder/build_model.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import datetime
+from distutils.dir_util import copy_tree
 import hashlib
 import json
 import logging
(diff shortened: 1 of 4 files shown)


milesgranger (approved):
Work of art

PR #462, opened 2019-09-24
Make metadata contextmanager

### LINKED ISSUE #422: Rewrite the way model metadata is fetched
Today the model metadata is fetched after model building by looking deep down in a Pipeline, or even in fields in BaseModels. This code is too complicated, recursive and not very robust to for example Gordo models not at the end of a Pipeline.

There should be a better way, a way that is more similar to logging, where inside the (custom) model class the metadata is written to some thing, and then this is fetched from the metadata thing afterwards.

ryanjdillon:
This PR introduces a new simple context manager for populating the metadata in a cleaner way. 

This will resolve #422 .

### DIFF
--- a/gordo_components/model/metadata.py
+++ b/gordo_components/model/metadata.py
@@ -1,0 +1,185 @@
+from typing import Any, Optional
+
+import json
+import logging
+import threading
+
+logger = logging.getLogger(__name__)
+
+
+class MetadataCollector:
+    """
+    Theaded metadata logging class
+
+    Attributes
+    ----------
+    _tls: theading.local
+        Variable with associated thread.
+    _tls._gordo_metadata: dict
+        Attribute of local variable to store metadata to.
+    metadata: dict
+        A helper reference to `_tls._gordo_metadata` for access under context manager.
+
+    Examples
+    --------
+    >>> def func():
+    ...     MetadataCollector.log_meta("test_key", 42)
+    ...     return
+    >>> with MetadataCollector() as mc:
+    ...     func()
+    ...     print(mc.metadata["test_key"])
+    42
+    """
+
+    _THREAD_LOCAL = threading.local()
+
+    def __enter__(self):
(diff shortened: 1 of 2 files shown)


gordo_components/model/metadata.py
```diff
@@ -0,0 +1,137 @@
...
+    0
+    """
+    mc = MetadataCollector(*args, **kwargs)
+    try:
+        yield mc
+    except Exception as e:
+        raise e
+    finally:
+        # Remove collecter object for Thread cleanliness
+        del mc
```
milesgranger: How much does this actually do? It would be dropped as it goes out of scope/reference count locally, but any client code which holds a reference to it would become the owner, the same way this code is valid:
```python
>>>  a = 'hi'
>>>  b = a
>>> del a
>>> b
'hi'
```
perhaps I misunderstand.

gordo_components/model/metadata.py
```diff
@@ -0,0 +1,137 @@
+from contextlib import contextmanager
+from typing import Any, Optional
+
+import threading
+import warnings
+
+
+class MetadataCollector(object):
```
milesgranger: `(object)` is not necessary.

milesgranger (commented):
Looking pretty good, but seems like `MetadataCollector` would itself be the context manager, no? Somewhat strange to use a function to get the object. IMO, you might implement the `__enter__` and `__exit__` as part of the object.

gordo_components/model/metadata.py
ryanjdillon: :open_mouth:  ... there wasn't a mind blown emoji.

gordo_components/model/metadata.py
ryanjdillon: Implementation is now different, and there is a test now to check that the thread local data is removed.

flikka:
Putting this "on hold", as it seemed hard to reach a consensus.

PR #464, opened 2019-09-25
Change retries in client to have exponential sleep time.

### LINKED ISSUE #411: Exponential retries in client

EinarJohnsen:
Change from the default 5 seconds sleep to exponential sleep based on number of retries. Fixes #411

### DIFF
--- a/tests/gordo_components/client/test_client.py
+++ b/tests/gordo_components/client/test_client.py
@@ -6,6 +6,8 @@ import json
 import logging
 import typing
 from dateutil.parser import isoparse  # type: ignore
+import asyncio
+import time
 
 import aiohttp
 import pytest
@@ -13,6 +15,7 @@ import pandas as pd
 import numpy as np
 from click.testing import CliRunner
 from sklearn.base import BaseEstimator
+from mock import patch, call
 
 from gordo_components.client import Client, utils as client_utils
 from gordo_components.client.utils import EndpointMetadata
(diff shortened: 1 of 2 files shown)


tests/gordo_components/client/test_client.py
```diff
@@ -490,3 +493,29 @@ def test_client_endpoint_filtering(
...
+
+def test_exponential_sleep_time(watchman_service):
+    endpoint = _endpoint_metadata("t1", False)
+
+    start, end = (
+        isoparse("2016-01-01T00:00:00+00:00"),
+        isoparse("2016-01-01T12:00:00+00:00"),
+    )
+
+    with patch(
```
epa095, line 506: nice

tests/gordo_components/client/test_client.py
```diff
@@ -490,3 +493,29 @@ def test_client_endpoint_filtering(
...
+        y = None
+        loop = asyncio.get_event_loop()
+        loop.run_until_complete(
+            client._process_post_prediction_task(
+                X=x, y=y, chunk=slice(0, 1), endpoint=endpoint, start=start, end=end
+            )
+        )
+        loop.close()
+
+    calls = [call(8), call(16), call(32), call(64), call(128)]
```
epa095: maybe call this variable something slightly more expressive, like "expected_calls" or something?

epa095 (commented):
Looking good. But maybe mention the exponential retries in the docstring of the class?

tests/gordo_components/client/test_client.py
```diff
@@ -490,3 +493,32 @@ def test_client_endpoint_filtering(
...
+                chunk=slice(0, 1),
+                endpoint=endpoint,
+                start=start,
+                end=end,
+            )
+        )
+        loop.close()
+
+    expected_calls = [call(8), call(16), call(32), call(64), call(128)]
+    time_sleep.assert_has_calls(expected_calls)
```
milesgranger, line 524: Cool! :mountain_cableway:

milesgranger (approved):
Looks good to me, sir! :bowtie:

PR #469, opened 2019-09-30
Use CNTK for tests, and workflow models

milesgranger:
Problem :timer_clock: 
---
It takes _forever_ to test the `module` component, and now that we are serving models from a single (albeit HPA) server, the LSTMs and Keras models in general, when using tensorflow backend takes  a very long time to respond.

Solution :checkered_flag: 
---
Trial using CNTK

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,9 +4,11 @@ version: 2.1
 executors:
   python-executor:
     working_directory: ~/code
+    environment:
+      KERAS_BACKEND: cntk
     machine:
       docker_layer_caching: true
-      image: "circleci/classic:201808-01"
+      image: "ubuntu-1604:201903-01"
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -153,15 +155,22 @@ jobs:
...
       - run:
-          name: Virtual Environment
+          name: Install CNTK
           command: |
-            python -m venv ~/code/venv
-            echo "source ~/code/venv/bin/activate" >> $BASH_ENV
-      - run: pip install --upgrade pip
+            python --version
+            pip install cntk
+            python -c "import ctnk"
```
milesgranger: cntk

milesgranger:
:crossed_swords: :latin_cross: :boat:

PR #472, opened 2019-10-01
Optionally use parquet format in client and server

milesgranger:
Situation :mushroom: 
---
Serving models from a single server means we can/need to improve efficiency of asking for predictions, especially from our client interactions

Solution :tulip: 
---
Support accepting and returning compressed data via `pyarrow`. 

- The client (and CLI) has a parameter `use_parquet` and `--parquet/--no-parquet`, respectively. Which is set as the default. So the client will compress the dataframe before sending.
- The server accepts either format, still having `X` and `y` and separate files, and can return compressed dataframes with `?format=parquet` URL arg.

### DIFF
--- a/gordo_components/client/client.py
+++ b/gordo_components/client/client.py
@@ -9,11 +9,13 @@ import itertools
 import time
 
 import typing
+from typing import Dict, Any
 from datetime import datetime
 from concurrent.futures import ThreadPoolExecutor, TimeoutError as FutureTimeoutError
 
 import aiohttp
 import pandas as pd
+import pyarrow as pa
 from sklearn.base import BaseEstimator
 from werkzeug.exceptions import BadRequest
 
@@ -49,11 +51,12 @@ class Client:
         metadata: typing.Optional[dict] = None,
         data_provider: typing.Optional[GordoBaseDataProvider] = None,
         prediction_forwarder: typing.Optional[PredictionForwarder] = None,
-        batch_size: int = 1000,
+        batch_size: int = 100000,
         parallelism: int = 10,
         forward_resampled_sensors: bool = False,
         ignore_unhealthy_targets: bool = False,
         n_retries: int = 5,
+        use_pyarrow: bool = False,
     ):
         """
 
@@ -96,6 +99,9 @@ class Client:
         n_retries: int
             Number of times the client should attempt to retry a failed prediction request. Each time the client
             retires the time it sleeps before retrying is exponentially calculated.
+        use_pyarrow: bool
+            Pass the data to the server using the msgpack protocol. Default is True
+            and recommended as it's more efficient for larger batch sizes.
         """
 
         self.base_url = f"{scheme}://{host}:{port}"
@@ -104,6 +110,7 @@ class Client:
         self.endpoints = self._endpoints_from_watchman(self.watchman_endpoint)
         self.prediction_forwarder = prediction_forwarder
         self.data_provider = data_provider
+        self.use_pyarrow = use_pyarrow
 
         # Default, failing back to /prediction on http code 422
         self.prediction_path = "/anomaly/prediction"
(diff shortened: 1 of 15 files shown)


gordo_components/client/client.py
```diff
@@ -104,13 +110,15 @@ def __init__(
...
         self.data_provider = data_provider
+        self.use_pyarrow = use_pyarrow
 
         # Default, failing back to /prediction on http code 422
         self.prediction_path = "/anomaly/prediction"
         self.batch_size = batch_size
         self.parallelism = parallelism
         self.forward_resampled_sensors = forward_resampled_sensors
         self.n_retries = n_retries
+        self.query_params = f"?format={'pyarrow' if use_pyarrow else 'json'}"
```
epa095: `pyarrow` is not really a format, `arrow` is. I think?

gordo_components/client/client.py
```diff
@@ -390,28 +398,38 @@ def predict(
...
-            "y": server_utils.dataframe_to_dict(y.iloc[chunk])
-            if y is not None
-            else None,
-        }
+        kwargs: Dict[str, Any] = dict(
+            session=session,
+            url=f"{endpoint.endpoint}{self.prediction_path}{self.query_params}",
+        )
 
+        # We're going to serialize the data as either JSON or MessagePack
```
epa095: not msg

gordo_components/client/client.py
```diff
@@ -390,28 +398,38 @@ def predict(
...
+        if self.use_pyarrow:
+            kwargs["files"] = {
+                "X": pa.serialize_pandas(X.iloc[chunk]).to_pybytes(),
```
epa095: It is my impression that this does not do any compression, but is pretty much a pure memory dump. Which will still provide smaller messages since it is pure bytes, instead of text. But we can compress by using the pyarrow parquet writer and use its compression (default is "snappy"). An example is [here](https://arrow.apache.org/docs/python/parquet.html#reading-and-writing-single-files). The "file" can be a "[NativeFile](https://arrow.apache.org/docs/python/memory.html#input-and-output)", of which "BufferOutputStream" is one of the implementations, so sounds like getting the bytestream should be pretty easy.

tests/gordo_components/client/test_client.py
```diff
@@ -63,8 +60,9 @@ def test_client_download_model(watchman_service):
 
 @pytest.mark.parametrize("batch_size", (10, 100))
 @pytest.mark.parametrize("use_data_provider", (False, True))
+@pytest.mark.parametrize("use_parquet", (True, False))
 def test_client_predictions_diff_batch_sizes_and_toggle_data_provider(
-    influxdb, watchman_service, use_data_provider: bool, batch_size: int
+    influxdb, watchman_service, use_data_provider: bool, batch_size: int, use_parquet
```
EinarJohnsen: Minor thing, but perhaps have ```use_parquet: bool``` instead

gordo_components/client/client.py
```diff
@@ -552,17 +574,20 @@ def predict(
             return PredictionResult(
                 name=endpoint.target_name, predictions=None, error_messages=[msg]
             )
+        else:
+            logger.info(f"Processing {start} -> {end}")
+            if isinstance(response, dict):
+                predictions = server_utils.dataframe_from_dict(response["data"])
+            else:
+                predictions = server_utils.dataframe_from_parquet_bytes(response)
```
epa095: Note that these lines do the same as the lined 475-480, but slightly different

gordo_components/client/io.py
```diff
@@ -65,24 +65,24 @@ class HttpUnprocessableEntity(BaseException):
...
 
-async def _handle_json(resp: aiohttp.ClientResponse) -> dict:
+async def _handle_response(resp: aiohttp.ClientResponse) -> Union[dict, bytes]:
     if 200 <= resp.status <= 299:
-        return await resp.json()
+        is_json = resp.content_type == "application/json"
+        return await (resp.json() if is_json else resp.content.read())
     else:
         content = await resp.content.read()
         msg = f"Failed to get JSON with status code: {resp.status}: {content}"
```
epa095: A bit of a stupid error message given the new status of this function.

gordo_components/server/utils.py
```diff
@@ -28,6 +32,47 @@
...
+) -> bytes:
+    """
+    Convert a dataframe into bytes representing a parquet table.
+
+    Parameters
+    ----------
+    df: pd.DataFrame
+        DataFrame to be compressed
+    compression: str
+        Compression to use, passed, to  ``pyarrow.parquet.write_table``
```
epa095: ,,,

gordo_components/server/utils.py
```diff
@@ -224,20 +269,33 @@ def wrapper_method(self, *args, **kwargs):
...
             if X is None:
                 message = dict(message='Cannot predict without "X"')
                 return make_response((jsonify(message), 400))
 
             # Convert X and (maybe) y into dataframes.
-            X = dataframe_from_dict(X)
+            if isinstance(X, FileStorage):
+                X = dataframe_from_parquet_bytes(X.read())
+            else:
+                X = dataframe_from_dict(X)
```
epa095: This is exactly the same test as in line 273-278, but tested "opposite", and instead of using "request.json" it tests on the type if X. Maybe write it similar to the previous lines to make it easier to read?

epa095 (commented):
nice, mostly

tests/gordo_components/client/test_client.py
milesgranger: Nice catch, I would have made the same comment. :steam_locomotive:

gordo_components/client/client.py
milesgranger: Moved into a common place.

epa095 (commented):
LGTM

After squashing offc

epa095 (approved):

PR #475, opened 2019-10-02
Remove docs from .dockerignore

epa095:
Before it did not get copied in to the docker builder image, which
caused git inside there to report the wrong version number,
causing the built images to report a bogus version number. Anything
which is tracked by git must be copied in (i.e. not be present in
.dockerignore)


milesgranger (approved):
Nice catch! :male_detective:

PR #477, opened 2019-10-03
Load possible functions defined in definition

milesgranger:
Problem :bomb: 
---
The `serializer.pipeline_into_definition` rightly dumps any function as it's path. ie. `np.mean` becomes the string `numpy.mean`. However `serializer.pipeline_from_definition` only looked at `FunctionTransformer`  specific params as possible functions.

Solution :boom: 
---
Generalize the loading of possible functions when encountering parameters who's value can be loaded as a callable.


epa095 (approved):
Awesome!

PR #492, opened 2019-10-03
Fix client.io post kwarg 'files' -> 'data'

milesgranger:
Problem :cloud_with_lightning: 
---
`requests.post` has parameter `files`, where aiohttp calls the equivalent `data`. aiohttp is what we use in prod, where requests is what we use in our patched tests.

Solution :partly_sunny: 
---
Change the kwarg to `data` and in the patched function, rename that to `files`; albeit, not the prettiest.


epa095 (approved):
Lgtm

PR #493, opened 2019-10-04
WIP: Update all in requirements.txt

epa095:


milesgranger (approved):
Notice there was no updates to `test_requirements.in`, if this was intended, proceed. It does appear that since `test_requirements.txt` contains updates, that perhaps so too, should the .in file.

epa095:
> Notice there was no updates to `test_requirements.in`, if this was intended, proceed. It does appear that since `test_requirements.txt` contains updates, that perhaps so too, should the .in file.

Its on purpose :-)

PR #497, opened 2019-10-04
Add `shift` parameter to TimeSeriesDataset

milesgranger:
New features :cake: 
---

- `TimeSeriesDataset` can _shift_ the target `y` and align that with `X` so we can train models to target future time step values.

### DIFF
--- a/gordo_components/dataset/datasets.py
+++ b/gordo_components/dataset/datasets.py
@@ -28,6 +28,7 @@ class TimeSeriesDataset(GordoBaseDataset):
         row_filter: str = "",
         aggregation_methods: Union[str, List[str], Callable] = "mean",
         row_filter_buffer_size: int = 0,
+        shift: int = 0,
         **_kwargs,
     ):
         """
@@ -72,7 +73,19 @@ class TimeSeriesDataset(GordoBaseDataset):
             Whatever elements are selected for removal based on the ``row_filter``, will also
             have this amount of elements removed fore and aft.
             Default is zero 0
+        shift: int
+            Amount to offset the target column(s).
         _kwargs
+
+        Notes
+        -----
+        In relation to the ``shift`` parameter:
+            Return the offset the ``y`` such that the Nth index of ``y`` relates to the
+            datetime in index ``N - shift`` index of ``X``.
+            Used if you want to train a model to predict future time step values.
+            ie. If the ``resolution`` is 10 minutes and a provided ``shift`` is 1, then
+            the target value(s) for any sample of ``X`` would have the target value(s)
+            10 minutes into the future.
         """
         self.from_ts = from_ts
         self.to_ts = to_ts
@@ -85,6 +98,7 @@ class TimeSeriesDataset(GordoBaseDataset):
         self.row_filter = row_filter
         self.aggregation_methods = aggregation_methods
         self.row_filter_buffer_size = row_filter_buffer_size
+        self.shift = shift
 
         if not self.from_ts.tzinfo or not self.to_ts.tzinfo:
             raise ValueError(
@@ -123,7 +137,21 @@ class TimeSeriesDataset(GordoBaseDataset):
         X = data[x_tag_names]
         y = data[y_tag_names] if self.target_tag_list else None
 
-        return X, y
+        # Apply any shift
+        if self.shift > 0:
+            return X[: -self.shift], y[self.shift :] if y is not None else None
+        elif self.shift < 0:
+            return X[self.shift :], y[: -self.shift] if y is not None else None
+        else:
+            return X, y
+
+    @property
+    def shift(self) -> int:
+        return self._shift
+
+    @shift.setter
+    def shift(self, value: int):
+        self._shift = int(value)
 
     def get_metadata(self):
         metadata = {
@@ -134,6 +162,7 @@ class TimeSeriesDataset(GordoBaseDataset):
             "resolution": self.resolution,
             "filter": self.row_filter,
             "row_filter_buffer_size": self.row_filter_buffer_size,
+            "shift": self.shift,
         }
         return metadata
 
--- a/tests/gordo_components/dataset/test_dataset.py
+++ b/tests/gordo_components/dataset/test_dataset.py
@@ -260,7 +260,13 @@ def test_time_series_no_resolution():
 def test_timeseries_target_tags(tag_list, target_tag_list):
     start = dateutil.parser.isoparse("2017-12-25 06:00:00Z")
     end = dateutil.parser.isoparse("2017-12-29 06:00:00Z")
-    tsd = TimeSeriesDataset(MockDataSource(), start, end, tag_list, target_tag_list)
+    tsd = TimeSeriesDataset(
+        data_provider=MockDataSource(),
+        from_ts=start,
+        to_ts=end,
+        tag_list=tag_list,
+        target_tag_list=target_tag_list,
+    )
     X, y = tsd.get_data()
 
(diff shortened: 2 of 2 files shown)


gordo_components/dataset/datasets.py
```diff
@@ -16,24 +16,23 @@
 logger = logging.getLogger(__name__)
 
 
-class TimeSeriesDataset(GordoBaseDataset):
+class RawTimeSeriesDataset(GordoBaseDataset):
```
epa095: Very minor thing, but can you put the `RawTimeSeriesDataset` below the `TimeSeriesDataset` ? Then maybe the diff will be less dumb

gordo_components/dataset/datasets.py
```diff
@@ -86,42 +72,160 @@ def __init__(
...
+        )
+        return metadata
+
+
+class ShiftedTimeSeriesDataset(TimeSeriesDataset):
+    def __init__(self, shift: int = 1, *args, **kwargs):
+        """
+        Return the same data as given by :class:`TimeSeriesDataset`
+        but offset the ``y`` such that the Nth index of ``y`` relates to the
+        datetime in index N-1 index of ``X``.
```
epa095: N-shift?

gordo_components/dataset/datasets.py
```diff
@@ -86,42 +72,160 @@ def __init__(
...
+        super().__init__(*args, **kwargs)
+
+    @property
+    def shift(self) -> int:
+        return self._shift
+
+    @shift.setter
+    def shift(self, value: int):
+        if value < 0:
+            raise ValueError("Shift must be greater than 0")
```
epa095: Idk, does it in some weird world make sense to not predict, but actually look back instead? So given a state of the sensors try to tell what the previous states were? Should we dissalow that in this dataset? Maybe rather log a warning or something? IDK

milesgranger: Let's go for it! :tada:

gordo_components/dataset/datasets.py
milesgranger: No can do, daddy-o. Then `RawTimeSeriesDataset` is referenced before assignment to the inheritance in `TimeSeriesDataset` :man_shrugging:

gordo_components/dataset/datasets.py
```diff
@@ -86,42 +72,161 @@ def __init__(
...
+        metadata = super().get_metadata()
+        metadata.update({"shift": self.shift})
+        return metadata
+
+    def get_data(self) -> Tuple[pd.DataFrame, Optional[pd.DataFrame]]:
+        X, y = super().get_data()
+        if self.shift > 0:
+            return X[: -self.shift], y[self.shift :] if y is not None else None
+        else:
+            return X[self.shift :], y[: -self.shift] if y is not None else None
```
flikka: Exciting stuff. One question: if y is none, then we just return None, right? Even if X is not none. This makes sense, but perhaps we should log an ERROR and/or throw an exception here, since you have a "shifted" timeseries set, but have not provided a y, which doesn't make sense?

flikka (changes requested):
Me like very much.

Question: Have we thought about how the vast masses of users/clients will use/call these types of models. Or rather, how to interpret the results. Imagine if we could supply timestamps in our Prediction API (if the user supplied it), I think that would make some things a bit less mind boggling. But that's another discussion.

Good work man!

gordo_components/dataset/datasets.py
```diff
@@ -86,42 +72,168 @@ def __init__(
...
+            {
+                "aggregation-methods": self.aggregation_methods,
+                "resolution": self.resolution,
+                "shift": self.shift,
+            }
+        )
+        return metadata
+
+
+class ShiftedTimeSeriesDataset(TimeSeriesDataset):
```
epa095: Why is it still a class?

milesgranger: :man_facepalming:

gordo_components/dataset/datasets.py
```diff
@@ -86,42 +72,150 @@ def __init__(
...
+            Aggregation method(s) to use for the resampled buckets. If a single
+            resample method is provided then the resulting dataframe will have names
+            identical to the names of the series it got in. If several
+            aggregation-methods are provided then the resulting dataframe will
+            have a multi-level column index, with the series-name as the first level,
+            and the aggregation method as the second level.
+            See :py:func::`pandas.core.resample.Resampler#aggregate` for more
+            information on possible aggregation methods.
+        shift: int
+            Amount to offset the target column.
```
epa095: column**s**

gordo_components/dataset/datasets.py
```diff
@@ -86,42 +72,150 @@ def __init__(
...
+class TimeSeriesDataset(RawTimeSeriesDataset):
+    def __init__(
+        self,
+        resolution: str = "10T",
+        aggregation_methods: Union[str, List[str], Callable] = "mean",
+        shift: int = 0,
+        **kwargs,
+    ):
+        """
+        Performs as :class:`RawTimeSeriesDataset`, but with additional resampling functionality.
```
epa095: and shifting

gordo_components/dataset/datasets.py
```diff
@@ -86,42 +72,150 @@ def __init__(
...
-            "train_end_date": self.to_ts,
-            "resolution": self.resolution,
+            "train_start_date": self.from_ts.isoformat(),
+            "train_end_date": self.to_ts.isoformat(),
             "filter": self.row_filter,
         }
         return metadata
 
 
+class TimeSeriesDataset(RawTimeSeriesDataset):
```
epa095: It is a bit subpar that important arguments to TimeSeriesDataset are no longer documented directly in its init, but one has to chase them down in the superclass.  Also, they are required, but that is no longer visible from the signature :-s 
IDK if this is a argument for adding them here as explicit arguments, or for merging the two raw and non-raw timeseriesdatasets

epa095:
I like the concept of being able to shift, but it has spured a bit of a discussion here about how to do it. Should we discuss and think a bit about it before we move on? Maybe split this PR into two, so we can merge the raw dataset?

epa095:
I still think we should wait a bit with this, as it is not completely clear for me that this should be in the dataset.

gordo_components/dataset/datasets.py
```diff
@@ -116,7 +130,21 @@ def get_data(self) -> Tuple[pd.DataFrame, Optional[pd.DataFrame]]:
...
+        else:
+            return X, y
+
+    @property
+    def shift(self) -> int:
+        return self._shift
+
+    @shift.setter
+    def shift(self, value: int):
+        self._shift = int(value)
```
EinarJohnsen, line 154: :cool: stuff

EinarJohnsen (approved):
:jack_o_lantern:  LCTM :ghost:

EinarJohnsen (approved):
:ghost: LGTM :ghost:

flikka (approved):
I also concur in approving this, but @epa095 can try to sell his case for not merging it. I know it might be tricky for a client to use this in a "scaled (thousands of models, maybe with different shifts)" way maybe, but I think it is a step in the right direction.

I assume this shifted option could be used _both_ to imitate the current "Look-ahead" LSTM implementation (which is more a "ignore the most recent data" setting) and a real "prediction" setting (what is the oil production in 2 days) - but the predictionclient for those two scenarios would need to fiddle with the input data differently to make sense of things.

milesgranger:
Doesn't seem to be needed.

PR #498, opened 2019-10-04
Remove GET support from ML server

milesgranger:
Problem :car: 
---
We have a `GET` endpoint for the ML server which has gone unused for some time and has shown no use for continued support.

Solution :truck: 
---
Remove it.

### DIFF
--- a/gordo_components/client/client.py
+++ b/gordo_components/client/client.py
@@ -290,18 +290,10 @@ class Client:
               1st element is the dataframe of the predictions; complete with a DateTime index.
               2nd element is a list of error messages (if any) for running the predictions
         """
-
-        # Determine which method we'll use to get predictions
-        # If we don't have a data_provider instance, we can't source our own data.
-        if self.data_provider is not None:
-            predict_method = self._predict_via_post
-        else:
-            predict_method = self._predict_via_get
-
         # For every endpoint, start making predictions for the time range
         jobs = asyncio.gather(
             *[
-                predict_method(endpoint=endpoint, start=start, end=end)
+                self._predict(endpoint=endpoint, start=start, end=end)
                 for endpoint in self.endpoints
             ]
         )
@@ -317,7 +309,7 @@ class Client:
             (pr.name, pr.predictions, pr.error_messages) for pr in prediction_results
         ]  # type: ignore
 
-    async def _predict_via_post(
+    async def _predict(
         self, endpoint: EndpointMetadata, start: datetime, end: datetime
     ) -> PredictionResult:
         """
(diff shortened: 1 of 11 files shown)


epa095 (commented):
Feels good to delete right? I think quite some code can be deleted from the CLI as well (all that takes inn influx credentials for the server)

milesgranger:
@epa095 guess we can ignore the patch coverage? :sweat_smile:

flikka:
We are sure that Lennart is not planning on using this in his client, right?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

gordo_components/server/views/anomaly.py
```diff
@@ -49,10 +42,6 @@
     "tags": fields.Nested(_tags),
     "total_abnormality": fields.Float,
 }
-API_MODEL_OUTPUT_GET = api.model(
-    "Prediction - Output from GET",
-    {"output": fields.List(fields.Nested(_single_prediction_record))},
-)
 
 
 class AnomalyView(BaseModelView):
```
epa095, line 47: Docstring is wrong now (get)

gordo_components/server/views/anomaly.py
```diff
@@ -93,6 +82,8 @@ class AnomalyView(BaseModelView):
      'time-seconds': '0.1937'}
     """
 
+    methods = ["POST"]
```
epa095, line 83: I cant see where this variable is used. If it is a constant maybe it should BE CAPS!?!?!

gordo_components/client/client.py
```diff
@@ -290,18 +290,10 @@ def predict(
               1st element is the dataframe of the predictions; complete with a DateTime index.
               2nd element is a list of error messages (if any) for running the predictions
         """
-
```
epa095, line 293: Docstring for this method must be updated

gordo_components/server/views/anomaly.py
epa095: Also holds for other methods in the file

epa095 (commented):
Mostly LGTM, but docstrings must be looked over

epa095 (approved):
I am going to prematurely approve this(!) since it is only docstring changes, and I have faith in you

gordo_components/server/views/anomaly.py
milesgranger: Flask uses it internally in their [Views](https://flask.palletsprojects.com/en/1.1.x/api/#flask.views.View.methods) to auto return [405](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405)s

PR #502, opened 2019-10-04
Remove coverage from requirements.txt

epa095:


epa095:
Il put this into #503

PR #510, opened 2019-10-07
Condense LSTMAutoEncoder and LSTMForecast models

### LINKED ISSUE #402: Unify KerasLSTMAutoEncoder and KerasLSTMForecast?
Are KerasLSTMAutoEncoder not just KerasLSTMForecast with lookahead=0? One can argue that it is nice to have `KerasLSTMAutoEncoder` still as a own class to really clarify the intent, and as a simple wrapper. But then we should still maybe refactor it so it actually becomes a simple wrapper, and reuse the code in KerasLSTMForecast instead of having copies of stuff like we do today.

milesgranger:
Will close #402 

Problem :paperclips: 
---
We have a lot of duplicate code/logic between the `KerasLSTMAutoEncoder` and `KerasLSTMForecast` models. Notably the `lookahead` attribute is different.

Solution :paperclip: 
---
Push everything into the `KerasLSTMBase` and set what is needed in the other two models.

### DIFF
--- a/gordo_components/model/models.py
+++ b/gordo_components/model/models.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 
+import abc
 import logging
 import json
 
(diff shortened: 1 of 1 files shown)


flikka (approved):
Holy smoke!

PR #514, opened 2019-10-07
Add favicon to docs

milesgranger:
Problem :x: 
---
We don't have a favicon

Solution :pig: 
---
Now we can!

image


flikka (changes requested):
As much as I LOVE the brains that it must have taken to make this PR, I don't think favicons are supposed to be that large. It might work, but hey, we care for the environment, and don't want people all over the world to waste nuclear energy on downloading that big piece of wonderfulness,  just to resize it? 

Not sure how big makes sense, but maybe not this big.

milesgranger:
@flikka: Always the green thumb :deciduous_tree:

PR #518, opened 2019-10-08
Remove get_model from model.py to avoid unnecessary import

### LINKED ISSUE #512: Clean up import paths to avoid tensorflow
Today tensorflow will end up being loaded almost independently of what part of gordo you want to use. Prominent examples are:
- The timeseriesdataset
- The client
In both of these cases it should not be necessary to load tensorflow, and by cleaning up the import paths it should be doable to avoid it. 

Feel free to split this into subtasks, and also add other "prominent examples"

EinarJohnsen:
`get_model` in model.py is moved to the test utils as it was not used in gordo-components. It made tensorflow load unnecessary. Fixes #512 
Issue is fixed for both timeseriesdataset and client.


epa095 (approved):
LGTM

gordo_components/model/model.py
```diff
@@ -1,18 +0,0 @@
-# -*- coding: utf-8 -*-
-
```
milesgranger, line 2: Did you delete this file?

EinarJohnsen: No, thanks for the reminder, I will remove it.

PR #519, opened 2019-10-08
Make client have Mutex for setting endpoints

### LINKED ISSUE #513: RFC: Client refactoring for repeated small calls to different models
Today the clients downloads all the metadata at init, and filters to whatever machine(s) it is set to use. Then it can only be used for those machines. This is a bit suboptimal if you want to do continuous calls to many (e.g. 9000) different machines, since with todays architecture you would either have to make new clients all the time (resulting in massive metadata-downloads), or you would cache all the 9000 clients. There are probably many ways to solve this, but here are some loose thoughts: 

- Maybe the client should by default download and keep the metadata for all machines at init, so a given client can be used with any machine? Then `predict` needs to take the machine name.
- Maybe the metadata should be handled by a seperate class, which given a project-name, project-version, and machine-name gives the metadata? And this thing can be passed to the client. Then the metadata is not duplicated in all the many client instances? This can be a singleton global object :scream:  ! 

Other thoughts @mrdeveloperdude  @milesgranger  ?

milesgranger:
Will close #513 

Problem :lollipop: 
---
If a situation arises which requires a user of the Client to create many clients; each one will fetch all the endpoints for each new instantiation. 

Solution :tea: 
---
Create a context manager which gives a modified version of `Client` which will make a single call, and new instantiations will reuse the first metadata.

### DIFF
--- a/gordo_components/client/client.py
+++ b/gordo_components/client/client.py
@@ -8,8 +8,7 @@ import logging
 import itertools
 import typing
 from time import sleep
-
-from contextlib import contextmanager
+from threading import Lock
 from typing import Dict, Any
 from datetime import datetime
 from concurrent.futures import ThreadPoolExecutor, TimeoutError as FutureTimeoutError
@@ -39,6 +38,8 @@ class Client:
 
     Enables some basic communication with a deployed Gordo project
     """
+
+    _mutex = Lock()
 
     def __init__(
         self,
@@ -128,33 +129,15 @@ class Client:
             ignore_unhealthy_targets=ignore_unhealthy_targets,
         )
 
-    @classmethod
-    @contextmanager
-    def cached_endpoints_client(cls):
-        """
-        Context manager for client, returns a modified version of :class:`Client`
-        which will only collect metadata from Watchman once. All subsequent instantiations
-        of this modified Client will use the previously loaded metadata/endpoints.
+    @property
+    def endpoints(self) -> typing.List[EndpointMetadata]:
+        with self._mutex:
+            return self._endpoints
 
-        Examples
-        --------
-        >>>
-        >>> with Client.cached_endpoints_client() as CachedClient:  # doctest: +SKIP
-        ...     # Only the first instance will make a call to fetch metadata
-        ...     # subsequent instances will use the cached metadata in this block
-        ...     clients = [CachedClient(...) for i in range(100)]
-        """
-        setattr(cls, "__endpoints", None)
-        yield cls
-        delattr(cls, "__endpoints")
-
-    @classmethod
-    def _cache_endpoints(cls, endpoints: typing.List[EndpointMetadata]):
-        """
-        Caches the endpoints for future possible instances of the Client
-        """
-        logger.debug(f"Setting {len(endpoints)} '__endpoints' for {cls.__name__}")
-        setattr(cls, "__endpoints", endpoints)
+    @endpoints.setter
+    def endpoints(self, endpoints: typing.List[EndpointMetadata]):
+        with self._mutex:
+            self._endpoints = endpoints
 
     @staticmethod
     def _filter_endpoints(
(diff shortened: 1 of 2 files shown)


gordo_components/client/client.py
```diff
@@ -198,36 +227,51 @@ def _endpoints_from_watchman(self, endpoint: str) -> typing.List[EndpointMetadat
...
-                endpoint=f'{self.base_url}{data["endpoint"].rstrip("/")}',
-                tag_list=None,
-                target_tag_list=None,
-                resolution=None,
-                model_offset=None,
-            )
-            for data in resp.json()["endpoints"]
-        ]
+        # Check if we're in a modified Client, and have previously cached the endpoints
+        if hasattr(self, "__endpoints") and getattr(self, "__endpoints") is not None:
```
flikka: I like the concept, but (maybe because I am somewhat n00b in the contextmanager business) I find this test/statement a bit ugly and raw. Not a refined beauty in the world of codelines. But maybe this is the way to do it.

gordo_components/client/client.py
```diff
@@ -198,36 +227,51 @@ def _endpoints_from_watchman(self, endpoint: str) -> typing.List[EndpointMetadat
...
+                    target_tag_list=None,
+                    resolution=None,
+                    model_offset=None,
+                )
+                for data in resp.json()["endpoints"]
+            ]
+
+            # If we're in a modified Client version, we will cache these endpoints
+            # into the class itself.
+            if hasattr(self, "__endpoints"):
```
flikka: Again, maybe (probably) just me not used to the reflection. Seems to be oh so popular.

flikka:
Looks good in my eyes, a part from the fact that the reflection hurts my eyes a bit, but I think it is just me that there is a problem with. Would push approve, but will not, someone else, with more experience with contextmanagers and reflection can do the honors 🤖

tests/gordo_components/client/test_client.py
```diff
@@ -58,6 +59,36 @@ def test_client_download_model(watchman_service):
...
+    # Using Client.call_metadata_once() will create a class attribute `endpoints`
+    # so that all following instantiations of `Client` will not make calls to watchman
+    with mock.patch(
+        "gordo_components.client.client.Client._cache_endpoints",
+        side_effect=Client._cache_endpoints,
+    ) as m:
+        with Client.cached_endpoints_client() as EfficientClient:
+            clients = [
+                EfficientClient(project=tu.GORDO_PROJECT, target=tu.GORDO_SINGLE_TARGET)
+                for i in range(10)
```
EinarJohnsen: Perhaps use ```for _ in range(10)```

tests/gordo_components/client/test_client.py
```diff
@@ -58,6 +59,36 @@ def test_client_download_model(watchman_service):
...
+            ]
+            assert m.call_count == 1
+            assert all(hasattr(client, "endpoints") for client in clients)
+
+        # Now it should not be called, because we will not set `endpoints` as
+        # an attribute of the class but per instance of the class
+        clients = [
+            Client(project=tu.GORDO_PROJECT, target=tu.GORDO_SINGLE_TARGET)
+            for i in range(10)
```
EinarJohnsen: Same as over.

gordo_components/client/client.py
```diff
@@ -127,6 +128,34 @@ def __init__(
...
+        ...     # Only the first instance will make a call to fetch metadata
+        ...     # subsequent instances will use the cached metadata in this block
+        ...     clients = [CachedClient(...) for i in range(100)]
+        """
+        setattr(cls, "__endpoints", None)
+        yield cls
+        delattr(cls, "__endpoints")
+
+    @classmethod
+    def _cache_endpoints(cls, endpoints: typing.List[EndpointMetadata]):
```
epa095: Would it not make sense if "endpoints" was a map from `(host, port, scheme, gordo_version)` to the endpoints? As it is now it will get confused if you try to connect to several projects.

gordo_components/client/client.py
epa095: getattr supports a default, so you can make it just `if getattr(self, "__endpoints", None) is not None`

gordo_components/client/client.py
```diff
@@ -227,51 +210,36 @@ def _endpoints_from_watchman(self, endpoint: str) -> typing.List[EndpointMetadat
...
+            else EndpointMetadata(
+                target_name=None,
+                healthy=data["healthy"],
+                endpoint=f'{self.base_url}{data["endpoint"].rstrip("/")}',
+                tag_list=None,
+                target_tag_list=None,
+                resolution=None,
+                model_offset=None,
+            )
+            for data in resp.json()["endpoints"]
```
ryanjdillon: This seems like it would be more legible without the comprehension.

gordo_components/client/client.py
```diff
@@ -128,33 +129,15 @@ def __init__(
...
 
-    @classmethod
-    @contextmanager
-    def cached_endpoints_client(cls):
-        """
-        Context manager for client, returns a modified version of :class:`Client`
-        which will only collect metadata from Watchman once. All subsequent instantiations
-        of this modified Client will use the previously loaded metadata/endpoints.
+    @property
+    def endpoints(self) -> typing.List[EndpointMetadata]:
```
ryanjdillon: Savvy solution :+1:

ryanjdillon (approved):
LGTM... I just reviewed the new changes as it looks like Erik and Kristian covered the rest.

gordo_components/client/client.py
```diff
@@ -128,6 +129,16 @@ def __init__(
...
         )
 
+    @property
+    def endpoints(self) -> typing.List[EndpointMetadata]:
+        with self._mutex:
+            return self._endpoints
+
+    @endpoints.setter
+    def endpoints(self, endpoints: typing.List[EndpointMetadata]):
+        with self._mutex:
```
epa095: Are you sure this works as intended? I would guess what would happen with 100 threads is that there would be 100 requests to watchman, and then each one of them gets written sequentially to the endpoints variable. But becuase of GIL python, the latter would happen anyway :-p Where the lock is needed is to ensure that only a single instance calls watchman.

epa095: So I would think the lock is only needed around line 125, together with a `if self.endpoint` check.

milesgranger: Spot on, my dear Watson

epa095: I always saw myself more as the Sherlock...

epa095 (approved):
LGTM except the self.endpoint lack

PR #520, opened 2019-10-09
Add pip requirement for dev_requirements

milesgranger:
Problem :ice_cream: 
---
readthedocs doesn't have update pip, so it can't find Tensorflow 2.0

Solution :sun_with_face: 
---
Add it to the `dev_requirements`


epa095 (approved):
:+1:

PR #521, opened 2019-10-09
Add .readthedocs.yml

milesgranger:
As suggested by readthedocs.org

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -15,6 +15,7 @@ python:
   version: 3.6
   install:
     - requirements: dev_requirements.txt
+    - requirements: dev_requirements.in
     - requirements: requirements.txt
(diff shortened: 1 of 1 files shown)


.readthedocs.yml
```diff
@@ -0,0 +1,20 @@
...
+  builder: html
+  configuration: docs/conf.py
+  fail_on_warning: true
+
+formats: all
+
+python:
+  version: 3.6
+  install:
+    - requirements: dev_requirements.txt
```
epa095, line 17: Note that this does *not* contain the pip, since it is regarded as unsafe in a .txt. So maybe add `dev_requirements.in` as well, but after `dev_requirements.txt`?

epa095 (approved):

PR #524, opened 2019-10-09
Fix threading error on server with TensorFlow models

### LINKED ISSUE #523: Server: Tensorflow models fails when multi-threaded(?)
When the server gets several concurrent requests which all utilize tensorflow models, it results in a plethora of exceptions:
```
  File "/usr/local/lib/python3.6/site-packages/tensorflow_core/python/keras/backend.py", line 3773, in function
    return EagerExecutionFunction(inputs, outputs, updates=updates, name=name)
  File "/usr/local/lib/python3.6/site-packages/tensorflow_core/python/keras/backend.py", line 3628, in __init__
    raise ValueError('Cannot create an execution function which is comprised '
ValueError: Cannot create an execution function which is comprised of elements from multiple graphs.
```
or

```
  File "/usr/local/lib/python3.6/site-packages/tensorflow_core/python/eager/function.py", line 511, in call
    ctx=ctx)
  File "/usr/local/lib/python3.6/site-packages/tensorflow_core/python/eager/execute.py", line 67, in quick_execute
    six.raise_from(core._status_to_exception(e.code, message), None)
  File "<string>", line 3, in raise_from
tensorflow.python.framework.errors_impl.InvalidArgumentError:  You must feed a value for placeholder tensor 'dense_24_3/BiasAdd/ReadVariableOp/resource' with dtype resource
         [[node dense_24_3/BiasAdd/ReadVariableOp/resource (defined at /site-packages/tensorflow_core/python/framework/ops.py:1751) ]] [Op:__inference_keras_scratch_graph_1534]

Function call stack:
keras_scratch_graph
```

Solution alternatives:
- Have a lock on the predict endpoint to ensure a single thread using it at the same time
- Set to threaded=False in flask. Maybe also move to gunicorn with async worker model to handle concurrent requests while still being singlethreaded.

ryanjdillon:
This PR will resolve #523 .

### DIFF
--- a/gordo_components/server/server.py
+++ b/gordo_components/server/server.py
@@ -125,7 +125,7 @@ def build_app():
 
 def run_server(host: str = "0.0.0.0", port: int = 5555, debug: bool = False):
     app = build_app()
-    app.run(host, port, debug=debug)
+    app.run(host, port, debug=debug, threaded=False)
 
 
 if __name__ == "__main__":


ryanjdillon:
This is a quick fix, but the plan forward is to rather use Gunicorn instead of Flask's development `app.run()` webserver.

epa095 (approved):

PR #534, opened 2019-10-14
Move workflow generator and related parts

milesgranger:
Problem :writing_hand: 
---
Gordo components isn't fully complete without the workflow generator and related parts of the 'workflow' process.

Solution :keyboard: 
---
This creates a `workflow` component which includes these pieces. Only some minor improvements have been made to make this inline with gordo-components. ie. Not using `argparse` but `click`, and therefore using `click.testing.CliRunner` instead of the ugly capturing of stdout use in infrastructure

Some reorganization may be required later, or in this PR?


Dockerfile-argo
```diff
@@ -0,0 +1,51 @@
+# Installs workflow-generator, kubectl and argo. Runs workflow-generator on the machine-config
```
epa095: maybe use the opportunity to rename this to follow the same scheme as the other dockerfiles?

pytest.ini
```diff
@@ -7,4 +7,4 @@ flakes-ignore =
 filterwarnings =
     ignore::DeprecationWarning
     ignore::PendingDeprecationWarning
-timeout = 100
+timeout = 300
```
epa095, line 10: On purpose ?

milesgranger: You betcha. The linting of the generated workflow can take quite some time occasionally. Was at 200 for part of the PR, but think it was just barely over that.

epa095 (approved):
LGTM

PR #536, opened 2019-10-15
Update gordo server to run with Gunicorn

ryanjdillon:
This PR does the following:
* Updates `server.server::run_server` to run the Flask server app with Gunicorn and subprocess
   - this includes instantiating the app with import, and running the module runs the flask `app.run()` method 
* Update cli `run-server` command a tad
* Update run server cli test to check input validation
* Update Dockerfile to call cli with no args (i.e. defaults). 

This could be done differently by adding some logic in `server.run_server()` and configuring with an env var or something, but I didn't like that approach.

### DIFF
--- a/gordo_components/cli/cli.py
+++ b/gordo_components/cli/cli.py
@@ -5,9 +5,7 @@ CLI interfaces
 """
 
 import logging
-import warnings
-import traceback
-import os
+from gunicorn.glogging import Logger
 
 import jinja2
 import yaml
@@ -28,45 +26,12 @@ from gordo_components.server import server
 from gordo_components import watchman, __version__
 from gordo_components.cli.workflow_generator import workflow_cli
 from gordo_components.cli.client import client as gordo_client
-from gordo_components.cli.custom_types import key_value_par, DataProviderParam
+from gordo_components.cli.custom_types import key_value_par, DataProviderParam, HostIP
 from gordo_components.dataset.sensor_tag import normalize_sensor_tags
 
 import dateutil.parser
 
-try:
-    # FIXME(https://github.com/abseil/abseil-py/issues/99)
-    # FIXME(https://github.com/abseil/abseil-py/issues/102)
-    # Unfortunately, many libraries that include absl (including Tensorflow)
-    # will get bitten by double-logging due to absl's incorrect use of
-    # the python logging library:
-    #   2019-07-19 23:47:38,829 my_logger   779 : test
-    #   I0719 23:47:38.829330 139904865122112 foo.py:63] test
-    #   2019-07-19 23:47:38,829 my_logger   779 : test
-    #   I0719 23:47:38.829469 139904865122112 foo.py:63] test
-    # The code below fixes this double-logging.  FMI see:
-    #   https://github.com/tensorflow/tensorflow/issues/26691#issuecomment-500369493
-
-    import absl.logging
-
-    logging.root.removeHandler(absl.logging._absl_handler)
-    absl.logging._warn_preinit_stderr = False
-
-except Exception:
-    warnings.warn(f"Failed to fix absl logging bug {traceback.format_exc()}")
-    pass
-
-
-# Set log level, defaulting to DEBUG
-log_level = os.getenv("LOG_LEVEL", "DEBUG").upper()
-azure_log_level = os.getenv("AZURE_DATALAKE_LOG_LEVEL", "INFO").upper()
-
-logging.basicConfig(
-    level=getattr(logging, log_level),
-    format="[%(asctime)s] %(levelname)s [%(name)s.%(funcName)s:%(lineno)d] %(message)s",
-)
-
 logger = logging.getLogger(__name__)
-logging.getLogger("azure.datalake").setLevel(azure_log_level)
 
 
 @click.group("gordo-components")
@@ -387,10 +352,51 @@ def get_all_score_strings(metadata):
 
 
 @click.command("run-server")
-@click.option("--host", type=str, help="The host to run the server on.")
-@click.option("--port", type=int, help="The port to run the server on.")
-def run_server_cli(host: str, port: int):
-    server.run_server(host, port)
+@click.option(
+    "--host",
+    type=HostIP(),
+    help="The host to run the server on.",
+    default="0.0.0.0",
+    envvar="GORDO_SERVER_HOST",
+    show_default=True,
+)
+@click.option(
+    "--port",
+    type=click.IntRange(1, 65535),
+    help="The port to run the server on.",
+    default=5555,
+    envvar="GORDO_SERVER_PORT",
+    show_default=True,
+)
+@click.option(
+    "--workers",
+    type=click.IntRange(1, 4),
+    help="The number of worker processes for handling requests.",
+    default=2,
+    envvar="GORDO_SERVER_WORKERS",
+    show_default=True,
+)
+@click.option(
+    "--worker-connections",
+    type=click.IntRange(1, 4000),
+    help="The maximum number of simultaneous clients per worker process.",
+    default=50,
+    envvar="GORDO_SERVER_WORKER_CONNECTIONS",
+    show_default=True,
+)
+@click.option(
+    "--log-level",
+    type=click.Choice(Logger.LOG_LEVELS.keys()),
+    help="The log level for the server.",
+    default="debug",
+    envvar="GORDO_SERVER_LOG_LEVEL",
+    show_default=True,
+)
+def run_server_cli(host, port, workers, worker_connections, log_level):
+    """
+    Run the gordo server app with Gunicorn
+    """
+    server.run_server(host, port, workers, worker_connections, log_level.lower())
 
 
 @click.command("run-watchman")
--- a/gordo_components/cli/custom_types.py
+++ b/gordo_components/cli/custom_types.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 
+import ipaddress
 import os
 import typing
 
@@ -47,6 +48,21 @@ class IsoFormatDateTime(click.ParamType):
             return parser.isoparse(value)
         except ValueError:
             self.fail(f"Failed to parse date '{value}' as ISO formatted date'")
+
+
+class HostIP(click.ParamType):
+    """
+    Validate input is a valid IP address
+    """
+
+    name = "host"
+
+    def convert(self, value, param, ctx):
+        try:
+            ipaddress.ip_address(value)
+            return value
+        except ValueError as e:
+            self.fail(e)
 
 
 def key_value_par(val) -> typing.Tuple[str, str]:
--- a/gordo_components/server/server.py
+++ b/gordo_components/server/server.py
@@ -1,8 +1,16 @@
 # -*- coding: utf-8 -*-
+"""
+This module contains code for generating the Gordo server Flask application.
+
+Running this module will run the application using Flask's development webserver.
+Gunicorn can be used to run the application as `gevent` async workers by using the
+:func:`~gordo_components.server.server.run_server` function.
+"""
 import logging
 import timeit
 import typing
 from functools import wraps
+import subprocess
 
 from flask import Flask, g
 from gordo_components.server import views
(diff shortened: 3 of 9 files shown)


ryanjdillon:
One thing I've not yet done is set the number of workers. [Gunicorn docs](http://docs.gunicorn.org/en/stable/design.html#how-many-workers) state that 4-12 workers can handle 100's of 1000's of requests. 

Setting these dynamically in the build_aioapp or cli methods seems wrong, as that would look at the number of cores on the workflow instance and not the spawned server instance. 

Would maybe make most sense to take the config of the server from the workflow config and calc the workers from there or something.

ryanjdillon:
Killer overview on Gunicorn worker type breakdown: [link](https://medium.com/building-the-system/gunicorn-3-means-of-concurrency-efbb547674b7)

:rage1:

gordo_components/server/server.py
```diff
@@ -123,10 +133,77 @@ def base_healthcheck():
...
+        The log level for the server.
+    """
 
+    try:
+        ipaddress.ip_address(host)
+    except Exception:
+        raise
+
+    if not (1 <= port <= 65535):
+        raise ValueError(f"{port} is not a valid port number")
```
milesgranger: Flask already does this check.

gordo_components/server/server.py
```diff
@@ -123,10 +133,77 @@ def base_healthcheck():
...
+    worker_connections: int
+        The maximum number of simultaneous clients per worker process.
+    log_level: str
+        The log level for the server.
+    """
 
+    try:
+        ipaddress.ip_address(host)
+    except Exception:
+        raise
```
milesgranger: Why bother with the `try`?

gordo_components/server/server.py
```diff
@@ -123,10 +133,77 @@ def base_healthcheck():
...
+
+    if not (1 <= port <= 65535):
+        raise ValueError(f"{port} is not a valid port number")
+
+    gunicorn_log_levels = Logger.LOG_LEVELS.keys()
+    if log_level not in gunicorn_log_levels:
+        raise ValueError(
+            f"'{log_level}' log level is not support by Gunicorn. Must be one of {', '.join(gunicorn_log_levels)}."
+        )
```
milesgranger: I propose you instead leave the validation to `click`

ie.
```python
@click.option('--log-level', type=click.Choice(Logger.LOG_LEVELS.keys(), case_sensitive=False))
```

gordo_components/server/server.py
```diff
@@ -123,10 +133,77 @@ def base_healthcheck():
...
+    cmd = [
+        "gunicorn",
+        "--bind",
+        f"{host}:{port}",
+        "--worker-class",
+        "gevent",
+        "--access-logfile",
+        "-",
+        "--error-logfile",
+        "-",
```
milesgranger: Not worth changing, but I suppose `-` for the log files indicates outputting to stdout? Is this not the default behavior?

gordo_components/server/server.py
```diff
@@ -123,10 +133,77 @@ def base_healthcheck():
...
+    if log_level not in gunicorn_log_levels:
+        raise ValueError(
+            f"'{log_level}' log level is not support by Gunicorn. Must be one of {', '.join(gunicorn_log_levels)}."
+        )
+
+    if workers <= 0:
+        raise ValueError("Workers must be greater than 0, not {workers}.")
+
+    if worker_connections <= 0:
+        raise ValueError("Workers must be greater than 0, not {workers}.")
```
milesgranger: `Worker connections must be...{worker_connections}`

gordo_components/server/server.py
```diff
@@ -123,10 +133,77 @@ def base_healthcheck():
...
+    host: str
+        The host to run the server on.
+    port: int
+        The port to run the server on.
+    workers: int
+        The number of worker processes for handling requests.
+    worker_connections: int
+        The maximum number of simultaneous clients per worker process.
+    log_level: str
+        The log level for the server.
```
milesgranger: If these are different than the valid ones for `logger` module, then perhaps we should mention what are the valid values, or where to find them?

milesgranger (commented):
Looking pretty goody!

gordo_components/server/server.py
ryanjdillon: Brilliant!

gordo_components/server/server.py
ryanjdillon: Awesome :+1:

ryanjdillon:
As a minor brain dump before the weekend... it seems like requests are taking a long time, and not clear at the moment what may be causing that.

This article suggests is could be the "heatbeat" getting blocked, and possibly the "work temp file" which may only apply to `sync` workers.
https://pythonspeed.com/articles/gunicorn-in-docker/

This dude says that got bad performance with kubernetes and gunicorn on AWS... said uwsgi performed a lot better for them.
https://stackoverflow.com/a/51855285/943773

Here is a page outlining setting things up with `uwgsi` and `gevent`:
https://uwsgi-docs.readthedocs.io/en/latest/Gevent.html

gordo_components/cli/cli.py
```diff
@@ -386,11 +352,70 @@ def get_all_score_strings(metadata):
     return all_scores
 
 
+def validate_ip(ctx, parameter, value):
```
milesgranger: Maybe better to follow what we've been doing with custom click params in cli.custom_types Whether you want to create a object out of it, or at least move these function into that.

gordo_components/cli/cli.py
```diff
@@ -386,11 +352,70 @@ def get_all_score_strings(metadata):
...
+    Validated click input is a valid IP address
+    """
+    try:
+        ipaddress.ip_address(value)
+    except ValueError as e:
+        raise click.BadParameter(e)
+    return value
+
+
+def validate_above_zero(ctx, parameter, value):
```
milesgranger: Same as `validate_ip` comment

gordo_components/server/server.py
```diff
@@ -123,10 +131,56 @@ def base_healthcheck():
...
+        The log level for the `gunicorn` webserver. Valid log level names can be found
+        in the [gunicorn documentation](http://docs.gunicorn.org/en/stable/settings.html#loglevel).
+    """
 
+    cmd = [
+        "gunicorn",
+        "--bind",
+        f"{host}:{port}",
+        "--log-level",
+        log_level,
```
milesgranger, line 175: Just a note here, you've set `case_sensitive=False` I assume that is true and you won't need to set `.lower()` or anything here?

gordo_components/server/server.py
```diff
@@ -123,10 +131,56 @@ def base_healthcheck():
...
+        subprocess.check_call(cmd)
+    except subprocess.CalledProcessError as e:
+        logger.exception(e)
+        raise
+    except OSError as e:
+        logger.exception(f"{e}: Check that Gunicorn is installed.")
+        raise
+
+
+app = build_app()
```
milesgranger, line 189: Should this not be in the `__main__` block?

gordo_components/server/server.py
```diff
@@ -1,8 +1,16 @@
 # -*- coding: utf-8 -*-
+"""
+This module contains code for generating the Gordo server Flask application.
+
+Running this module will run the application using Flask's development webserver.
+Gunicorn can be used to run the application as AsyncIO workers, having build the Flask
+app with :func:`~gordo_components.server.build_aioapp`.
```
milesgranger: Is this a real location? `build_aioapp`?

gordo_components/server/server.py
ryanjdillon: This was for referencing the `app` in the gunicorn call:
`"gordo_components.server.server:app",`

I'd initially created a new module for doing, but then decided this was perhaps simpler. Maybe that is better?

**production.py**
```
from gordo_components.server.server import build_app

app = build_app()
```
**server.py**
```
def run_server():
   cmd = ["gunicorn",  "gordo_components.server.production:app"]
```

gordo_components/cli/cli.py
ryanjdillon: Good to know :ok_hand:

gordo_components/server/server.py
ryanjdillon: whoops... nope, not any more. Thanks!

gordo_components/server/server.py
ryanjdillon: Does that seem OK to you, or should I do it a different way you think?

gordo_components/server/server.py
ryanjdillon: Removing these here unnecessary args.

gordo_components/cli/client.py
```diff
@@ -28,7 +28,7 @@
 @click.option("--port", help="Port the server is running on", default=443)
 @click.option("--scheme", help="tcp/http/https", default="https")
 @click.option("--gordo-version", help="Version of gordo", default="v0")
-@click.option("--batch-size", help="How many samples to send", default=100000)
```
milesgranger: Is this what we _really_ want?! :fearful:

gordo_components/cli/custom_types.py
```diff
@@ -49,6 +50,43 @@ def convert(self, value, param, ctx):
...
+                self.fail(
+                    f"{param} must be greater than zero, got {value!r}.", param, ctx
+                )
+            return value
+        except ValueError:
+            self.fail(
+                f"{value!r} of type {type(value).__name__} is not a valid integer.",
+                param,
+                ctx,
+            )
```
milesgranger: Nice! :+1:

ryanjdillon: :blush:

gordo_components/cli/client.py
ryanjdillon: Whoops... that was just for my WIP testing, things are looking good now so I'll make that get outa here.

gordo_components/cli/cli.py
```diff
@@ -387,10 +357,46 @@ def get_all_score_strings(metadata):
...
+@click.option(
+    "--workers",
+    type=PositiveInteger(),
+    help="The number of worker processes for handling requests.",
+    default=2,
+    envvar="GORDO_SERVER_WORKERS",
+)
+@click.option(
+    "--worker-connections",
+    type=PositiveInteger(),
```
milesgranger: Whats the difference between these and `click.IntRange(1, some_sensible_limit)`?

gordo_components/server/server.py
```diff
@@ -123,10 +131,55 @@ def base_healthcheck():
     return app
 
 
-def run_server(host: str = "0.0.0.0", port: int = 5555, debug: bool = False):
-    app = build_app()
-    app.run(host, port, debug=debug, threaded=False)
+def run_cmd(cmd):
```
milesgranger, line 134: This function is called once, correct? Is it worth it to have an entirely different function for one line? and what I see it doesn't handle `CalledProcessError` and/or `OSError` like the doc says; it just propagates it out like it would anyway?

tests/gordo_components/cli/test_cli.py
```diff
@@ -275,8 +274,8 @@ def test_build_cv_mode_cross_val_cache(
     tmp_dir: tempfile.TemporaryDirectory,
 ):
     """
-    Checks that cv_scores uses cache if runned after a full build. Loads the same model, and can 
-    print the cv_scores from them. 
+    Checks that cv_scores uses cache if runned after a full build. Loads the same model, and can
```
milesgranger: `if runned`...

tests/gordo_components/cli/test_cli.py
```diff
@@ -333,24 +332,37 @@ def test_build_cv_mode_build_only(tmp_dir: tempfile.TemporaryDirectory):
...
-            args.extend(["--host", host])
-        if port is not None:
-            args.extend(["--port", str(port)])
-        result = runner.invoke(cli.gordo, args)
-
-        assert result.exit_code == 0
-        m.assert_called_once_with(host, port)
+    args = ["run-server", arg, value]
+
+    monkeypatch.setattr(
```
milesgranger: do you think this should have something like `patched.assert_called_with(args)` or something? Not a big deal I suppose, but won't catch the fail if someone just deletes `subprocess.check_call` all together.

gordo_components/cli/cli.py
ryanjdillon: True, true... :) I'll put `workers` at 5 and `worker_connections` at 4000, which should contain things well enough for our application here. :+1:

gordo_components/server/server.py
ryanjdillon: I ended up doing that to make the tests straight forward for checking error cases. I need to update the doc. The way I use `check_call` now correctly handles CalledProcessErrors and OSErrors, but without extra handling on our side.

tests/gordo_components/cli/test_cli.py
ryanjdillon: Hmmm... good point. I'll use `assert_called_with` then.

epa095 (commented):
LGTM. Il let @milesgranger handle the final approve since he has comments

milesgranger (approved):

PR #539, opened 2019-10-15
Support no resampling in TimeSeriesDataset

milesgranger:
Access the timeseries data without any forced resampling/aggregations

Moving over from #497

### DIFF
--- a/gordo_components/dataset/datasets.py
+++ b/gordo_components/dataset/datasets.py
@@ -54,7 +54,7 @@ class TimeSeriesDataset(GordoBaseDataset):
         resolution: Optional[str]
             The bucket size for grouping all incoming time data (e.g. "10T").
             Available strings come from https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#dateoffset-objects
-            **Note**: If this parameter is None, then _no_ aggregation/resampling is applied to the data.
+            **Note**: If this parameter is ``None`` or ``False``, then _no_ aggregation/resampling is applied to the data.
         row_filter: str
             Filter on the rows. Only rows satisfying the filter will be in the dataset.
             See :func:`gordo_components.dataset.filter_rows.pandas_filter_rows` for
@@ -101,9 +101,7 @@ class TimeSeriesDataset(GordoBaseDataset):
         )
 
         # Resample if we have a resolution set, otherwise simply join the series.
-        if self.resolution is None:
-            data = pd.concat(series_iter, axis=1, join="inner")
-        else:
+        if self.resolution:
             data = self.join_timeseries(
                 series_iter,
                 self.from_ts,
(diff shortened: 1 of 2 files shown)


gordo_components/dataset/datasets.py
```diff
@@ -86,42 +72,116 @@ def __init__(
...
+        """
+        Performs as :class:`RawTimeSeriesDataset`, but with additional
+        resampling functionality
+
+        A TimeSeriesDataset is a dataset backed by RawTimeSeriesDataset, but resampled.
+        Time series are still aligned, and (optionally) filtered as with
+        :class:`gordo_components.dataset.datasets.RawTimeSeriesDataset`
+
+        Parameters
+        ----------
```
epa095: It is a bit subpar that important arguments to TimeSeriesDataset are no longer documented directly in its init, but one has to chase them down in the superclass. Also, they are required, but that is no longer visible from the signature :-s
IDK if this is a argument for adding them here as explicit arguments, or for merging the two raw and non-raw timeseriesdatasets

tests/gordo_components/dataset/test_dataset.py
```diff
@@ -173,21 +178,25 @@ def test_row_filter(self):
...
         X, _ = TimeSeriesDataset(
-            MockDataSource(),
-            start,
-            end,
+            data_provider=MockDataSource(),
+            from_ts=start,
+            to_ts=end,
             tag_list=tag_list,
             row_filter="'Tag 1' / 'Tag 3' < 0.999",
         ).get_data()
```
EinarJohnsen: Not saying it should be done in this PR, but we should might think about using pytest to parameterize this instead, it looks like unnecessary duplication.

gordo_components/dataset/datasets.py
```diff
@@ -98,13 +99,18 @@ def get_data(self) -> Tuple[pd.DataFrame, Optional[pd.DataFrame]]:
...
-        data: pd.DataFrame = self.join_timeseries(
-            series_iter,
-            self.from_ts,
-            self.to_ts,
-            self.resolution,
-            aggregation_methods=self.aggregation_methods,
-        )
+
+        # Resample if we have a resolution set, otherwise simply join the series.
+        if self.resolution is None:
```
epa095: Is it not more pythonic to do `if self.resolution`...`else`? Then empty string is also regarded as Falsy, just as `False` and `None`

tests/gordo_components/dataset/test_dataset.py
```diff
@@ -238,6 +225,22 @@ def test_aggregation_methods(self):
...
+    kwargs = dict(
+        data_provider=MockDataSource(),
+        tag_list=[
+            SensorTag("Tag 1", None),
+            SensorTag("Tag 2", None),
+            SensorTag("Tag 3", None),
+        ],
+        from_ts=dateutil.parser.isoparse("2017-12-25 06:00:00Z"),
+        to_ts=dateutil.parser.isoparse("2017-12-29 06:00:00Z"),
+    )
```
ryanjdillon, line 238: You could refactor this out and append test specific keys as needed.

ryanjdillon (approved):
LGTM :+1:

epa095 (approved):
LGTM

milesgranger:
@choukha This is what you were asking about a while ago, access to raw data without any resampling. Will be out in the next release. Or use master of course. :+1:

PR #545, opened 2019-10-17
Change client CLI batch-size to match Client's init default

### LINKED ISSUE #538: Client: Change default batch size
It has been changed other places, but not in the cli.
https://github.com/equinor/gordo-components/blob/8eb8bfbe66d0267a5012d82655909c88e0b1dea0/gordo_components/cli/client.py#L31

milesgranger:
Will close #538 

Problem :rotating_light: 
---
The `Client`'s init has a default `batch_size` of 100k, while the CLI entry default for `batch-size` is still at 1k. 

Solution :bulb: 
---
Make them match, so less calls are done between the client and server.

### DIFF
--- a/gordo_components/cli/client.py
+++ b/gordo_components/cli/client.py
@@ -28,7 +28,7 @@ from gordo_components.client.forwarders import ForwardPredictionsIntoInflux
 @click.option("--port", help="Port the server is running on", default=443)
 @click.option("--scheme", help="tcp/http/https", default="https")
 @click.option("--gordo-version", help="Version of gordo", default="v0")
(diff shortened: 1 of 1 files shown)


epa095 (commented):
This means that we have not actually tested the large batches though maybe, or did we test by explicitly setting in the workflow? I think maybe we did, but I cant remember. I propose we get the workflow inn, test and stabilize, and then pull this one inn after.

milesgranger:
I can attest that we _did_ the first time when we ran it by setting the `batch-size` explicitly in the workflow template. :+1: But trying it again seems like a prudent choice; with all these changes going on.

epa095 (approved):
LGTM

PR #546, opened 2019-10-17
Fix CLI connection to watchman-to-sql command

milesgranger:
Problem :moon: 
---
During the initial move of workflow generator into components #534 , the `watchman-to-sql` CLI command wasn't not hooked up to the `gordo-components` CLI entry point and not updated in the workflow template.

Solution :sun_behind_large_cloud: 
---
Hook it up under `gordo-components workflow watchman-to-sql` and update the workflow template to reflect this.


epa095 (approved):
LBTM

PR #548, opened 2019-10-17
Allow parametrization of image sources in workflow

milesgranger:
Problem :credit_card: 
---
It is currently not possible to parametrize where the images are pulled from. ie. If someone (us) makes images available on, say, Dockerhub, we can't choose to use them if deploying on something like minikube or somewhere which we cannot access our azure images.

Solution :moneybag: 
---
Add ability in workflow generator CLI to specify the image registry and repository to pull from.


flikka (approved):
It _looks_ fine fine to me, you probably also tested it, so hoy hoy.

PR #553, opened 2019-10-18
Add a base_path argument to NCS reader

flikka:
Instead of hard-coding demo assets or other cases where you want to decide the base path (testing, demos) - provide a parameter for it.


epa095 (changes requested):
I propose we rather add an optional parameter to the data lake data reader `base_path` which, if provided, is used as the base-path, ignoring assets completly

flikka:
> I propose we rather add an optional parameter to the data lake data reader `base_path` which, if provided, is used as the base-path, ignoring assets completly

Yepp.

epa095 (approved):
You tested it? It looks pretty good to me.

Edit: except the way-to-long commit message of course!!!

epa095:
Should we merge this in before a release @flikka  ?

PR #556, opened 2019-10-21
Remove CLI build modifications to passed configs.

milesgranger:
Problem :face_with_head_bandage: 
---
Attempting to run a raw config file from a notebook presents a number of problems.

1. Keys to `dataset` are renamed internally.
2. The renaming doesn't happen where you might expect, it happens in the CLI
3. `data_provider` is expected as another argument to the CLI build which is then injected(!) into the data config.
4. The tags are parsed externally from the dataset they're being handed to.
5. The datetime strings are also parsed externally between the config and dataset init.

All this requires a significant manual, and error prone, preprocessing to run a raw config locally to build a model.

Solution :cowboy_hat_face: 
---
By correcting these, we can do a 'local_build' function that is parsed by NormalizedConfig and make it such that this will _just work_, without having to do these same modifications in a notebook/terminal
```python
config = """
      dataset:
          tags:
            - SOME-TAG1
            - SOME-TAG2
          target_tag_list:
            - SOME-TAG3
            - SOME-TAG4
          train_end_date: '2019-03-01T00:00:00+00:00'
          train_start_date: '2019-01-01T00:00:00+00:00'
      metadata:
         information: 'Use basic LinearRegression model with multiple target features with DiffBasedAnomalyDetector'
      model:
          gordo_components.model.anomaly.diff.DiffBasedAnomalyDetector:
            base_estimator:
              sklearn.pipeline.Pipeline:
                steps:
                - sklearn.decomposition.pca.PCA
                - sklearn.multioutput.MultiOutputRegressor:
                    estimator: sklearn.linear_model.base.LinearRegression
      name: trolla-supervised-lin-reg-multi
"""
from gordo_components.builder import local_build
local_build(config)
```

Extended notes
---
Much of the raw config preprocessing occurs in the CLI, outside
of the context of builder.build_model. Things like renaming "train_start_date"
inside of this CLI makes it quite confusing and opaque to the user of how
the raw configs are translated into the object used to get the data
and build the models.

This removes, or at least condenses, this logic into the objects themselves,
rather than doing the preprocessing between the time the yaml is submitted
and when the objects are initialized from those configs.


gordo_components/builder/local_build.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def local_build(config_str: str) -> Iterable[Tuple[Union[BaseEstimator, None], dict]]:
+    """
+    Build model(s) from a bare Gordo config file locally.
+
+    This uses the much same process the the normal workflow generation and subsequent
+    deployment the normal Gordo deployment process makes. Should help developing locally
+    as well as giving a good indication that your config is valid for deployment
+    with Gordo.
```
EinarJohnsen: I assume you mean something in the ballpark of: "This uses much of the same process as the nornaml.."?

milesgranger: me english best. Talk everyone at great me english knowledge.

gordo_components/dataset/datasets.py
```diff
@@ -3,10 +3,16 @@
 
 from typing import Tuple, List, Dict, Union, Optional, Iterable, Callable
 from datetime import datetime
+from dateutil.parser import isoparse
+
```
EinarJohnsen: Nice catch :space_invader:

tests/gordo_components/dataset/test_dataset.py
```diff
@@ -298,3 +307,20 @@ def load_series(
                 index=days, data=list(range(i, len(days) + i)), name=name
             )
             yield series
+
+
+def test__timeseries_dataset_compat():
```
EinarJohnsen: I like underscores, but perhaps one is enough?

examples/Gordo-Workflow-High-Level.ipynb
```diff
@@ -41,90 +41,57 @@
...
    "metadata": {},
    "outputs": [],
    "source": [
-    "pipe, metadata = build_model(\n",
-    "    name=config['model'],\n",
-    "    model_config=model_config, \n",
-    "    data_config=data_config,\n",
-    "    metadata=machine_config['metadata']\n",
-    ")"
+    "pipe, metadata = list(local_build(config))[0]"
```
ryanjdillon: Perhaps it is clearer what `local_build` is returning by using `next(local_build(config))`?

ryanjdillon (approved):
LGTM :hamburger:

gordo_components/cli/cli.py
```diff
@@ -61,14 +59,6 @@ def gordo():
...
     type=yaml.safe_load,
 )
-@click.option(
-    "--data-provider",
-    type=DataProviderParam(),
-    envvar="DATA_PROVIDER",
-    default='{"type": "DataLakeProvider"}',
-    help="DataProvider dict encoded as json. Must contain a 'type' key with the name of"
-    " a DataProvider as value.",
-)
```
milesgranger, line 71: Keep this, but set default to `None`

gordo_components/cli/cli.py
```diff
@@ -157,27 +138,6 @@ def build(
...
-    data_config["tag_list"] = data_config.pop("tags")
-
-    data_config["from_ts"] = dateutil.parser.isoparse(
-        data_config.pop("train_start_date")
-    )
-
-    data_config["to_ts"] = dateutil.parser.isoparse(data_config.pop("train_end_date"))
-
-    # Set default data provider for data config
-    data_config["data_provider"] = data_provider
```
milesgranger, line 175: Only set this, `if data_provider is not None`; for backwards compatibility.

gordo_components/dataset/datasets.py
```diff
@@ -72,19 +105,26 @@ def __init__(
...
             Default is zero 0
+        asset: Optional[str]
+            Asset for which the tags are associated with.
         _kwargs
         """
-        self.from_ts = from_ts
-        self.to_ts = to_ts
-        self.tag_list = normalize_sensor_tags(tag_list)
+        self.from_ts = from_ts if isinstance(from_ts, datetime) else isoparse(from_ts)
+        self.to_ts = to_ts if isinstance(to_ts, datetime) else isoparse(to_ts)
```
milesgranger: Do the same parsing while checking for timezone, in `workflow` component

milesgranger:
@flikka @epa095  Tested this with the test project (01.11.2019), and seems a-okay, mans!

milesgranger:
@flikka @epa095 ^^

I'm going to run with the Approval, less anyone objects by tomorrow. :+1:

milesgranger:
I've tested this once more (07.11.2019) and it has no problems.
image

PR #567, opened 2019-10-22
[SCHWIP] Integration testing with k8s

milesgranger:
First stab :hocho:


milesgranger:
..getting stuck on networking issues when launching the ml-server. Think it's something to do with the Istio / ambassador mix. Leaving for now until we fully switch to Istio and sort that out.

PR #569, opened 2019-10-23
Add .python-version for dependabot

milesgranger:
https://github.com/dependabot/dependabot-core/issues/1455#issuecomment-545389290


epa095:
Why did we close this? It seems like it might be an possible explanation for all the dependabot weirdness, e.g https://github.com/dependabot/feedback/issues/785#issuecomment-562391399

milesgranger:
~As I recall, and you'll like this, Dependabot only supports certain patch level versions of Python. None of which aligned with CircleCI at the time. So Circle would error saying the version specified in that file isn't available.~ edit: <- All this is completely false. :sweat_smile: 

Albeit, given the current trouble, probably worth diving deeper into maybe creating that environment in Circle somehow?

epa095 (approved):
Lets just try?

PR #572, opened 2019-10-23
Workflow server after ensure single wf

### LINKED ISSUE #571: Workflow: Start ml-server after ensure-single-workflow
`ensure-single-workflow` is intended to run at start to ensure we are the only version of the project running. "bad things" can happen otherwise, and ml-server should wait a bit before it starts.

epa095:
This closes #571

It also sets the ml-server as a dependency for the model-building. This ensures that before we start building all the models we have watchman and the server up, which is probably a good idea so we dont risk using all the resources to such a degree that the ml-server dont get to start.


epa095:
Tested and works

milesgranger (approved):

PR #573, opened 2019-10-24
Add MLflow logging to model builder

### LINKED ISSUE #441: Create and use model-logging using AzureML services/MLFlow
A suggested approach is to make a thin wrapper that supports "Experiment/Run" concepts, starting a run/build, log metrics etc, ending run/build. This wrapper then uses either MLFlow or AzureML services backend, switched in an elegant way, using for example a supplied provider.

Should log as much as possible, from the reading of data, building model, etc. during the gordo building process. The same stuff will be used in the "experimental" work.

First stab: Set up the system for the gordo model building on the cluster.
Second: Set up system for the experimental use (data science).

ryanjdillon:
Will close #441 

# Problem :airplane: 
The DS team would like to have additional logging of metrics for various configuration of deployed models for post-analysis.

# Solution :rocket: 
Add MLFlow logging to the the CLI `build` command and `gordo_components.builder.build_model::build_model`.

This includes:
- [x] Add a method that authenticates to AzureML and sets MLFlow logging to a wokspace. The configuration is passed to the CLI `build` as a dict, via the env var `MLFLOW_CONFIG`. Default behavior is to require interactive authentication and log MLFlow locally. 
- [x]  Set the workspace (with auth) at the CLI level. 
- [x] Log the `model_id` (changed from `cache_id`) at CLI level.
- [x] Log metadata at `build_model` level
- [x] Add test for `set_mlflow_tracking`
- [x] Add test for `build_model`

### DIFF
--- a/gordo_components/builder/local_build.py
+++ b/gordo_components/builder/local_build.py
@@ -10,10 +10,11 @@ from gordo_components.workflow.workflow_generator.workflow_generator import (
     get_dict_from_yaml,
 )
 from gordo_components.builder import ModelBuilder
+from gordo_components.builder.mlflow_utils import mlflow_context, log_metadata
 
 
 def local_build(
-    config_str: str, enable_remote_logging: bool = True, workspace_kwargs: dict = {}
+    config_str: str, enable_mlflow: bool = True, workspace_kwargs: dict = {}
 ) -> Iterable[Tuple[Union[BaseEstimator, None], dict]]:
     """
     Build model(s) from a bare Gordo config file locally.
@@ -77,10 +78,20 @@ def local_build(
     config = get_dict_from_yaml(io.StringIO(config_str))
     normed = NormalizedConfig(config, project_name="local-build")
     for machine in normed.machines:
-        yield ModelBuilder(
+        model, metadata = ModelBuilder(
             name=machine.name,
             model_config=machine.model,
             data_config=machine.dataset.to_dict(),
             metadata=machine.metadata,
             evaluation_config=machine.evaluation,
-        )._build()
+        ).build()
+
+        if enable_mlflow:
+            # This will enforce a single interactive login and automatically
+            # generate a new key for each model built
+            with mlflow_context(
+                name=machine.name, workspace_kwargs=workspace_kwargs
+            ) as (mlflow_client, run_id):
+                log_metadata(mlflow_client, run_id, metadata)
+
+        yield model, metadata
--- a/gordo_components/builder/local_build.py
+++ b/gordo_components/builder/local_build.py
@@ -12,7 +12,9 @@ from gordo_components.workflow.workflow_generator.workflow_generator import (
 from gordo_components.builder import ModelBuilder
 
 
-def local_build(config_str: str) -> Iterable[Tuple[Union[BaseEstimator, None], dict]]:
+def local_build(
+    config_str: str, enable_remote_logging: bool = True, workspace_kwargs: dict = {}
+) -> Iterable[Tuple[Union[BaseEstimator, None], dict]]:
     """
     Build model(s) from a bare Gordo config file locally.
 
@@ -25,6 +27,15 @@ def local_build(config_str: str) -> Iterable[Tuple[Union[BaseEstimator, None], d
     ----------
     config_str: str
         The raw yaml config file in string format.
+    enable_mlflow: bool
+        Flag to enable Mlflow logging of model building results. With `enable_mlflow`
+        set to `True`, passing `workspace_kwargs` an empty dict (default) will result in
+        local MLflow logging, while passing a dict of keyword arguments as defined in
+        :func:`~gordo_components.builder.mlflow_utils.get_mlflow_client` will result in
+        remote logging.
+    workspace_kwargs: dict
+        AzureML Workspace configuration to use for remote MLFlow tracking. See
+        :func:`gordo_components.builder.azure_utils.get_mlflow_client`.
 
     Examples
     --------
@@ -55,7 +66,7 @@ def local_build(config_str: str) -> Iterable[Tuple[Union[BaseEstimator, None], d
     ...                     estimator: sklearn.linear_model.base.LinearRegression
     ...         name: crazy-sweet-name
     ... '''
-    >>> models_n_metadata = local_build(config)
+    >>> models_n_metadata = local_build(config, enable_mlflow=False)
     >>> assert len(list(models_n_metadata)) == 1
 
     Returns
--- a/gordo_components/builder/mlflow_utils.py
+++ b/gordo_components/builder/mlflow_utils.py
@@ -1,0 +1,490 @@
+from contextlib import contextmanager
+from datetime import datetime
+import json
+import logging
+import os
+import tempfile
+from typing import List
+from uuid import uuid4
+
+from azureml.core import Workspace
+from azureml.core.authentication import (
+    InteractiveLoginAuthentication,
+    ServicePrincipalAuthentication,
+)
+from mlflow.entities import Metric, Param
+from mlflow.tracking import MlflowClient
+from pytz import UTC
+
+from gordo_components.dataset.sensor_tag import normalize_sensor_tags
+
+logger = logging.getLogger(__name__)
+
+
+class MlflowLoggingError(ValueError):
+    pass
+
+
+def _validate_dict(d: dict, required_keys: List[str]):
+    """
+    Validate the required keys are contained in provided dictionary
+
+    Parameters
+    ----------
+    d: dict
+        Dictionary to validate.
+    required_keys: List[str]
+        Keys that must be present in provided dictionary.
+    """
+    if any([key not in d for key in required_keys]):
+        raise MlflowLoggingError(
+            f"Required keys for this dictionary include {', '.join(required_keys)}."
+        )
+
+
+def get_mlflow_client(
+    workspace_kwargs: dict = {}, service_principal_kwargs: dict = {}
+) -> MlflowClient:
+    """
+    Set remote tracking URI for mlflow to AzureML workspace
+
+    Parameters
+    ----------
+    workspace_kwargs: dict
+        AzureML Workspace configuration to use for remote MLFlow tracking. An
+        empty dict will result in local logging by the MlflowClient.
+        Example::
+
+            `{
+                 "subscription_id":<value>,
+                 "resource_group":<value>,
+                 "workspace_name":<value>
+             }`
+    service_principal_kwargs: dict
+        AzureML ServicePrincipalAuthentication keyword arguments. An empty dict
+        will result in interactive authentication.
+        Example::
+
+            `{
+                 "tenant_id":<value>,
+                 "service_principal_id":<value>,
+                 "service_principal_password":<value>
+             }`
+
+    Returns
+    -------
+    client: mlflow.tracking.MlflowClient
+        Client with tracking uri set to AzureML if configured.
+    """
+    logger.info("Creating MLflow tracking client.")
+
+    tracking_uri = None
+
+    # Get AzureML tracking_uri if using Azure as backend
+    if workspace_kwargs:
+        required_keys = ["subscription_id", "resource_group", "workspace_name"]
+        _validate_dict(workspace_kwargs, required_keys)
+
+        msg = "Configuring AzureML backend with {auth_type} authentication."
+        if service_principal_kwargs:
+            required_keys = [
+                "tenant_id",
+                "service_principal_id",
+                "service_principal_password",
+            ]
+            _validate_dict(service_principal_kwargs, required_keys)
+
+            logger.info(msg.format(auth_type="ServicePrincipalAuthentication"))
+            workspace_kwargs["auth"] = ServicePrincipalAuthentication(
+                **service_principal_kwargs
+            )
+        else:
+            logger.info(msg.format(auth_type="interactive"))
+            workspace_kwargs["auth"] = InteractiveLoginAuthentication(force=True)
+
+        tracking_uri = Workspace(**workspace_kwargs).get_mlflow_tracking_uri()
+
+    return MlflowClient(tracking_uri)
+
+
+def get_run_id(client: MlflowClient, experiment_name: str, model_key: str) -> str:
+    """
+    Get an existing or create a new run for the given model_key and experiment_name.
+
+    The model key corresponds to a unique configuration of the model. The corresponding
+    run must be manually stopped using the `mlflow.tracking.MlflowClient.set_terminated`
+    method.
+
+    Parameters
+    ----------
+    client: mlflow.tracking.MlflowClient
+        Client with tracking uri set to AzureML if configured.
+    experiment_name: str
+        Name of experiment to log to.
+    model_key: str
+        Unique ID of model configuration.
+
+    Returns
+    -------
+    run_id: str
+        Unique ID of MLflow run to log to.
+    """
+    experiment = client.get_experiment_by_name(experiment_name)
+
+    experiment_id = (
+        getattr(experiment, "experiment_id")
+        if experiment
+        else client.create_experiment(experiment_name)
+    )
+
+    return client.create_run(experiment_id, tags={"model_key": model_key}).info.run_id
+
+
+def _datetime_to_ms_since_epoch(dt: datetime) -> int:
+    """
+    Convert datetime to milliseconds since Unix epoch (UTC)
+
+    Parameters
+    ----------
+    dt: datetime.datetime
+        Timestamp to convert (can be timezone aware or naive).
+
+    Returns
+    -------
+    dt: int
+        Timestamp as milliseconds since Unix epoch
+
+    Example
+    -------
+    >>> dt = datetime(1970, 1, 1, 0, 0)
+    >>> _datetime_to_ms_since_epoch(dt)
+    0
+    """
+    epoch = datetime.utcfromtimestamp(0).replace(tzinfo=dt.tzinfo)
+    dt = dt.astimezone(UTC) if dt.tzinfo else dt.replace(tzinfo=dt.tzinfo)
+    return round((dt - epoch).total_seconds() * 1000.0)
+
+
+def epoch_now() -> int:
+    """
+    Get current timestamp in UTC as milliseconds since Unix epoch.
+
+    Returns
+    -------
+    now: int
+        Milliseconds since Unix epoch.
+    """
+    return _datetime_to_ms_since_epoch(datetime.now(tz=UTC))
+
+
+def get_params(d: dict, keys: List[str]) -> List[Param]:
+    """
+    Extract list of keys and their values from a dictionary as MLflow Params
+
+    Parameters
+    ----------
+    d: dict
+        Dictionary to extract key, value pairs from.
+    keys: List[str]
+        List of keys to extract.
+
+    Returns
+    -------
+    param_list: List[Param]
+        List of extracted key, value pairs stored as `mlflow.entities.Param`s.
+    """
+    param_list: List[Param] = list()
+    for k in keys:
+        if k in d:
+            param_list.append(
+                Param(k, d[k].isoformat() if type(d[k]) is datetime else str(d[k]))
+            )
+    return param_list
+
+
+def get_metrics(d: dict, keys: List[str]) -> List[Metric]:
+    """
+    Extract list of keys and their values from a dictionary as MLflow Metrics
+
+    Parameters
+    ----------
+    d: dict
+        Dictionary to extract key, value pairs from.
+    keys: List[str]
+        List of keys to extract.
+
+    Returns
+    -------
+    param_list: List[Metric]
+        List of extracted key, value pairs stored as `mlflow.entities.Metric`s.
+    """
+    return [Metric(k, float(d[k]), epoch_now(), 0) for k in keys if k in d]
+
+
+def get_batch_kwargs(metadata: dict) -> dict:
+    """
+    Create flat lists of MLflow logging entities from multilevel dictionary
+
+    For more information, see the mlflow docs:
+    https://www.mlflow.org/docs/latest/python_api/mlflow.tracking.html#mlflow.tracking.MlflowClient.log_batch
+
+    Parameters
+    ----------
+    metadata: dict
+        Multi-level dictionary with (key, value) loggables to parse.
+
+    Returns
+    -------
+    batch_kwargs: dict
+        Dict with `metrics` and `param_list` lists for passing to
+        `mlflow.tracking.MlflowClient.log_batch`. The metrics key contains a list of
+        Mlflow `Metric` instances, and the param_list key contains a list of `Param`
+        instances.
+    """
+
+    param_list = get_params(metadata, ["project-name", "name"])
+
+    dataset_keys = [
+        "train_start_date",
+        "train_end_date",
+        "resolution",
+        "filter",
+        "row_filter_buffer_size",
+    ]
+    model_keys = ["model-creation-date", "model-builder-version", "model-offset"]
+    for section, keys in zip(["dataset", "model"], [dataset_keys, model_keys]):
+        param_list += get_params(metadata[section], keys)
+
+    metric_list: List[Metric] = list()
+
+    # Parse cross-validation metrics
+    if metadata["model"].get("cross-validation", {}).get("scores", None):
+        scores = metadata["model"]["cross-validation"]["scores"]
+        keys = sorted(list(scores.keys()))
+        subkeys = ["mean", "max", "min", "std"]
+
+        n_folds = len(scores[keys[0]]) - len(subkeys)
+        tag_list = normalize_sensor_tags(metadata["dataset"]["tag_list"])
+        for k in keys:
+            # Skip per tag data, produces too many params for MLflow
+            if any([t.name in k for t in tag_list]):
+                continue
+
+            # Summary stats per metric
+            for sk in subkeys:
+                metric_list.append(
+                    Metric(f"{k}_{sk}", scores[k][f"fold-{sk}"], epoch_now(), 0)
+                )
+            # Append value for each fold with increasing steps
+            metric_list += [
+                Metric(k, scores[k][f"fold-{i+1}"], epoch_now(), i)
+                for i in range(n_folds)
+            ]
+
+    # Parse fit metrics
+    if metadata["model"].get("history", None):
+        meta_params = metadata["model"]["history"]["params"]
+
+        metric_list += get_metrics(
+            metadata["model"],
+            ["data-query-duration-sec", "model-training-duration-sec"],
+        )
+        param_list += get_params(
+            meta_params, [p for p in meta_params if p != "metrics"]
+        )
+        for m in meta_params["metrics"]:
+            data = metadata["model"]["history"][m]
+            metric_list += [
+                Metric(m, float(x), timestamp=epoch_now(), step=i)
+                for i, x in enumerate(data)
+            ]
+
+    return {"metrics": metric_list, "params": param_list}
+
+
+def get_kwargs_from_secret(name: str, keys: List[str]) -> dict:
+    """
+    Get keyword arguments dictionary from secrets environment variable
+
+    Parameters
+    ----------
+    name: str
+        Name of the environment variable whose content is a colon separated
+        list of secrets.
+
+    Returns
+    -------
+    kwargs: dict
+        Dictionary of keyword arguments parsed from environment variable.
+    """
+    secret_str = os.getenv(name)
+
+    if secret_str is None:
+        raise MlflowLoggingError(f"The value for env var '{name}' must not be `None`.")
+
+    if secret_str:
+        elements = secret_str.split(":")
+        if len(elements) != len(keys):
+            raise MlflowLoggingError(
+                "`keys` len {len(keys)} must be of equal length with env var {name} elements {len(elements)}."
+            )
+        kwargs = {key: elements[i] for i, key in enumerate(keys)}
+    else:
+        kwargs = {}
+
+    return kwargs
+
+
+def get_workspace_kwargs() -> dict:
+    """Get AzureML keyword arguments from environment
+
+    The name of this environment variable is set in the Argo workflow template,
+    and its value should be in the format:
+    `<subscription_id>:<resource_group>:<workspace_name>`.
+
+    Returns
+    -------
+    workspace_kwargs: dict
+        AzureML Workspace configuration to use for remote MLFlow tracking. See
+        :func:`gordo_components.builder.mlflow_utils.get_mlflow_client`.
+    """
+    return get_kwargs_from_secret(
+        "AZUREML_WORKSPACE_STR", ["subscription_id", "resource_group", "workspace_name"]
+    )
+
+
+def get_spauth_kwargs() -> dict:
+    """Get AzureML keyword arguments from environment
+
+    The name of this environment variable is set in the Argo workflow template,
+    and its value should be in the format:
+    `<tenant_id>:<service_principal_id>:<service_principal_password>`
+
+    Returns
+    -------
+    service_principal_kwargs: dict
+        AzureML ServicePrincipalAuthentication keyword arguments. See
+        :func:`gordo_components.builder.mlflow_utils.get_mlflow_client`
+    """
+    return get_kwargs_from_secret(
+        "DL_SERVICE_AUTH_STR",
+        ["tenant_id", "service_principal_id", "service_principal_password"],
+    )
+
+
+class DatetimeEncoder(json.JSONEncoder):
+    """
+    Encode datetime.datetime objects as strings
+
+    Example
+    -------
+    >>> s = json.dumps({"now":datetime.now(tz=UTC)}, cls=DatetimeEncoder, indent=4)
+    >>> s = '{"now": "2019-11-22 08:34:41.636356+"}'
+    """
+
+    def default(self, obj):
+        if isinstance(obj, datetime):
+            return obj.strftime("%Y-%m-%d %H:%M:%S.%f+%z")
+        else:
+            return json.JSONEncoder.default(self, obj)
+
+
+def send_artifacts(client: MlflowClient, run_id: str, metadata: dict):
+    """
+    Send artifacts
+
+    Parameters
+    ----------
+    client: MlflowClient
+        MLflow tracking client.
+    run_id: str
+        MLflow Run ID.
+    metadata: dict
+        Metadata produced by :func:`~gordo_components.builder.build_model.build_model`
+        to extract configs from.
+    """
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        for name, obj in [
+            ("model-config", metadata["model"]["model-config"]),
+            ("dataset-config", metadata["dataset"]),
+            ("user-config", metadata["user-defined"]),
+        ]:
+            fp = os.path.join(tmp_dir, f"{name}.json")
+            json.dump(obj, open(fp, "w"), cls=DatetimeEncoder)
+        client.log_artifacts(run_id, local_dir=tmp_dir)
+
+
+@contextmanager
+def mlflow_context(
+    name: str,
+    model_key: str = uuid4().hex,
+    workspace_kwargs: dict = {},
+    service_principal_kwargs: dict = {},
+):
+    """
+    Generate MLflow logger function with either a local or AzureML backend
+
+    Parameters
+    ----------
+    name: str
+        The name of the log group to log to (e.g. a model name).
+    model_key: str
+        Unique ID of logging run.
+    workspace_kwargs: dict
+        AzureML Workspace configuration to use for remote MLFlow tracking. See
+        :func:`gordo_components.builder.mlflow_utils.get_mlflow_client`.
+    service_principal_kwargs: dict
+        AzureML ServicePrincipalAuthentication keyword arguments. See
+        :func:`gordo_components.builder.mlflow_utils.get_mlflow_client`
+
+    Example
+    -------
+    >>> with tempfile.TemporaryDirectory as tmp_dir:
+    ...     mlflow.set_tracking_uri(f"file:{tmp_dir}")
+    ...     with mlflow_context("log_group", "unique_key", {}, {}) as (mlflow_client, run_id):
+    ...         log_metadata(metadata) # doctest: +SKIP
+    """
+    mlflow_client = get_mlflow_client(workspace_kwargs, service_principal_kwargs)
+    run_id = get_run_id(mlflow_client, experiment_name=name, model_key=model_key)
+
+    logger.info(
+        f"MLflow client configured to use {'AzureML' if workspace_kwargs else 'local backend'}"
+    )
+
+    yield mlflow_client, run_id
+
+    mlflow_client.set_terminated(run_id)
+
+
+def log_metadata(mlflow_client, run_id, metadata: dict):
+    """
+    Send logs to configured MLflow backend
+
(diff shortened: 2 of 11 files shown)


ryanjdillon:
`azureml-core` has a dependency on `ruamel.yaml<=0.15.89,>=0.15.35`, which breaks 39 tests that are using `ruamel.yaml==0.16.5`.

```
E               yaml.scanner.ScannerError: mapping values are not allowed here
E                   in "<unicode string>", line 2, column 34:
E                       Traceback (most recent call last):
```

We could craft the URL ourselves to pass to MLflow, but this then requires an auth object of class `[ServicePrincipalAuthentication](https://docs.microsoft.com/en-us/python/api/azureml-core/azureml.core.authentication.serviceprincipalauthentication?view=azure-ml-py)`, which also requires `azureml-core`, leading down a :rabbit:  hole.  

Going to mothball this for now and either these things get moved to a more stable release soon and updated, or we decide to make some fix if it is wanted enough.

```
def _get_mlflow_tracking_uri(self, with_auth=True):
    """
    Retrieve the tracking URI from Workspace for use in AzureMLflow.

    Return a URI identifying the workspace, with optionally the auth header
    embedded as a query string within the URI. The authentication header does not include the "Bearer " prefix.

    :return: Returns the URI pointing to this workspace, with the auth query paramter if with_auth is True.
    :rtype: str
    """
    queries = []
    if with_auth:
        auth = self._auth_object
        logger.debug("Creating a tracking uri with an {} auth token".format(auth.__class__.__name__))
        header = auth.get_authentication_header()
        token = header["Authorization"][len(_TOKEN_PREFIX):]
        queries.append(_AUTH_TYPE + "=" + auth.__class__.__name__)
        queries.append(_AUTH_HEAD + "=" + token)

    service_location = parse.urlparse(self.service_context._get_run_history_url()).netloc
    workspace_scope = self.service_context._get_workspace_scope()
    logger.debug("Creating a tracking uri in {} for workspace {}".format(service_location, workspace_scope))

    return "azureml://{}/history/v1.0{}{}".format(
        service_location,
        workspace_scope,
        "?" + "&".join(queries) if queries else "")
```

ryanjdillon:
@flikka  the [MLflow docs](https://www.mlflow.org/docs/latest/tracking.html#where-runs-are-recorded) state the following as being supported alternatives for remote tracking of "runs":
1. Local file path (specified as `file:/my/local/dir`), where data is just directly stored locally.
2. Database encoded as `<dialect>+<driver>://<username>:<password>@<host>:<port>/<database>`. Mlflow supports the dialects mysql, mssql, sqlite, and postgresql. For more details, see SQLAlchemy database uri.
3.  HTTP server (specified as `https://my-server:5000`), which is a server hosting an MLFlow tracking server.
4. Databricks workspace (specified as databricks or as `databricks://<profileName>`, a Databricks CLI profile. See docs on logging to Databricks-hosted MLflow, or the quickstart to easily get started with hosted MLflow on Databricks Community Edition.

The MLFlow tracking server [supports Azure blob storage](https://www.mlflow.org/docs/latest/tracking.html#azure-blob-storage), so I'm going to give a shot at setting up a simple server that runs on an Azure Function and go from there.

ryanjdillon:
Load testing AzureML revealed that there are of coarse rate limits on these MLflow log requests. As per the [Azure rate limit docs](https://docs.microsoft.com/en-us/azure/azure-subscription-service-limits#azure-machine-learning-service-limits), it looks like if we had gordo on its own subscription, writes are capped at 1200/hr. 


If the 10k were to complete within an hour or two this won't cut it. I'm checking with Igor on the Aurora team to see what level we could get the limit to.

For reference, [here](https://gist.github.com/ryanjdillon/a5ac4f1fb9fe66ed5f0e7cb7b705f13c) is the current load-testing script.

ryanjdillon:
Per conversation with @henrikny @flikka @epa095 , We will first continue load testing the AzureML server to ascertain the its use in terms of query flexibility and performance (e.g. request read/write rate limits, performance with 100k rows, etc).

Henrik will confirm with Alexis who the the project owner and potential users of this logging service/database will be. 

If the AzureML service does not meet load required by Gordo deployments or is not flexible enough for the expected end user(s), then we will revisit the possibility of using an [Azure Postgres DB](https://azure.microsoft.com/en-us/services/postgresql/) , /[Azure Kusto](https://docs.microsoft.com/en-us/azure/kusto/concepts/) solution.

Information that the DS team would like logged is [documented here](https://github.com/equinor/gordo-analysis/blob/master/documentation/work_notes.md#gordo-output). Namely:

>Metrics
>
>For each scaled variable and for each time series cross-validation split:
>
> -   Mean squared errors
> -   Explained variance
> -   R^2
> -   Variable threshold
>
>In addition to this, the complete model threshold should also be returned.
>
>Metadata
>
> -   Start and end date for each fold.
> -   Model ID
> -   ModelTag
> -   ModelHash
> -   ModelID+Tag
> -   Complete config information (parsed yaml)
> -   Data query duration
> -   Training time duration

ryanjdillon:
For reference, [here is the script](https://gist.github.com/ryanjdillon/a5ac4f1fb9fe66ed5f0e7cb7b705f13c) for testing logging to AzureML with MLflow.

I'm currently populating the service with a simulated 2 deployments, where there are 9000 experiments, corresponding to each machine, with two runs (i.e. deployments) each. This means 1800 rows, each with 14 logs each (see the script above and the `log_single_build` method.

To query these, you can use the [MlflowClient](https://www.mlflow.org/docs/latest/python_api/mlflow.tracking.html#mlflow.tracking.MlflowClient).

Example using funcs from load testing script
```python
auth = InteractiveLoginAuthentication(force=True)
client = get_tracking_client(auth)
experiments = client.list_experiments()
# Here name represents what the model name would be, so in real life you'd 
# search for "precomp" or "TrollA" or something
experiment_ids = [exp.experiment_id for exp in experiments if exp.name.startswith("test_b")

# Using order by a metric
runs = client.search_runs(experiment_ids=experiment_ids, max_results=50, order_by=["metrics.r_2"])
# Using an SQL filter string
runs = client.search_runs(experiment_ids=experiment_ids, filter_string='params.model_id = "193845098392472" AND metrics.r_2 <= 0.2',  max_results=10)
```

gordo_components/builder/azure_utils.py
```diff
@@ -0,0 +1,192 @@
...
+        Workspace parameters with a missing our empty `auth` key will result in
+        interactive authentication to AzureML.
+
+        Example::
+
+            `{
+                 "subscription_id":<value>,
+                 "resource_group":<value>,
+                 "workspace_name":<value>
+                 "auth": {
```
milesgranger: Think it's more understandable if this is a separate parameter `service_principal_auth_kwargs` with a reference that this is actually kwargs to be supplied to `ServicePrincipalAuthentication` 

And the others are `Workspace` kwargs, no? (less the injection of 'auth' if such kwargs are supplied of course)

gordo_components/builder/azure_utils.py
```diff
@@ -0,0 +1,192 @@
...
+def get_loggables(
+    d: dict, parent_key: str = None
+) -> Iterator[Tuple[str, Union[str, int, float]]]:
+    """
+    Recursively generate key, value pairs form multilevel dictionary for logging
+
+    Parameters
+    ----------
+    d: dict
+        Dictionary to flatted to loggable (key, value) pairs.
```
milesgranger: "Dictionary to flatted to loggable..."

gordo_components/builder/build_model.py
```diff
@@ -61,6 +67,8 @@ def build_model(
         Mapping of the Dataset to initialize, following the same logic as model_config.
     metadata: dict
         Mapping of arbitrary metadata data.
+    azureml_config: mlflow.tracking.MlflowClient
+        Client with tracking uri set to AzureML.
```
milesgranger: Not actually a parameter, and I would strongly urge against making it as such. Current parameters here are suppose to represent basically individual keys per machine, where the contents of each are quite literally the loaded yaml's dump of each corresponding key.

gordo_components/builder/build_model.py
```diff
@@ -189,6 +218,15 @@ def build_model(
...
     metadata["model"].update(_get_metadata(model))
+
+    # Log metadata to MLflow
+    mlflow_client = get_mlflow_client(azureml_config)
+    run_id = get_run_id(
+        mlflow_client, experiment_name=name, model_key=azure_config["model_key"]
+    )
+    mlflow_client.log_batch(run_id, **get_batch_kwargs(metadata))
+    mlflow_client.set_terminated(run_id)
+
```
milesgranger: So that argument above, it seems like one could rather create a decorator/wrapper to this function which snaps up any input required, ie calculating the 'model_key', executing the function, grabbing the metadata, logging it, and passing it along. 

Advantage here is we're not integrating _any_ specific logging client, and sets us up for different or multiple logging clients without also in twinning them into this function.

milesgranger (commented):
Didn't look too deeply, as it's a WIP, but made a few comments to consider along the way. :mountain_cableway:

gordo_components/builder/azure_utils.py
```diff
@@ -0,0 +1,192 @@
...
+    Returns
+    -------
+    client: mlflow.tracking.MlflowClient
+        Client with tracking uri set to AzureML if configured.
+    """
+    logger.info("Creating MLflow tracking client.")
+
+    # Get AzureML tracking_uri if using Azure as backend
+    tracking_uri = None
+    if azureml_config:
```
flikka: Can this if test ever not trigger? Unless someone explicitly sends in azureml_config = None / False?

I would perhaps expect that it was default "None", and if so, use "local" storage (the default mlflow folder or whatever it is).

gordo_components/builder/build_model.py
ryanjdillon: After looking at the decorator I setup again, it can be used it as a wrapper type thing with some slight changes, while still being able to use it as a decorator if desired. I'll just put it setup like that and see what you think.

gordo_components/builder/azure_utils.py
ryanjdillon: Agreed :+1: 

Right, those 4 are Workspace args, where `auth` is either an `InteractiveLoginAuthentication` or `ServicePrincipalAuthentication` instance, the latter requiring the kwars shown in the `auth` example.

gordo_components/builder/azure_utils.py
```diff
@@ -0,0 +1,192 @@
...
+        Example::
+
+            `{
+                 "subscription_id":<value>,
+                 "resource_group":<value>,
+                 "workspace_name":<value>
+                 "auth": {
+                     "tenant_id":<value>,
+                     "service_principal_id":<value>,
+                     "service_principal_password":<value>
```
flikka: I feel we must (unfortunately?) look at the Datalake authentication we already have. It appears that it is using service principal, with much the same data (tenant/id/secret). In a way it is good, but this "triplet" already comes in as an environment variable, so you could probably just re-use it.

gordo_components/cli/cli.py
```diff
@@ -101,6 +101,9 @@ def gordo():
     ),
     type=yaml.safe_load,
 )
+@click.option(
```
flikka, line 109: Maybe a bit premature to comment on this, but I think we must split the "which AZUREML workspace to use" and the "service principal" data. The service principle data is already available as an environment variable (on the cluster). YOHO!

gordo_components/builder/azure_utils.py
```diff
@@ -0,0 +1,192 @@
...
+        raise ValueError(
+            f"Timestamp ({timestamp} needs to include timezone information"
+        )
+
+    if type(step) is not int:
+        raise ValueError(f"`step` argument must be of type '{str(int)}' not '{type(step)}'")
+
+    metrics = list()
+    params = list()
+    for k, v in get_loggables(d):
```
flikka: Suppose this kind of makes sense, although I think it will be a bit random in that things that are "numeric" parameters will be logged as metrics. But not sure it really matters.

gordo_components/builder/azure_utils.py
```diff
@@ -0,0 +1,192 @@
...
+    """
+    experiment = client.get_experiment_by_name(experiment_name)
+
+    experiment_id = (
+        getattr(experiment, "experiment_id")
+        if experiment
+        else client.create_experiment(experiment_name)
+    )
+
+    return client.create_run(experiment_id, tags={"model_key": model_key}).info.run_id
```
flikka: Do you know what the difference between a "tag" and the metrics/params is?

flikka (commented):
Just went through it to get up to date with the work, tried to add some maybe helpful comments :-) Will be exciting to see in the wild, pumping out those metrics! 👍

gordo_components/builder/azure_utils.py
ryanjdillon: These tags are for adding some metadata to the `mlflow` `Run`. Doing it this way here perhaps doesn't make as much sense as logging it as a "param"; though, that would require creating `model_key` again inside `build_model` and added to the metadata there.

gordo_components/builder/azure_utils.py
ryanjdillon: I'm not sure how I would do it differently while still having it be able to generally take a dict of unkown structure. The difference between `Param` and `Metric` instances is the latter has a timestamp/step associated with them and they are populated at plottables on the frontend.

gordo_components/builder/azure_utils.py
ryanjdillon: We already talked about this, but just in case any other perusers have additional thoughs: I'm using the `azureml.core.authentication.ServicePrincipalAuthentication` class for auth. Not sure if those are the same, but I'll that those secrets will work and report back here.

gordo_components/builder/azure_utils.py
ryanjdillon: Empty dicts evaluate to False, so things here behave as suggested.

ryanjdillon:
It looks like various bits are in line now. I need to do the integration testing tomorrow morning, but otherwise I think things are ready for more in depth review :+1:

gordo_components/builder/azure_utils.py
```diff
@@ -0,0 +1,277 @@
...
+
+
+def _is_number(x: Any) -> bool:
```
milesgranger: I suggest you remove this altogether. It is functionally called only once, but adds over 50 lines of code. :confused: 
I think it'd be better do perform the simple try/catch right at the needed call.

gordo_components/builder/azure_utils.py
```diff
@@ -0,0 +1,277 @@
...
+    """
+    valid = True
+    try:
+        float(x)
+    except (TypeError, ValueError):
+        valid = False
+    return valid
+
+
+def _datetime_to_ms_since_epoch(dt) -> int:
```
milesgranger: I also suggest removing this, as it too is only called once and takes up lines as the function definition and the tests for something which seems rather concise anyway.

gordo_components/builder/build_model.py
```diff
@@ -192,6 +189,7 @@ def build_model(
     }
 
     metadata["model"].update(_get_metadata(model))
+
```
milesgranger: Are all these changes in the `build_model` needed?

gordo_components/builder/local_build.py
```diff
@@ -55,8 +61,8 @@ def local_build(config_str: str) -> Iterable[Tuple[Union[BaseEstimator, None], d
     ...                     estimator: sklearn.linear_model.base.LinearRegression
     ...         name: crazy-sweet-name
     ... '''
-    >>> models_n_metadata = local_build(config)
-    >>> assert len(list(models_n_metadata)) == 1
+    >>> models_n_metadata = local_build(config) # doctest: +SKIP
+    >>> assert len(list(models_n_metadata)) == 1 # doctest: +SKIP
```
milesgranger: Why are we skipping these?

gordo_components/builder/local_build.py
```diff
@@ -66,7 +72,10 @@ def local_build(config_str: str) -> Iterable[Tuple[Union[BaseEstimator, None], d
     config = get_dict_from_yaml(io.StringIO(config_str))
     normed = NormalizedConfig(config, project_name="local-build")
     for machine in normed.machines:
-        yield build_model(
+        build_model_wrapped = LogSender(
```
milesgranger: Does this force me to use logging when using `local_build`?

gordo_components/builder/logsender.py
```diff
@@ -0,0 +1,127 @@
...
+    3
+    """
+
+    def __init__(
+        self,
+        func: Callable = None,
+        name: str = "",
+        model_key: str = "",
+        workspace_kwargs: dict = {
+            "subscription_id": "019958ea-fe2c-4e14-bbd9-0d2db8ed7cfc",
```
milesgranger: This seems weird to have here, I don't know much about it though...but _very_ strange to have this in an open source project.

gordo_components/builder/logsender.py
```diff
@@ -0,0 +1,127 @@
...
+        *args: dict
+            Keyword arguments to pass to the function whose metadata is logged.
+
+        Returns
+        -------
+        first: Any
+            The first returned value of `self.func`.
+        d: dict
+            Multilevel dictionary of key, value pairs to be logged.
+        """
```
milesgranger: Not sure about this, but seems that all this info about what is essentially this wrapper, should rather be in the `init` or class declaration and then `self.func = functools.wraps(func)` then any introspecting would reflect the wrapped function.

gordo_components/cli/cli.py
```diff
@@ -110,6 +111,7 @@ def build(
     print_cv_scores,
     model_parameter,
     evaluation_config,
+    #    workspace_config,
```
milesgranger: Delete

tests/gordo_components/builder/test_azure_utils.py
```diff
@@ -0,0 +1,338 @@
...
+    _validate_dict,
+    MlflowClient,
+    get_mlflow_client,
+    get_run_id,
+    get_loggables,
+    _is_number,
+    _datetime_to_ms_since_epoch,
+    get_batch_kwargs,
+    get_spauth_kwargs,
+)
```
milesgranger: I suggest `from gordo_components.builder import azure_utils as au` or something. All these free functions floating around, and would save 10 lines. ;-)

gordo_components/builder/azure_utils.py
```diff
@@ -0,0 +1,277 @@
...
+    ----------
+    d: dict
+        Dictionary to validate.
+    name: str
+        Name of dictionary to validate.
+    required_keys: List[str]
+        Keys that must be present in provided dictionary.
+    """
+    if any([key not in d for key in required_keys]):
+        raise ValueError(f"`{name}` must contain keys {', '.join(required_keys)}.")
```
milesgranger: Why are we giving the dictionary a name?

tests/gordo_components/builder/test_azure_utils.py
```diff
@@ -0,0 +1,338 @@
...
+        assert MockWorkspace.call_count == n_get_uri
+        assert MockClient.called_once()
+
+
+@mock.patch("gordo_components.builder.azure_utils.InteractiveLoginAuthentication")
+@mock.patch("gordo_components.builder.azure_utils.Workspace")
+@mock.patch("gordo_components.builder.azure_utils.MlflowClient")
+def test_get_mlflow_client_config(MockClient, MockWorkspace, MockInteractiveLogin):
+    """
+    Test that error is raises with incomplete kwargs
```
milesgranger: error is raises

milesgranger (commented):
There is a _lot_ of code here... (1k+ lines!) I'm taking a break before going more into the tests...

tests/test_formatting.py
```diff
@@ -1,24 +1,22 @@
...
+        "black",
+        "--check",
+        "-v",
+        gordo_components_path,
+        tests_path,
+        "--exclude",
+        r".*_version.py",
+    ]
+    exit_code = os.system(" ".join(cmd))
+    assert exit_code == 0
```
milesgranger: I would suggest not doing any conversions of unittests to pytests in this PR, enough changes are going on here; just pop all these out into a separate PR please.

gordo_components/builder/azure_utils.py
```diff
@@ -0,0 +1,277 @@
+from typing import Any, List
+import os
+import logging
+from pytz import UTC
```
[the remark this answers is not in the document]
ryanjdillon: Aha... got it. Is there any style for organizing `from _ import` vs `import _`?

gordo_components/builder/azure_utils.py
```diff
@@ -0,0 +1,277 @@
...
+    for k, v in get_loggables(d):
+        if _is_number(v):
+            metrics.append(Metric(k, v, timestamp=timestamp_int, step=step))
+        else:
+            v_str = " ".join([s.strip() for s in str(v).split()])
+            params.append(Param(k, v_str[:250]))
+            # else:
+            #    raise NotImplementedError(
+            #        f"Parameter value exceeds Mlflow character limit. This should be saved as an artifact, which is not yet implemented, key: {k}, value: {v}."
+            #    )
```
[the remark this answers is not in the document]
ryanjdillon: Ewww.... :grimacing:

gordo_components/builder/logsender.py
ryanjdillon: I like the sound of that :+1:

gordo_components/builder/azure_utils.py
ryanjdillon: That's strictly to make the error more verbose while generalizing it to be reused.

tests/test_formatting.py
ryanjdillon: I'll pull this out to another PR :+1:

gordo_components/builder/azure_utils.py
```diff
@@ -0,0 +1,470 @@
...
+
+    def get_params(d: dict, keys: List[str], prefix: str = "") -> List[Param]:
+        """Get list of single summary metrics in dictionary defined by keys"""
+        return [Param(k, to_str(d[k])) for k in keys if k in d]
+
+    def get_metrics(d: dict, keys: List[str], prefix: str = "") -> List[Metric]:
+        """Get list of single summary metrics in dictionary defined by keys"""
+        return [Metric(k, float(d[k]), ts_int, 0) for k in keys if k in d]
+
+    metric_list: List[Metric] = list()
```
flikka: Could you just call the get_metrics to initialize? Seems a bit unneccessarry to first have an empty list, when you have a method that is called in the next step?

gordo_components/builder/azure_utils.py
```diff
@@ -0,0 +1,470 @@
...
+    >>> dt = datetime(1970, 1, 1, 0, 0)
+    >>> _datetime_to_ms_since_epoch(dt)
+    0
+    """
+    epoch = datetime.utcfromtimestamp(0).replace(tzinfo=dt.tzinfo)
+    dt = dt.astimezone(UTC) if dt.tzinfo else dt.replace(tzinfo=dt.tzinfo)
+    return round((dt - epoch).total_seconds() * 1000.0)
+
+
+def get_batch_kwargs(
```
flikka: This function is in my opinion a bit too long, and I am not so keen on the nested functions and imports. I think it reads more easily if the main steps (the function definition) is a bit shorter, and rather extracts parts to different functions? But I am not going to die fighting for it, just my feeling.

gordo_components/builder/azure_utils.py
```diff
@@ -0,0 +1,470 @@
...
+
+
+class DatetimeEncoder(json.JSONEncoder):
+    """
+    Encode datetime.datetime objects as strings
+
+    Example
+    -------
+    >>> s = json.dumps({"now":datetime.now(tz=UTC)}, cls=DatetimeEncoder, indent=4)
+    >>> s = '{"now": "2019-11-22 08:34:41.636356+"}'
```
flikka: I don't understand this example, where did the timezone go?

gordo_components/builder/azure_utils.py
```diff
@@ -0,0 +1,470 @@
...
+    Send artifacts
+
+    Parameters
+    ----------
+    client: MlflowClient
+        MLflow tracking client.
+    run_id: str
+        MLflow Run ID.
+    metadata: dict
+        Mode builder metadata to extract configs from.
```
flikka: Is there a typo here? Mode builder?

gordo_components/builder/azure_utils.py
```diff
@@ -0,0 +1,470 @@
...
+        # Log params and metrics
+        mlflow_client.log_batch(run_id, **get_batch_kwargs(metadata))
+
+        # Send configs as JSON artifacts
+        try:
+            send_artifacts(mlflow_client, run_id, metadata)
+        except Exception as e:
+            logger.error(e)
+
+        mlflow_client.set_terminated(run_id)
```
flikka: This set_terminated looks like something that is a bit important to ensure gets called? Should it be in some finally block or something? In case an exception is thrown or something? What happens if this call fails, do we need to do some cleanup next time we log to this run_id?

gordo_components/builder/azure_utils.py
```diff
@@ -0,0 +1,470 @@
...
+
+        logger.info(
+            f"Logging dictionary with MLflow to {'AzureML' if workspace_kwargs else 'local backend'}"
+        )
+
+        mlflow_client = get_mlflow_client(workspace_kwargs, service_principal_kwargs)
+        run_id = get_run_id(mlflow_client, experiment_name=name, model_key=model_key)
+
+        # Log params and metrics
+        mlflow_client.log_batch(run_id, **get_batch_kwargs(metadata))
```
flikka: Do we ever use the other arguments, the datetime or the step? Not sure what the step argument is though...

gordo_components/cli/cli.py
```diff
@@ -169,7 +178,10 @@ def build(
                 Example::
 
                     {"cv_mode": "cross_val_only"}
+    enable_remote_logging: bool
+        Flag to enable remote logging of model building results.
```
flikka: Maybe add "to Azure ML services" and that otherwise a local mlflow folder will be created?

gordo_components/cli/cli.py
```diff
@@ -238,26 +248,32 @@ def build(
...
                 print(score)
+
+        # If enabled, configure remote logging to AzureML, otherwise logs locally
+        workspace_kwargs = get_workspace_kwargs() if enable_remote_logging else {}
+        service_principal_kwargs = get_spauth_kwargs() if enable_remote_logging else {}
+        mlflow_logger = get_mlflow_logger(
+            name, model_key, workspace_kwargs, service_principal_kwargs
+        )
+
+        mlflow_logger(metadata)
```
flikka: This confuses me, is it a "factory" that produces a function? And then call the function?

gordo_components/builder/azure_utils.py
```diff
@@ -0,0 +1,467 @@
...
+    https://www.mlflow.org/docs/latest/python_api/mlflow.tracking.html#mlflow.tracking.MlflowClient.log_batch
+
+    Parameters
+    ----------
+    metadata: dict
+        Multi-level dictionary with (key, value) loggables to parse.
+    timestamp: datetime
+        Timezone aware timestamp of time to associate with all metrics logged.
+    step: int
+        Training step to associate with all metrics.
```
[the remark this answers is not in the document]
ryanjdillon: Here step is synonymous with fold. I thought I'd keep things named as they are where they're used in MLflow methods, but perhaps just calling this "fold" here would be clearer.

gordo_components/builder/azure_utils.py
ryanjdillon: just so i don't forget... also rename module to mlflow_utils or something

gordo_components/cli/cli.py
flikka: I am just a bit stuck in the non-functional world, just ignore me - this is just fine for me 👍

flikka (changes requested):
Left a bunch of comments - not much major stuff. Also, suggest renaming the azure_utils file to something else, maybe mlflow_utils?

gordo_components/builder/local_build.py
```diff
@@ -10,9 +10,12 @@
     get_dict_from_yaml,
 )
 from gordo_components.builder.build_model import build_model
+from gordo_components.builder.azure_utils import get_mlflow_logger
 
 
-def local_build(config_str: str) -> Iterable[Tuple[Union[BaseEstimator, None], dict]]:
+def local_build(
+    config_str: str, enable_remote_logging: bool = True, workspace_kwargs: dict = {}
```
flikka: Should it not be default = False?

gordo_components/builder/azure_utils.py
ryanjdillon: ok renamed and refactored some things out to make the metadata parsing more digestible.

gordo_components/builder/local_build.py
ryanjdillon: This should really not have been called that. For the local builder, I instead wanted it to be default log locally to Mlflow, but one could disable Mlflow entirely. While in the CLI, disabling MLflow is just allowing it to log locally rather than log remotely. I've renamed and updated things to make that clearer.

gordo_components/builder/local_build.py
```diff
@@ -25,6 +28,11 @@ def local_build(config_str: str) -> Iterable[Tuple[Union[BaseEstimator, None], d
     ----------
     config_str: str
         The raw yaml config file in string format.
+    enable_mlflow: bool
+        Flag to enable Mlflow logging of model building results.
```
milesgranger: Should this not be inferred from any `workspace_kwargs`?

gordo_components/cli/cli.py
```diff
@@ -19,6 +19,12 @@
 
 from gordo_components.builder.build_model import ModelBuilder
 from gordo_components import serializer
+from gordo_components.builder.mlflow_utils import (
+    get_workspace_kwargs,
+    get_spauth_kwargs,
+    mlflow_context,
+    log_metadata,
```
milesgranger: Possible to import `mlflow_utils` and use the functions from there? For me, that's easier to follow the code

gordo_components/cli/cli.py
```diff
@@ -231,12 +246,21 @@ def build(
...
+        ) as (mlflow_client, run_id):
+            log_metadata(mlflow_client, run_id, metadata)
+
     except Exception as e:
         exit_code = EXCEPTION_TO_EXITCODE.get(e.__class__, 1)
         traceback.print_exc()
         sys.exit(exit_code)
-    else:
-        return 0
+
```
milesgranger: Small thing, but seems rebasing removed the `else` here.

tests/gordo_components/builder/test_mlflow_utils.py
```diff
@@ -0,0 +1,406 @@
...
+    mock_client.log_batch.return_value = "test"
+
+    # Function with a metadata dict returned
+    with mlu.mlflow_context("returns metadata", "unique_key", {}, {}) as (
+        mlflow_client,
+        run_id,
+    ):
+        mlu.log_metadata(mlflow_client, run_id, metadata)
+
+    assert mock_client.log_batch.called
```
milesgranger, line 406: I hope none of these tests ever break.. :sweat_smile: I definitely don't want to try fixing them.

tests/test_examples.py
```diff
@@ -28,54 +28,53 @@ def _fake_data():
...
 
-                    importlib.import_module(os.path.basename(tmpdir), ".")
+                # Import this module to 'run' the code
+                module_dir = os.path.join(tmpdir, "..")
+                sys.path.insert(0, module_dir)
 
-                    sys.path.remove(module_dir)
+                importlib.import_module(os.path.basename(tmpdir), ".")
+
+                sys.path.remove(module_dir)
```
milesgranger: Again, I think the PR is enough as it is, migrating tests to pytest is probably better left for a different PR.

tests/gordo_components/cli/test_cli.py
```diff
@@ -39,207 +32,199 @@
 logger = logging.getLogger(__name__)
 
 
-class CliTestCase(unittest.TestCase):
+@pytest.fixture
+def runner(tmp_dir):
```
milesgranger, line 36: Is this required anytime we want to run some CLI tests against anything which would build the model with the CLI?

milesgranger (commented):
In general, I _think_ it looks ok.. but seems like we're definitely indulging in some fairly significant technical debt which I can't fully wrap my mind around yet. :man_shrugging: Likely my last review here and I'll let the adults make any final comments/approvals.

tests/gordo_components/cli/test_cli.py
ryanjdillon: Not *required*, but recommended. Else you'll end up with an `mlruns/` dir. Either that or you'd have to patch the context manager to use a mock MlflowClient.

gordo_components/builder/local_build.py
ryanjdillon: I did it like this, as you could potentially do a local build and log to a different remote workspace.

ie.
`enable_mlflow=True, workspace_kwargs={}` -> local logging
`enable_mlflow=True, workspace_kwargs={workspace_id, ...}` -> remote logging
`enable_mlflow=False, workspace_kwargs={} or {workspace_id, ...}` -> no logging

gordo_components/cli/cli.py
ryanjdillon: That was just some clean up, now it is like this
```python
try:
    exit_code <- build()
exception:
    sys.exit(exit_code)
return 0
```

tests/gordo_components/builder/test_mlflow_utils.py
ryanjdillon: No way, these tests are like :tropical_drink: :sunglasses: :palm_tree: .

tests/test_examples.py
ryanjdillon: I'll pull this out to another PR.

gordo_components/cli/cli.py
milesgranger: https://github.com/equinor/gordo-components/blob/4098901a0da5222a5f27b0338865e089610ac735/gordo_components/cli/cli.py#L238

ryanjdillon: Ah... I fixed that already locally and apparently haven't pushed that. That said, I see now the point of the `try/except/else` pattern here, so changing it back :+1:

gordo_components/builder/local_build.py
ryanjdillon: I added some extra explanation to the docstring to clarify that a bit :+1:

milesgranger: Ok, so it seems tied to a concept of `workspace` for remote logging. Otherwise I would have imagined that it would be something like `logging_kwargs` or something which could determine the difference. Anway, just ignore my comment. I think _that_ would involve much more changes.

ryanjdillon:
The tests pasts last go-around, just doing a rebase. Also running a test project named `test-mlflow18`. So long as that finishes fine, it should all be good to go.

ryanjdillon:
Test project passed. This is now ready to be merged and break things! :beer: :hammer:

gordo_components/builder/mlflow_utils.py
```diff
@@ -0,0 +1,493 @@
...
+    Parameters
+    ----------
+    d: dict
+        Dictionary to validate.
+    required_keys: List[str]
+        Keys that must be present in provided dictionary.
+    """
+    if any([key not in d for key in required_keys]):
+        raise MlflowLoggingError(
+            f"`Required keys for this dictionary include {', '.join(required_keys)}."
```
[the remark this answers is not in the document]
ryanjdillon: Most absolutely definitely not :)

gordo_components/builder/mlflow_utils.py
```diff
@@ -0,0 +1,493 @@
...
+    logger.info(
+        f"MLflow client configured to use {'AzureML' if workspace_kwargs else 'local backend'}"
+    )
+
+    try:
+        yield mlflow_client, run_id
+    except Exception:
+        raise MlflowLoggingError("Generation of MlflowClient and run_id failed.")
+    finally:
+        mlflow_client.set_terminated(run_id)
```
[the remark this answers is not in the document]
ryanjdillon: In looking at this now, I don't really recall any reason for doing this catching in the first place. Removing :+1:

flikka:
Trying to run it now, on gordodev49, getting this warning:

Could not import from mlflow. Please upgrade to Mlflow 1.4.0 or higher.

tests/gordo_components/builder/test_mlflow_utils.py
```diff
@@ -0,0 +1,406 @@
+from datetime import datetime
+import pytest
+
+import mlflow
+from mlflow.entities import Metric, Param
+import mock
+
+import gordo_components.builder.mlflow_utils as mlu
+from gordo_components.dataset.sensor_tag import SensorTag
+from pytz import UTC
```
EinarJohnsen: This grouping seems a bit weird

flikka: Oh, yes!

flikka:
Let's have one more feature! The project-name should be a part of the logged parameters!

flikka:
BTW tested this on the gordo-test-project, and it works so fine fine.

gordo_components/workflow/workflow_generator/resources/argo-workflow.yml.template
```diff
@@ -1326,6 +1337,8 @@ spec:
           template: model-builder
           arguments:
             parameters:
+              - name: project-name
+                value: "{{ project_name }}"
```
flikka, line 1341: I think you can just remove this (project-name) from the "input" stuff, and just inject it directly in the environment variable section?

gordo_components/workflow/workflow_generator/resources/argo-workflow.yml.template
```diff
@@ -631,19 +631,23 @@ spec:
...
+      - name: enable-remote-logging
     container:
       image: {{ docker_registry }}/{{ docker_repository }}/gordo-model-builder:{{gordo_version}}
       env:
       - name: OUTPUT_DIR
         value: "/gordo/models/{{project_name}}/{{project_version}}/{{'{{inputs.parameters.model-name}}'}}"
       - name: MODEL_REGISTER_DIR
         value: "/gordo/models/{{project_name}}/model_register"
+      - name: PROJECT_NAME
+        value: {{ '"{{inputs.parameters.project-name}}" '}}
```
flikka: couldn't this just be: `value: "{{  project_name }}"` directly, and not go via the input stuff?

flikka (changes requested):

gordo_components/workflow/workflow_generator/resources/argo-workflow.yml.template
ryanjdillon: The same could go for all of the input parameters then. Is there a reason it was done for the other input params?

flikka: Just for the record: @ryanjdillon and I spoke about this and clarified

milesgranger:
As this is a large PR with lots of changes around a single big feature I suggest you squash and merge. Which also implies you should move the conversion of pytest stuff to a different PR, because it's not relevant here.

flikka (approved):

flikka:
I think we have pondered this long enough now, and hopefully learned some lessons along the way :-) Feel free to merge if you feel ready @ryanjdillon I suggest we fire off a biggish test after merging, before we do any relase of this.

ryanjdillon:
:running_man:  :fire:

Issue #537, opened 2019-10-15
Client: Support public, authenticated endpoints


flikka:
Today the client doesn't support endpoints where authentication is needed. It should.

PR #574, opened 2019-10-24
Added optional requests.Session injection parameter

mrdeveloperdude:
To use gordo-client with azure authentication, we want to decorate the session with some auth headers.

By allowing the caller to inject their own session object, this can happen without gordo-client ever having to know or care about it.

### DIFF
--- a/gordo_components/client/client.py
+++ b/gordo_components/client/client.py
@@ -56,6 +56,7 @@ class Client:
         ignore_unhealthy_targets: bool = False,
         n_retries: int = 5,
         use_parquet: bool = False,
+        session: typing.Optional[requests.Session] = None,
     ):
         """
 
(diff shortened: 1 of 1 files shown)


ryanjdillon:
Sweet! This closes #537?

mrdeveloperdude:
Yes, but code to create the session is not includrd. I have that code, it
is 60 lines or so in a session factory. Not sure where to put that.

ryanjdillon:
The `io` module seems to make sense to me. I guess if it is up then folks could give their input. Thanks Lennart!

milesgranger:
Correct, subsequent requests are made using `aiohttp.ClientSession` Some work would need to be done to get the sessions unified. However, most of the parameters between [`aiohttp.ClientSession`](https://aiohttp.readthedocs.io/en/stable/client_reference.html#aiohttp.ClientSession) are compatible with [`request.Session`](https://requests.kennethreitz.org/en/master/user/advanced/#session-objects) so _maybe_ until we either get all http requests to use one or the other, we could take `session_kwargs` or something to that effect?

mrdeveloperdude:
Actually the session I used was none of those. please see
https://github.com/equinor/latigo/blob/master/app/latigo/gordo/auth.py

(warning, code is w. i. p.)

epa095:
Very nice! It is superseded by #587 though which stole your idea and ran with it! So you wont get this PR counting towards your hacktober t-shirt, but thanks for the idea ;-)

milesgranger:
As a courtesy, if you are looking for another as a replacement. there exists an `httpbin` fixture in `conftest.py` which can simply be deleted. :thought_balloon:

PR #576, opened 2019-10-24
Add more logging to server

### LINKED ISSUE #560: Server: Log a bit more
Today we are logging the time it takes to serve the whole request, but maybe we should log a few more things:
- The size of X and y (if passed)
- The time it takes to parse and handle X,y?
- The time it takes to load the model

EinarJohnsen:
Add logging on the time it takes to handle X, y and the time it takes to load the model. Closes #560

### DIFF
--- a/gordo_components/server/utils.py
+++ b/gordo_components/server/utils.py
@@ -5,6 +5,7 @@ import functools
 import os
 import io
 import dateutil
+import timeit
 from datetime import datetime
 from typing import Union, List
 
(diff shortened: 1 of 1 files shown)


gordo_components/server/utils.py
```diff
@@ -301,6 +302,12 @@ def wrapper_method(self, *args, **kwargs):
         # Assign X and y to the request's global context
         g.X, g.y = X, y
 
+        try:
+            logger.debug(f"Size of X: {X.size}, size of y: {y.size}")
+        except Exception:
```
epa095: If your except-claus assumes that a single exception-type is happening (Y is none), then I think you should only catch that exception which happends when `y` is none.

epa095 (approved):
LGTM

PR #587, opened 2019-10-25
Remove Client async to allow init session

### LINKED ISSUE #537: Client: Support public, authenticated endpoints

milesgranger:
Remove `async` stuff from Client, as it doesn't play well with external session objects and will be used in a multi-threaded environment anyway.

Spawns from and resolves #574 and will close #537


gordo_components/client/client.py
```diff
@@ -296,25 +298,16 @@ def predict(
...
-            (pr.name, pr.predictions, pr.error_messages) for pr in prediction_results
-        ]  # type: ignore
+        with ThreadPoolExecutor(max_workers=self.parallelism) as executor:
+            results = list()
+            jobs = executor.map(
+                lambda ep: self.predict_single_endpoint(ep, start, end), self.endpoints
+            )
+            for result in jobs:
+                results.append((result.name, result.predictions, result.error_messages))
+            return results
```
epa095: This can be a single list-comprehension, maybe nicer?

epa095: Also, should we then maybe return an iterator `(b for b in blabla)` vs `[b for b in blabla]`

gordo_components/client/client.py
```diff
@@ -13,7 +12,7 @@
 from datetime import datetime
 from concurrent.futures import ThreadPoolExecutor, TimeoutError as FutureTimeoutError
```
epa095: can we get a futuretimeouterror anymore?

gordo_components/client/forwarders.py
```diff
@@ -44,7 +42,7 @@ def __call__(
         endpoint: EndpointMetadata = None,
         metadata: dict = dict(),
         resampled_sensor_data: pd.DataFrame = None,
-    ) -> typing.Awaitable[None]:
+    ) -> None:
         ...
```
epa095: Change documentation away from " it is a coroutine which accepts predicti"....

gordo_components/client/forwarders.py
```diff
@@ -159,56 +155,47 @@ def __init__(
             if len(sub_df.columns) == len(endpoint.tag_list):
                 sub_df.columns = [tag.name for tag in endpoint.tag_list]
 
-            await self._write_to_influx_with_retries(sub_df, tags, top_lvl_name)
+            self._write_to_influx_with_retries(sub_df, tags, top_lvl_name)
 
-    async def _write_to_influx_with_retries(self, df, tags, measurement):
+    def _write_to_influx_with_retries(self, df, tags, measurement):
         """Async write data to influx with retries and exponential backof. Will sleep
```
epa095: not async

gordo_components/client/forwarders.py
```diff
@@ -159,56 +155,47 @@ def __init__(
...
-                    break
+                    msg = f"Failed to forward data to influx. Error: {exc}"
+                    logger.error(msg)
+            else:
+                break
 
-    async def send_sensor_data(self, sensors: pd.DataFrame):
+    def send_sensor_data(self, sensors: pd.DataFrame):
         """
         Async write sensor-data to influx
```
epa095: nope

epa095 (commented):
looking gooder and gooder

gordo_components/client/client.py
```diff
@@ -246,34 +249,29 @@ def download_model(self) -> typing.Dict[str, BaseEstimator]:
         """
         models = dict()
         for endpoint in self.endpoints:
-            resp = requests.get(f"{endpoint.endpoint}/download-model")
+            resp = self.session.get(f"{endpoint.endpoint}/download-model")
             if resp.ok:
                 models[endpoint.target_name] = serializer.loads(resp.content)
             else:
-                raise IOError(f"Failed to download model: '{resp.content}'")
+                raise IOError(f"Failed to download model: '{resp.content!r}'")
```
EinarJohnsen: Perhaps consider using `repr(resp.content)` instead of `!r`, as it is considered redundant according to `PEP 498`

gordo_components/client/client.py
```diff
@@ -246,34 +249,29 @@ def download_model(self) -> typing.Dict[str, BaseEstimator]:
...
         """
         metadata = dict()
         for endpoint in self.endpoints:
-            resp = requests.get(f"{endpoint.endpoint}/metadata")
+            resp = self.session.get(f"{endpoint.endpoint}/metadata")
             if resp.ok:
                 metadata[endpoint.target_name] = resp.json()
             else:
-                raise IOError(f"Failed to get metadata: '{resp.content}'")
+                raise IOError(f"Failed to get metadata: '{resp.content!r}'")
```
EinarJohnsen: ^

gordo_components/client/client.py
milesgranger: hmm, I was literally following mypy's suggestion

gordo_components/client/forwarders.py
```diff
@@ -159,56 +142,47 @@ def __init__(
             if len(sub_df.columns) == len(endpoint.tag_list):
                 sub_df.columns = [tag.name for tag in endpoint.tag_list]
 
-            await self._write_to_influx_with_retries(sub_df, tags, top_lvl_name)
+            self._write_to_influx_with_retries(sub_df, tags, top_lvl_name)
 
-    async def _write_to_influx_with_retries(self, df, tags, measurement):
-        """Async write data to influx with retries and exponential backof. Will sleep
+    def _write_to_influx_with_retries(self, df, tags, measurement):
+        """Write data to influx with retries and exponential backof. Will sleep
```
EinarJohnsen: Perhaps use a newline after `"""`

gordo_components/client/client.py
EinarJohnsen: I am not entirely sure, as I can not find any "best practices" on this exact case, but according to `PEP 498`: 
"The !s, !r, and !a conversions are not strictly required. Because arbitrary expressions are allowed inside the f-strings ...

... However, !s, !r, and !a are supported by this PEP in order to minimize the differences with str.format(). !s, !r, and !a are required in str.format() because it does not allow the execution of arbitrary expressions."

gordo_components/client/forwarders.py
```diff
@@ -159,56 +142,48 @@ def __init__(
...
+                )
+            except Exception as exc:
+                if current_attempt <= self.n_retries:
+                    # Sleep at most 5 min
+                    time_to_sleep = min(2 ** (current_attempt + 2), 300)
+                    logger.warning(
+                        f"Failed to forward data to influx on attempt "
+                        f"{current_attempt} out of {self.n_retries} attempts. "
+                        f"Error: {exc}."
+                        f"Sleeping {time_to_sleep} seconds and trying again"
```
EinarJohnsen: Punctuation would be nice, even though it was not there originally.

gordo_components/client/client.py
milesgranger: Ya I don't know, but I can't tell if it doesn't fall into that because it's interpolating strings with bytes. 
`On Python 3 '{}'.format(b'abc') produces "b'abc'"; use !r if this is a desired behavior`

but using `repr()` is the same for me, and that works too :man_shrugging:

EinarJohnsen (approved):
:bat: :jack_o_lantern: :ghost: :night_with_stars: LGTM

epa095 (approved):
:heart:

PR #588, opened 2019-10-28
Fixed typo logging -> logger

mrdeveloperdude:
Fixed tiny bug that produced a lot of log output that could not be quenched easily


epa095 (approved):
Awesome!

PR #590, opened 2019-10-28
`gordo-components workflow generate` updates

milesgranger:
Problems :1234: 
---
- We still had `cleanup_version` which was left over from gordo-infrastructure days
- Not all options were picked up from environment variables, and now that https://github.com/equinor/gordo-controller/pull/27 is in, we can set these param via the Gordo CRD config file in `spec.deploy-environment` mapping.
- The `kwargs` held much of the variables which `context` ends up having, might as well treat them the same. 

Solutions :checkered_flag: 
---
- Remove `cleanup_version` in favor of `gordo_version`
- Add `envvar` params to each option
- Rename `kwargs` to `context` and remove duplicate lines.

### DIFF
--- a/gordo_components/cli/workflow_generator.py
+++ b/gordo_components/cli/workflow_generator.py
@@ -44,11 +44,11 @@ def workflow_cli():
     envvar=f"{PREFIX}_OWNER_REFERENCES",
 )
 @click.option(
-    "--version",
+    "--gordo-version",
     type=str,
     default=wg._docker_friendly_version(__version__),
     help="Version of gordo to use, if different than this one",
-    envvar=f"{PREFIX}_VERSION",
+    envvar=f"{PREFIX}_GORDO_VERSION",
 )
 @click.option(
     "--project-name",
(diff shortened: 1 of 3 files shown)


gordo_components/cli/workflow_generator.py
```diff
@@ -182,12 +177,10 @@ def workflow_generator_cli(**kwargs: dict):
 
     # Context requiring pre-processing
     context["target_names"] = [machine.name for machine in config.machines]
+    context["owner_references"] = json.dumps(context["owner_references"])
```
epa095: Now it is defacto required, so then it should be required in the cli as well

gordo_components/cli/workflow_generator.py
```diff
@@ -196,17 +189,17 @@ def workflow_generator_cli(**kwargs: dict):
         template = wg.load_workflow_template(workflow_template)
 
     # Clear output file
-    if kwargs["output_file"]:
-        open(kwargs["output_file"], "w").close()  # type: ignore
-    for i in range(0, len(config.machines), kwargs["split_workflows"]):  # type: ignore
+    if context["output_file"]:
+        open(context["output_file"], "w").close()  # type: ignore
```
epa095, line 203: can this logic not be fixed with some fancy click-type on "output_file" maybe?

epa095: since you are fixing everyhting anywho I mean ;-)

epa095 (commented):
lovely

gordo_components/cli/workflow_generator.py
milesgranger: If it defaults to `None` then it gets dumped as `null`. Is that not ok?

gordo_components/cli/workflow_generator.py
epa095: Nha seems not really :-/

milesgranger: Did it despite your skepticism, lemme know what you think.

gordo_components/cli/workflow_generator.py
epa095: Ahh, tha tjo yeah, I dont know. Inside the workflow there is some tests that owner_reference is set or not, and I dont know what happends if `context["owner_references"]=None`.
```
  {% if owner_references is defined %}
  ownerReferences: {{owner_references}}
  {% endif %}
```
I suspect that it will then set ownerReferences: None, and I dont know what k8s thinks about that.

gordo_components/cli/workflow_generator.py
milesgranger: nvm.. i see now it wouldn't work. :sweat:

epa095 (approved):
LGTM

PR #591, opened 2019-10-29
Upgrade python to 3.7

### LINKED ISSUE #559: Try python 3.7
This involves:
- Changing all the docker images to be based on python 3.7, and test that stuff works
- Install gordo in an environment with 3.7 and check if things work. 
- Change python version in travis and circleci.

EinarJohnsen:
Fixes #559 if we decide to upgrade.


Dockerfile-Client
```diff
@@ -10,7 +10,7 @@ RUN rm -rf /code/dist \
     && python setup.py sdist \
     && mv /code/dist/$(ls /code/dist | head -1) /code/dist/gordo-components-packed.tar.gz
 
-FROM python:3.6.9-slim-stretch
+FROM python:3.7.4-slim-stretch
```
epa095: There is a 3.7.5, any reason not to go directly to it?

.circleci/config.yml
```diff
@@ -160,7 +160,7 @@ jobs:
           name: Set Python Version
           command: |
             pyenv versions
-            pyenv global 3.6.5
+            pyenv global 3.7.0
```
epa095, line 184: no 3.7.5?

EinarJohnsen:
Tested with `gordo-test-project` and it works as expected.

milesgranger:
Why is this on hold, I forget?

epa095:
Probably cause I am a 🐔, but I don't remember. Let's just do it!

EinarJohnsen:
Let's see if this stuff still passes :eyes:

epa095 (dismissed):
Now there is *yet another* place to set things, https://github.com/equinor/gordo-components/blob/master/.python-version

.python-version
```diff
@@ -1 +1 @@
-3.6.5
+3.7.5
```
milesgranger: Maybe you've [already seen](https://circleci.com/gh/equinor/gordo-components/14228?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), but CircleCI's pyenv only has `3.7.0` and I'm not sure if it's better to see how to change that fact or be 'good-enough' with `3.7`

EinarJohnsen: Yeah, I vaguely recall something about this. Thanks! :eyes:

EinarJohnsen:
:wave: :snake:-3.6

milesgranger:
Have you ran it through the test project for sanity purposes?

EinarJohnsen:
I have done this earlier, but since it's been laying around for a while I will run it again just to make sure :+1:

milesgranger (approved):
Giving approval on the assumption the test project goes through and no one else has any objections. @ryanjdillon @flikka

EinarJohnsen:
@milesgranger @flikka @ryanjdillon it has now passed in the test-project.

.circleci/config.yml
```diff
@@ -147,7 +147,7 @@ jobs:
   # Make the Sphinx documentation for the project; without errors or warnings.
   make-docs:
     docker:
-      - image: circleci/python:3.6
+      - image: circleci/python:3.7
```
flikka: Shouldn't we use 3.7.5 perhaps? Since that is what is the "latest" AND it is what the Dockers are using?

.python-version
```diff
@@ -1 +1 @@
-3.6.5
+3.7.0
```
flikka: 3.7.5?

Dockerfile-Watchman
```diff
@@ -0,0 +1,35 @@
+FROM python:3.7.5 as builder
```
flikka: This whole file is gone now, right? Probably some rebasing leftover.

flikka:
I suggest we do this, I had some questions on using 3.7.4 in stead of 3.7.0 - and the commit message needs to be re-worded, but let's move into the new world!

Dockerfile-Watchman
EinarJohnsen: Yes, this should be removed, I have probably added it with a mistake.

.python-version
EinarJohnsen: Circleci pyenv does not support 3.7.5, so this has to be 3.7.0 for now, as @milesgranger pointed out earlier.

.circleci/config.yml
EinarJohnsen: I'll test this.

epa095:
Why is the ".Python-version" file gone? Should it not work (as long as it is set to 3.7), and help dependabot be a bit less dumb?

EinarJohnsen:
@epa095 It was added to make dependabot less dumb and stop removing stuff, but as milg points out, it did not solve the issue. Therefore we removed it :man_shrugging:

PR #598, opened 2019-10-29
Move ml-server in workflow template, outside of influx block

### LINKED ISSUE #592: workflow template: Move ml-server task outside of influx if part
Today the ml-server is inside an if that only triggers if at least one model has influx enabled

flikka:
Close #592 

Tested with a project without influxes, that failed before. Now it doesn't.


epa095 (approved):
LGTM :heart_decoration:

milesgranger (approved):
:keyboard:

PR #599, opened 2019-10-30
Scale on scoring

epa095:


tests/gordo_components/model/test_utils.py
```diff
@@ -3,8 +3,34 @@
...
+    # make the features in y be in different scales
+    y = np.array([[1, 1], [2, 2], [3, 3], [4, 4], [5, 5]]) * [1, 100]
+
+    # With no scaler provided it is relevant which of the two series gets an 80% error
+    metric_func_noscaler = model_utils.metric_wrapper(mean_squared_error)
+
+    mse_feature_one_wrong = metric_func_noscaler(y, y * [0.8, 1])
+    mse_feature_two_wrong = metric_func_noscaler(y, y * [1, 0.8])
+
+    assert not math.isclose(mse_feature_one_wrong, mse_feature_two_wrong)
```
milesgranger: Very _very_ small note, that you could use `np.isclose(...)`, save you one line. :laughing:

gordo_components/cli/cli.py
```diff
@@ -93,7 +93,8 @@ def gordo():
 @click.option(
     "--evaluation-config",
     envvar="EVALUATION_CONFIG",
-    default='{"cv_mode": "full_build"}',
+    default='{"cv_mode": "full_build", '
+    '"scoring_scaler": "sklearn.preprocessing.RobustScaler"}',
```
milesgranger: Just noticing this now, but should be rather do:
```python
yaml.safe_dump({"cv_mode": "full_build", "scoring_scaler": "sklearn.preprocessing.RobustScaler"})
```
instead of writing it out as a string, which is less likely to catch syntax errors?

gordo_components/builder/build_model.py
```diff
@@ -68,6 +71,14 @@ def build_model(
                 Example::
 
                     {"cv_mode": "cross_val_only"}
+            - scoring_scaler: Optional[str]
```
milesgranger, line 71: think this should go above the example, and maybe update the example with this as well?

milesgranger (commented):
Just small comments, looks goodie good!

milesgranger (approved):
:dollar:

milesgranger (approved):
Refreshing my approval, _2 days ago_. Nice work!

PR #601, opened 2019-10-30
Add CHANGELOG.md

milesgranger:
Problem :card_file_box: 
---
We don't have a [CHANGELOG](https://keepachangelog.com/en/1.0.0/), and our release notes are not really consistent.

Solution :computer: 
---
Maybe we should?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1507 @@
+
+<a name="0.38.0"></a>
+## [0.38.0](https://github.com/equinor/gordo-components/compare/0.38.0-rc1...0.38.0) (2019-10-29)
+
+
+<a name="0.38.0-rc1"></a>
+## [0.38.0-rc1](https://github.com/equinor/gordo-components/compare/0.37.0...0.38.0-rc1) (2019-10-29)
+
+### Accept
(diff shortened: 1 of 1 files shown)


epa095:
IDK, should we? How do we do it, do we tag a release in github, and then make a commit right after with the updated changelog? And it is clearly autogenerated, almost like `git log` but with everything under a header with the first word of the commit message :-p

milesgranger:
This was just 'getting us up to speed' categorizing them by the first word in the commit message. Which we _could_ continue to do, categorizing them like `Add: blah blah` or `Breaking: blah`.. I honestly don't know if I like or dislike that approach.

Alternatively, with each PR/commit we can manually add to the appropriate field in the section [unreleased] and then, yes, before actually releasing, we update this field to represent the actual release.

Additionally, we can get extra fancy by using git pre-commit hooks to run `gitchng-log` to update the CHANGELOG accordingly. But that seems difficult to integrate new contributors, instead of saying with each PR they should update the appropriate field in the CHANGELOG.

This was just something @flikka and I were discussing.. so I'm not married to the idea, it was just easy enough to demo.

PR #602, opened 2019-10-31
Consolidate CI tests into ~4-5min run times

milesgranger:


epa095 (approved):
Awesome! But maybe change commit message so it is clear it is only a CI change. As it stands it can be misunderstood as you did something with the actual tests

PR #605, opened 2019-10-31
WIP: test skipping install step if cache is set

epa095:


.circleci/config.yml
```diff
@@ -175,23 +175,27 @@ jobs:
...
           name: Virtual Environment
           command: |
+            [ -d "/home/circleci/code/venv" ]
             python -m venv ~/code/venv
             echo "source ~/code/venv/bin/activate" >> $BASH_ENV
       - run: pip install --upgrade pip
       - run:
           name: Install Requirements
           command: |
+            [ -d "/home/circleci/code/venv" ]
```
milesgranger, line 192: Wouldn't this always exit, because `Virtual Environment` would actually create this, but it would still be "empty"?

epa095: neinei, and I realise this was stupid anywho

PR #609, opened 2019-11-01
Remove custom keras model serializations

milesgranger:
Problem-ish :spaghetti: 
---
Previously (our) Keras models couldn't be pickled. We worked around this by creating a directory tree of individually pickled steps in the model/pipeline, and any keras model's encountred would be saved and loaded using `.save_to_dir()`/`load_from_dir()` methods. 

This introduced quite complex saving logic and thus code and tests

Solution? :hamburger: 
---
Implement custom `__getstate__` and `__setstate__` methods for our models, so we can "just" pickle the entire pipeline/model directly.

### DIFF
--- a/gordo_components/model/base.py
+++ b/gordo_components/model/base.py
@@ -18,16 +18,6 @@ class GordoBase(abc.ABC):
         """Return a dict containing all parameters used to initialized object"""
         ...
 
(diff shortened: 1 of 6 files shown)


epa095 (approved):
lovely

PR #611, opened 2019-11-04
Remove bad chars in Statfjord path

EinarJohnsen:
Remove invisible chars in the path as they are removed from the datalake. Tested and works like a charm.


milesgranger (approved):
Good eye. ;-)

PR #612, opened 2019-11-04
Fix parallel testing

milesgranger:
Add `--ignore benchmarks` to the `test`, so it will run tests, ignoring benchmarks and docker tests.

Side note: checked to see what would happen in CircleCI if we ran two tests, one with the docker tests and the other without (using parallelism) but that job still took >10mins.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,7 +1,7 @@
 [aliases]
 
 # All tests in parallel, minus any requiring docker
-test             = pytest --addopts "-n auto -m 'not dockertest'"
+test             = pytest --addopts "-n auto -m 'not dockertest' --ignore benchmarks"
 
 # Test _everything_
 testall          = pytest --addopts "--ignore benchmarks"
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -1,10 +1,10 @@
 [aliases]
 
 # All tests in parallel, minus any requiring docker
-test             = pytest --addopts "-n auto -m 'not dockertest'"
+test             = pytest --addopts "-n auto --ignore benchmarks -m 'not dockertest'"
 
 # Test _everything_
-testall          = pytest --addopts "--ignore benchmarks"
+testall          = pytest --addopts "-n auto --ignore benchmarks -m 'not dockertest'"
```
epa095: `testall`  seems suspiciously similar to `test` now, and neither of them tests all...

milesgranger: WIP

epa095 (commented):
LGTM, although the commit message `Fix parallel testing` is pretty non-informative

epa095 (approved):

PR #613, opened 2019-11-04
Add missing limits to postgres, grafana and misc workers

epa095:
There were a few things we deploy which did not contain resource limits, and its quite possible that in the future there will be default limits imposed, so we should set them ourself instead so we are in control. 

Tested and it works


milesgranger (approved):
:zap:

PR #614, opened 2019-11-04
Rename Dockerfile-Argo -> Dockerfile-GordoDeploy

milesgranger:
Dockerfile-Argo isn't reflective to what it is/does


epa095 (approved):
LGTM

PR #622, opened 2019-11-05
Support configuring client Session via CLI

### LINKED ISSUE #593: Client-CLI: Support public, authenticated endpoints
A bit like #537, but different.

milesgranger:
Will close #593

### DIFF
--- a/gordo_components/cli/client.py
+++ b/gordo_components/cli/client.py
@@ -8,6 +8,7 @@ from datetime import datetime
 from pprint import pprint
 
 import click
+import yaml
 import pandas as pd  # noqa
 from requests import Session
 
@@ -18,7 +19,6 @@ from gordo_components.cli.custom_types import (
     IsoFormatDateTime,
     DataProviderParam,
     key_value_par,
-    key_value_lit_eval,
 )
 from gordo_components.client.forwarders import ForwardPredictionsIntoInflux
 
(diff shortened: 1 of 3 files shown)


gordo_components/cli/custom_types.py
```diff
@@ -70,3 +71,12 @@ def key_value_par(val) -> typing.Tuple[str, str]:
     Helpder for CLI input of 'key,val'
     """
     return val.split(",")
+
+
+def key_value_lit_eval(val) -> typing.Tuple[str, str]:
+    """
+    Helpder for CLI input of key value pair where the value should be
```
epa095: Helpder

epa095 (approved):
LGTM

gordo_components/cli/client.py
```diff
@@ -38,11 +40,23 @@
...
+    help="Config json/yaml to set on the requests.Session object. Useful when needing to supply"
+    "authentication parameters such as header keys. ie. --session-config {'headers': {'API-KEY': 'foo-bar'}}",
+)
 @click.pass_context
 def client(ctx: click.Context, *args, **kwargs):
     """
     Entry sub-command for client related activities
     """
+    # Setup the session with any attributes set by the user
+    session = Session()
```
epa095: Should this maybe just happen if session-config is actually something? I.e. that it is not in general the CLI's task to start the session, only if we actually need it?

PR #626, opened 2019-11-07
Client.predict(..) optionally take specific endpoints

### LINKED ISSUE #247: Client: Add ability to have several targes
Today the client can either work towards a single target or all of them. It would be reasonable for it to be able to take in a list of targets. If `--ignore-unhealthy-targets` is toggled and one of the targets in the list is down then it should still succeed. If `--ignore-unhealthy-targets` is not toggled then it should fail if a single one **of the targets** are unhealthy.

milesgranger:
Problem :egg: 
---
`Client.predict` runs predictions for all acquired endpoints.

Solution :fried_egg: 
---
Optionally take a list of specific endpoints to run predictions for.

Will close #247

### DIFF
--- a/gordo_components/client/client.py
+++ b/gordo_components/client/client.py
@@ -8,7 +8,7 @@ import itertools
 import typing
 from time import sleep
 from threading import Lock
-from typing import Dict, Any, List, Callable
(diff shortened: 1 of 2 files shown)


epa095 (commented):
LGTM

PR #627, opened 2019-11-07
Implement an AnomalyDetectorBase ABC

### LINKED ISSUE #393: Make an AnomalyDetectorBase ABC
#368 abstracts the previous anomaly calculations into an `DiffBasedAnomalyDetector` which basically implements a `.anomaly` method.

To lay the foundation for implementing different types of anomaly detectors, we should create an ABC for such models.

Should update `gordo_components/server/views/anomaly.py ` to check for `isinstance` rather than catching the `AttributeError`

milesgranger:
Will close #393

### DIFF
--- a/gordo_components/model/anomaly/base.py
+++ b/gordo_components/model/anomaly/base.py
@@ -1,0 +1,18 @@
+import abc
+import pandas as pd
+from typing import Optional
+from datetime import timedelta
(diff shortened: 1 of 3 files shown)


EinarJohnsen (approved):
:spades: Cool stuff :clubs:

PR #630, opened 2019-11-07
Add invalid model config

### LINKED ISSUE #624: Update ValidateModel to do some model validation
Today, we just check that it is a dict or a string. We could do more! 

@milesgranger probably have good input on this! 

https://github.com/equinor/gordo-components/blob/b65a54e94b8e912063a57ff6a607e696147c8233/gordo_components/workflow/config_elements/validators.py#L80

EinarJohnsen:
I think this is ok now, will remove WIP. We could rewrite the whole unittest stuff, but I think it is better as a separate PR then. Fixes #624

### DIFF
--- a/tests/gordo_components/workflow/test_config_elements.py
+++ b/tests/gordo_components/workflow/test_config_elements.py
@@ -4,15 +4,16 @@ import ast
 import json
 import logging
 import unittest
-
+from io import StringIO
 from datetime import datetime, timezone
 
 import yaml
+
 from gordo_components.workflow.config_elements.dataset import Dataset
 from gordo_components.workflow.config_elements.machine import Machine
 from gordo_components.workflow.config_elements.normalized_config import NormalizedConfig
 from gordo_components.workflow.workflow_generator.workflow_generator import (
-    _timestamp_constructor,
+    get_dict_from_yaml,
 )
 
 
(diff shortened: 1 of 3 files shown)


tests/gordo_components/workflow/test_config_elements.py
```diff
@@ -182,3 +182,41 @@ def test_from_config(self):
...
+                scoring_scaler: Null
+            metadata:
+              id: special-id
+        """
+        yaml.FullLoader.add_constructor(
+            tag="tag:yaml.org,2002:timestamp", constructor=_timestamp_constructor
+        )
+        element = yaml.load(element_str, Loader=yaml.FullLoader)
+
+        print(element)
```
epa095: I know, I know, it is still WIP. But I like it.

tests/gordo_components/workflow/test_config_elements.py
```diff
@@ -182,3 +182,41 @@ def test_from_config(self):
...
+                step:
+                  - sklearn.preprocessing.data.MinMaxScaler
+                  - gordo_components.model.models.KerasAutoEncoder:
+                      kind: feedforward_hourglass
+            evaluation:
+                scoring_scaler: Null
+            metadata:
+              id: special-id
+        """
+        yaml.FullLoader.add_constructor(
```
milesgranger: Maybe you could use `workflow_generator.get_dict_from_yaml` here?

EinarJohnsen: @milesgranger Then I belive the function in `workflow_generator.get_dict_from_yaml` has to be rewritten. It seems that it only will work for `io.StringIO`, as a string will fail on `if hasattr(config_file, "read")`. So it should probably be rewritten anyway since it is supposed to accept `str`.

milesgranger (approved):
Awesome, possum! :mouse:

PR #631, opened 2019-11-07
Accept metrics list in evaluation config key

### LINKED ISSUE #242: Modelbuilder: Customize scoring

milesgranger:
Problem :worried: 
---
We've added the ability to score the model on multiple scoring functions, but those are fixed to 
- r2_score
- explained_variance_score
- mean_squared_error
- mean_absolute_error

Solution :clown_face: 
---
Evaluation config can now take any scorer name in [sklearn.metrics](https://scikit-learn.org/stable/modules/classes.html#sklearn-metrics-metrics)

```yaml
evaluation:
   metrics:
      - r2_score
      - max_error
```

Will close #242

### DIFF
--- a/gordo_components/builder/build_model.py
+++ b/gordo_components/builder/build_model.py
@@ -25,6 +25,7 @@ from gordo_components.dataset.dataset import _get_dataset
 from gordo_components.dataset.base import GordoBaseDataset
 from gordo_components.model.base import GordoBase
 from gordo_components.model.utils import metric_wrapper
+from gordo_components.workflow.config_elements.normalized_config import NormalizedConfig
 
 logger = logging.getLogger(__name__)
 
@@ -609,10 +610,7 @@ def metrics_from_list(metric_list: Optional[List[str]] = None) -> List[Callable]
     AttributeError
         If the metric name is not found in `sklearn.metrics`
     """
-    metric_list = metric_list or [
-        "explained_variance_score",
-        "r2_score",
-        "mean_squared_error",
-        "mean_absolute_error",
-    ]
+    metric_list = (
+        metric_list or NormalizedConfig.DEFAULT_CONFIG_GLOBALS["evaluation"]["metrics"]
+    )
     return [getattr(metrics, metric) for metric in metric_list]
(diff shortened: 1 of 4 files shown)


EinarJohnsen (approved):
:clown_face: Nice work, Sir! :tophat:

gordo_components/builder/build_model.py
```diff
@@ -593,3 +586,31 @@ def provide_saved_model(
         logger.info(f"Writing model-location to model registry")
         disk_registry.write_key(model_register_dir, cache_key, model_location)
     return model_location
+
+
+def metrics_from_list(metric_list: Optional[List[str]] = None) -> List[Callable]:
```
epa095, line 591: I know this is accepted already, so sorry for my nitpicking, but does not make sense to *add* the posibility to specify the whole function, like `sklearn.metrics.r2_score` ? The argument for allowing this is that pretty much everywhere else we are refering to functions we are refering to them with their full path, so it is not unlikely that a user would use the whole path (I certainly would). 
I think it is OK to also do what is done here, allowing the use of all metrics in `sklearn.metrics` without the full path. It can be a bit confusing after some time knowing where that is allowed and where it is not, but I think the convenience outweights that problem. But if we could say that "full paths are always allowed" that makes it a bit nicer at least(?)

milesgranger: Good points.

I'm not sure though what other functions would be supported, ever. As we wrap specific scoring functions with expected signatures. I will happily update this function to do as you say otherwise. It just appears that the need for referring to other scoring functions would become quite rare indeed; as their signatures would need to be the same as what our wrapper expects, no?

milesgranger: We also don't specify the full path to the dataset type. We would update that as well?

epa095: > We also don't specify the full path to the dataset type. We would update that as well?

True, I meant that we specify full paths when we use types not defined by us. I don't think we should update the dataset type to full path, we reserve the right to move it wherever we want without that leaking out in the config file.

I don't know if we will ever refer to functions outside sklearn.metrics, but that is not actually the primary reason I want it. It is mostly a "consistency / aesthetic" thing, that everywhere else I refer to sklearn stuff I write `sklearn.preprocessing...`, and it is a bit weird that this particular place that is not possible

PR #632, opened 2019-11-08
Support setting random seed in evaluation key

### LINKED ISSUE #452: [Feature request]:  gordo exposes fixing the seed to the user
At the moment gordo does not support fixing the seed in order to reproduce results.

Example:
This code does not fix the seed. 

`numpy.random.seed(42)`
`tensorflow.set_random_seed(42)`

If you re train a gordo model, it will produce different results.

The ability to reproduce results is important to study the influence of the hyperparameters on a model.

milesgranger:
Problem :thinking: 
---
_Sometimes_ it is appropriate to set the seed when training/evaluating models.

Solution :open_hands: 
---
Add ability to do the following in a config in the `evaluation` key:
```yaml
evalutation:
  seed: 1234
```

Will close #452

### DIFF
--- a/gordo_components/builder/build_model.py
+++ b/gordo_components/builder/build_model.py
@@ -6,11 +6,13 @@ import json
 import logging
 import os
 import time
+import random
 from pathlib import Path
 from typing import Union, Optional, Dict, Any, Tuple
 
 import pandas as pd
 import numpy as np
+import tensorflow as tf
 
 from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.metrics import (
@@ -88,6 +90,13 @@ def build_model(
     -------
         Tuple[Optional[sklearn.base.BaseEstimator], dict]
     """
+    if evaluation_config.get("seed"):
+        seed = evaluation_config["seed"]
+        logger.info(f"Setting random seed: '{seed}'")
+        tf.random.set_seed(seed)
+        np.random.seed(seed)
+        random.seed(seed)
+
     # Get the dataset from config
     logger.debug(f"Initializing Dataset with config {data_config}")
 
--- a/tests/gordo_components/builder/test_builder.py
+++ b/tests/gordo_components/builder/test_builder.py
@@ -627,17 +627,26 @@ def test_model_builder_cv_scores_only(should_be_equal: bool, evaluation_config: 
 
 
 @pytest.mark.parametrize("seed", (None, 1234))
-def test_setting_seed(seed):
+@pytest.mark.parametrize(
+    "model_config",
+    (
+        {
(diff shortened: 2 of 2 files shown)


gordo_components/builder/build_model.py
```diff
@@ -88,6 +90,13 @@ def build_model(
     -------
         Tuple[Optional[sklearn.base.BaseEstimator], dict]
     """
+    if evaluation_config.get("seed"):
+        seed = evaluation_config["seed"]
+        logger.info(f"Setting random seed: '{seed}'")
+        tf.random.set_seed(seed)
+        np.random.seed(seed)
+        random.seed(seed)
```
ryanjdillon, line 98: Would it be worth testing that these all infact got set to `seed`?  Maybe if these libs change how things get set, and these calls don't act as planned in a way that wouldn't affect the `df1==df2` test but could affect some other pipeline not represented in the tests.

ryanjdillon:
I don't recall what was landed on, or perhaps not present with the final decision of things, with regards to how exposed the seed setting should be. Otherwise, this LGTM! :icecream:

ryanjdillon (approved):

gordo_components/builder/build_model.py
```diff
@@ -88,6 +90,13 @@ def build_model(
     -------
         Tuple[Optional[sklearn.base.BaseEstimator], dict]
     """
+    if evaluation_config.get("seed"):
```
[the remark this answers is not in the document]
milesgranger: Same for me :man_shrugging:

gordo_components/builder/build_model.py
milesgranger: It would be _very_ strange indeed if the api changed in such a way the seed was _not_ set and yet the test as it is, still passed. 

anyhow, according to _some guy_, it doesn't seem (that) possible. 
> [The short answer is that you simply can't (at least not in general).](https://stackoverflow.com/questions/32172054/how-can-i-retrieve-the-current-seed-of-numpys-random-number-generator)

But manually looking is appears one could (for numpy anyway) index into the state `np.random.get_state()[1][0]` but I don't think it's worth it.

EinarJohnsen:
Perhaps a bit late, but should this also be tested with a NN Keras model? I've heard there is problems with neural networks earlier on, not sure if it is fixed after TF 2.0. @milesgranger

milesgranger:
Good point, I added it and seems a-ok. So TF 2.0 ftw :checkered_flag:

gordo_components/builder/build_model.py
ryanjdillon: Right on :+1:  Sounds to good me. The comments in *some dude*'s answer got *heated*.

PR #633, opened 2019-11-11
Take paths to metrics as well as function names

milesgranger:
In relation to: https://github.com/equinor/gordo-components/pull/631#discussion_r344616190

Should take the full paths to the function names, rather than assuming they're in `sklearn.metrics`


epa095 (commented):
Just to make it clear, I have no problems with supporting giving just the short-names **as well**. Do you have a preference for allowing **only** the full paths?

milesgranger:
Nope, nope!

epa095 (commented):
LGTM

PR #638, opened 2019-11-11
Workflow exponential retries

epa095:
This replaces the argo workflow `resource` template type with a script which does the same, but with possibility for exponential backoff. By default it retries 6 times starting with 15 seconds (15, 30, 60, 120, 240, 480), for a total of 945s (15.75min). This pod is again retried 4 times (because what if it just gets killed at some point, it must be also be retried :-/), for a max wait of ~1 hour. 

Parameters can be changed. And the whole idea can be ditched. Who knows?


milesgranger (approved):

flikka (approved):
So flexible and beautiful 👸

PR #639, opened 2019-11-11
Docs: Update examples and add AnomalyDetectorBase

milesgranger:
Problem :black_circle: 
---
Our `Endpoints` documentation link has fallen out of date, saying we support `GET` and the format of the example responses is slightly different.

Solution :red_circle: 
---
Update examples with use of our server utility functions, update our formatting, and add the `AnomalyDetectorBase` to the documentation as well.


epa095 (approved):
Veldig bra

PR #640, opened 2019-11-11
Piptools chain req and test dev

epa095:
This adds requirements.txt as a check in test_requirements.in.
This means that if there are incompatibilities pip-tools will give an error,
instead of allowing us to have different environments in test/prod.


milesgranger (approved):
Cool! :1st_place_medal:

PR #643, opened 2019-11-13
Refactor logic in build_model

ryanjdillon:
This PR includes some slight re-factoring of build model to simplify the logic and reduce lines.

### DIFF
--- a/gordo_components/builder/build_model.py
+++ b/gordo_components/builder/build_model.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+
 import datetime
 from distutils.dir_util import copy_tree
 import hashlib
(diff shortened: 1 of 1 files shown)


gordo_components/builder/build_model.py
```diff
@@ -192,6 +189,7 @@ def build_model(
     }
```
EinarJohnsen, line 189: Perhaps remove this line if there is to be a newline between metadata and the return statement :)

EinarJohnsen (approved):
LGTM with minor comment :poodle:

gordo_components/builder/build_model.py
ryanjdillon: Sweet. Thanks Einar!

PR #644, opened 2019-11-13
Fix deprecation warning h5py

EinarJohnsen:
H5py will change default file mode to 'r', so the mode is now explicitly set to avoid any future surprises.


epa095 (approved):
LGTM when tests pass

PR #646, opened 2019-11-13
Fix assert of out_file in test_workflow_generator

ryanjdillon:
This tiny guy should fix the random testing errors we are getting. 

*tiny guy = using the correct variable for the workflow generator output file


EinarJohnsen (approved):
:trophy:

tests/gordo_components/workflow/test_workflow_generator/test_workflow_generator.py
```diff
@@ -381,7 +381,7 @@ def test_main_tag_list(output_to_file, path_to_config_files, tmp_dir):
     assert result.exit_code == 0
 
     if output_to_file:
-        assert os.path.isfile(output_to_file)
+        assert os.path.isfile(out_file)
```
epa095, line 384: Good catch!

PR #647, opened 2019-11-13
Update tmp_dir fixture to yield teardown pattern

ryanjdillon:
# Problem :beer: 
The fixture for creating a temporary directory leaves behind some tmp-crud at times.

# Solution :ireland: 
This uses a context manager with a yield to make sure its cleaned up aftwerwards.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -46,9 +46,10 @@ def sensors():
 @pytest.fixture
 def tmp_dir():
     """
-    Fixture: Temporary directory
+    Fixture: Temporary directory, removed after test completion
     """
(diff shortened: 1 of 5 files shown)


milesgranger (commented):
Looks good, but the commit does more than what it says.

flikka:
> Looks good, but the commit does more than what it says.

But not that much more right? Some reformatting and cleaning? Any reason not to approve @milesgranger?

milesgranger (approved):

milesgranger:
There you go @flikka

epa095:
After a rebase on master I think the tests should pass again  @ryanjdillon

PR #648, opened 2019-11-13
Increase test influx timeout 30s->120s

epa095:
Could this be why circleci has problems starting influx?


milesgranger (approved):

flikka (approved):
Such a beautiful castle we are building! 🏰

PR #653, opened 2019-11-13
Change matplotlib log level to info

### LINKED ISSUE #651: Remove matplotlib debug logg messages
We are getting a bunch of debug messages from matplotlib now. That is annoying. Maybe we just set the loglevel of the matplotlib logger to something else?

EinarJohnsen:
I think this is ok for now, unless you think it is better to swap everything to INFO, and specify what should be debug, but it will be either or anyway. Fixes #651


milesgranger (approved):
:whale2:  Good enough for me! :whale:

epa095 (approved):
Lgtm

PR #654, opened 2019-11-14
Workflow: Move model notification up to top-level

epa095:
Simplifies the workflow by removing nested templates. I.e. instead of having
a template consisting of building-then-post-service, we add the building
and service-posting as top-level tasks in do-all, and add the builder as a
dependency of the service.

Needed beause of https://github.com/argoproj/argo/issues/1713


flikka (approved):
JEPP

PR #660, opened 2019-11-15
Add junit report generation to pytest

epa095:
This saves test-output in junitxml format in the folder ./junit. This
is then read by circleci to give nice graphs.


milesgranger (approved):
More addons :tada:

PR #668, opened 2019-11-18
Model builder: Add different exit codes for different exceptions

### LINKED ISSUE #663: Return different exit codes for different errors?
Today we return exitcode 1 on all errors, e.g. on `FileNotFoundError`. We could return different error codes for different errors. These are displayed when you do `argo get ...`, so then you can quickly see what went wrong (when you know the codes), and we could maybe do something programatically for them in the future.

epa095:
For now we only specify the following mapping, every other exception is
mapped to 1

PermissionError: 20,
FileNotFoundError: 30

This closes #663

### DIFF
--- a/gordo_components/cli/cli.py
+++ b/gordo_components/cli/cli.py
@@ -5,11 +5,15 @@ CLI interfaces
 """
 
 import logging
+import sys
+import traceback
+
 from gunicorn.glogging import Logger
 
 import jinja2
 import yaml
 import click
+from typing import Dict, Type
 
 from gordo_components.builder.build_model import (
     provide_saved_model,
@@ -28,6 +32,10 @@ from gordo_components.cli.workflow_generator import workflow_cli
 from gordo_components.cli.client import client as gordo_client
 from gordo_components.cli.custom_types import key_value_par, HostIP, DataProviderParam
 
+EXCEPTION_TO_EXITCODE: Dict[Type[Exception], int] = {
+    PermissionError: 20,
+    FileNotFoundError: 30,
+}
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


gordo_components/cli/cli.py
```diff
@@ -181,62 +189,71 @@ def build(
...
+                )
+                saved_metadata = load_metadata(model_location)
+                all_scores = get_all_score_strings(saved_metadata)
+
+            for score in all_scores:
+                print(score)
+    except Exception as e:
+        exit_code = EXCEPTION_TO_EXITCODE.get(e.__class__, 1)
+        traceback.print_exc()
+        sys.exit(exit_code)
```
milesgranger, line 256: Small thing: add an `else` ?

tests/gordo_components/cli/test_cli.py
```diff
@@ -224,6 +228,36 @@ def test_expand_model_complains_on_missing_vars(self):
...
+            MODEL_NAME="model-name",
+            OUTPUT_DIR=tmp_dir,
+            DATA_CONFIG=DATA_CONFIG,
+            MODEL_CONFIG=json.dumps(MODEL_CONFIG_WITH_PREDICT),
+        ):
+            result = runner.invoke(
+                cli.gordo,
+                ["build", '--evaluation-config={"cv_mode": "cross_val_only"}'],
+            )
+            assert result.exit_code == exit_code
```
milesgranger, line 258: Tasty.

milesgranger (approved):
At some point, maybe we should re-evaluate this CLI's build function.. but otherwise looks acceptable to me.

gordo_components/cli/cli.py
epa095: Containing what? The return?

milesgranger: yespi. Just to emphasize that two possible outcomes can happen in in this `try` block. 
I guess this is a style choice if you prefer implicit or explicit forms similar to:
```python
if foo:
    return foo
return bar
```
or 
```python
if foo:
    return foo
else:
    return bar
```

milesgranger: It's a free country, though.. :norway:

PR #672, opened 2019-11-18
Add Loader to yaml.load

EinarJohnsen:
Add Loader to yaml.load to avoid warnings in tests, and also make it safer :lock:


epa095 (approved):
LGTM!

milesgranger (approved):
Sublime. :green_apple:

Issue #650, opened 2019-11-13
Gordo production (aka prediction) output


henrikny:
### Model output

Once in production, the following values and variables should be returned:
(n observations, m input variables)

- [x] `tag_pred`: Model predictions original scale (n x m).
- [x] `tag_scaled_res`: Residuals based on scaled inputs and predictions (n x m).
- [x] `tag_abs_res`: Absolute residuals based on **unscaled** inputs and predictions (n x m).
- [x] `model_mse`: Based on the squared residuals from the **scaled** inputs and predictions, calculate the _average over all variables_ for _each point in time_ (n x 1).
- [x] Current _total anomaly_ (scalar) based on the Frobenius (`numpy.linalg.norm`) norm can be dropped.

milesgranger:
It seems like almost all of this, or at least certainly the "model output", can be implemented as a separate [`AnomalyDetectorBase`](https://gordo-components.readthedocs.io/en/latest/components/model/anomaly.html#gordo_components.model.anomaly.diff.DiffBasedAnomalyDetector) model or updating the current [`DiffBasedAnomalyDetector`](https://gordo-components.readthedocs.io/en/latest/components/model/anomaly.html#gordo_components.model.anomaly.diff.DiffBasedAnomalyDetector)

flikka:
So @henrikny, @milesgranger have worked on a PR that solves a lot of the things you mention here: https://github.com/equinor/gordo-components/pull/673

Some issues / questions:
* The output today + this PR, from the "anomaly" endpoint is:
    * The (original) model input (aka the X values, in the tags list), called model-input, for each tag
    * The calculated output (aka the Yhat values, from the target_tag_list list) in the original scale, called model-output, for each output tag.
    * tag-anomaly values for each output-tag. These are _normalized_/scaled values based on the DiffBasedAnomalyDetector. The reason for scaling is that they are then more easily comparable.
    * total-anomaly, one value per timestamp. This is just all the tag-anomalies, also on scaled data.
    * (new): anomaly-confidence: one value per output-tag, where the 6 point min sliding window was used to device a threshold (not reported) and then based on these thresholds, we produce the anomaly-confidence. The `anomaly-confidence` is just the a tag's anomaly, divided by the threshold devised.
[...]

PR #673, opened 2019-11-18
Support anomaly factor in DiffBasedAnomalyDetector

### LINKED ISSUE #195: Add "anomaly" concept in gordo?

milesgranger:
Problem :worried: 
---
It seems decent to return a confidence of how certain we are of a given sample being anomalous, but we deal in absolutes; concrete calculations of the difference in predicted vs expected.

Approach :slightly_smiling_face: 
---
Use the currently accepted idea that we can smooth differences during the cross validation and use the intermediate model's predictions for each fold to come up with thresholds for what we might determine is truly anomalous.

This is then applied in the `.anomaly` method which attaches additional `anomaly-confidence` scores per tag/feature using the aforementioned threshold values from the cross validation step.

Its output now looks like this:

```
   start   end model-input                     model-output                     tag-anomaly                     total-anomaly anomaly-confidence                    
                         0         1         2            0         1         2           0         1         2                                0         1         2
0   None  None    0.616486  0.503274  0.976017     0.029085 -0.130138  0.149330    1.386507  1.249649  1.950956      2.700047           2.658618  3.912564  3.834872
1   None  None    0.236190  0.816305  0.148481     0.015155 -0.085456  0.105218    0.521732  1.779074  0.102099      1.856807           1.000417  5.570155  0.200689
2   None  None    0.660414  0.715988  0.714318     0.027229 -0.111745  0.128906    1.494574  1.633024  1.381555      2.609447           2.865836  5.112885  2.715636
3   None  None    0.783051  0.359459  0.709639     0.024508 -0.044006  0.051228    1.790473  0.795991  1.553829      2.500756           3.433220  2.492192  3.054264
4   None  None    0.330104  0.748338  0.667890     0.027701 -0.153407  0.177224    0.713796  1.779042  1.157957      2.239500           1.368699  5.570057  2.276122

```

Given that thresholds were calculated as:
```
0    0.521514
1    0.319394
2    0.508741
```

---

Part of #650 and will close #195 (?)

### DIFF
--- a/gordo_components/model/anomaly/diff.py
+++ b/gordo_components/model/anomaly/diff.py
@@ -247,14 +247,11 @@ class DiffBasedAnomalyDetector(AnomalyDetectorBase):
 
         # If we have `thresholds_` values, then we can calculate anomaly confidence
         if hasattr(self, "thresholds_"):
-            y = y.values if hasattr(y, "values") else y
-            model_output = data["model-output"].values
-            abs_diff = np.abs(model_output - y[-len(model_output) :])
-            confidence_percentage = np.clip(abs_diff / self.thresholds_.values, 0, 1)
+            confidence = np.clip(tag_anomaly.values / self.thresholds_.values, 0, 1)
 
             # Dataframe of % abs_diff is of the thresholds
             anomaly_confidence_scores = pd.DataFrame(
-                confidence_percentage,
+                confidence,
                 columns=pd.MultiIndex.from_product(
                     (("anomaly-confidence",), data["model-output"].columns)
                 ),
--- a/gordo_components/model/anomaly/diff.py
+++ b/gordo_components/model/anomaly/diff.py
@@ -21,6 +21,7 @@ class DiffBasedAnomalyDetector(AnomalyDetectorBase):
         self,
         base_estimator: BaseEstimator = KerasAutoEncoder(kind="feedforward_hourglass"),
         scaler: TransformerMixin = RobustScaler(),
+        require_thresholds: bool = True,
     ):
         """
         Classifier which wraps a ``base_estimator`` and provides a diff error
@@ -40,9 +41,15 @@ class DiffBasedAnomalyDetector(AnomalyDetectorBase):
             Defaults to ``sklearn.preprocessing.RobustScaler``
             Used for transforming model output and the original ``y`` to calculate
             the difference/error in model output vs expected.
+        require_thresholds: bool
+            Requires calculating ``thresholds_`` via a call to :func:`~DiffBasedAnomalyDetector.cross_validate`.
+            If this is set (default True), but :func:`~DiffBasedAnomalyDetector.cross_validate`
+            was not called before calling :func:`~DiffBasedAnomalyDetector.anomaly` an ``AttributeError``
+            will be raised.
         """
         self.base_estimator = base_estimator
         self.scaler = scaler
+        self.require_thresholds = require_thresholds
 
     def __getattr__(self, item):
         """
@@ -253,5 +260,11 @@ class DiffBasedAnomalyDetector(AnomalyDetectorBase):
                 ),
             )
             data = data.join(anomaly_confidence_scores)
+
+        elif self.require_thresholds:
+            raise AttributeError(
+                f"`require_thresholds={self.require_thresholds}` however `.cross_validate`"
+                f"needs to be called in order to calculate these thresholds before calling `.anomaly`"
+            )
 
         return data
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -25,6 +25,7 @@ wheels/
 .installed.cfg
 *.egg
 MANIFEST
+junit/
```
epa095: :+1:

gordo_components/model/anomaly/diff.py
```diff
@@ -78,6 +79,111 @@ def fit(self, X: np.ndarray, y: np.ndarray):
...
+            Any additional kwargs to be passed to :func:`sklearn.model_selection.cross_validate`
+
+        Returns
+        -------
+        dict
+        """
+        # Depend on having the trained fold models
+        kwargs.update(dict(return_estimator=True, cv=cv))
+
+        X = X if not isinstance(X, pd.DataFrame) else X.values
```
epa095: Too many levels of negation for my head, I find `X = X.values if isinstance(X, pd.DataFrame) else X` much easier to read. No?

gordo_components/model/anomaly/diff.py
```diff
@@ -78,6 +79,112 @@ def fit(self, X: np.ndarray, y: np.ndarray):
...
+        y = y if not isinstance(y, pd.DataFrame) else y.values
+
+        # Base cv_output dict, which we'll supplement
```
epa095: Do we supplement it? Can not see where that happends

gordo_components/model/anomaly/diff.py
```diff
@@ -78,6 +79,112 @@ def fit(self, X: np.ndarray, y: np.ndarray):
...
+        cv_output = cross_validate(self, X=X, y=y, **kwargs)
+
+        thresholds = pd.DataFrame()
+
+        for i, ((test_idxs, _train_idxs), model) in enumerate(
```
epa095: maybe call the model `split_model` or something, to indicate to first-readers that it varies across the splits

gordo_components/model/anomaly/diff.py
```diff
@@ -132,4 +239,20 @@ def anomaly(
         # Calculate the total anomaly
         data["total-anomaly"] = np.linalg.norm(data["tag-anomaly"], axis=1)
 
+        # If we have `thresholds_` values, then we can calculate anomaly confidence
+        if hasattr(self, "thresholds_"):
+            y = y if not hasattr(y, "values") else y.values
```
epa095: Exactly as on line 110, but done differently

gordo_components/model/anomaly/diff.py
```diff
@@ -78,6 +79,112 @@ def fit(self, X: np.ndarray, y: np.ndarray):
...
+            y_pred = model.predict(X[test_idxs])
+            y_true = y[test_idxs]
+
+            diff = self._fold_thresholds(y_true=y_true, y_pred=y_pred, fold=i)
+
+            # Accumulate the rolling mins of diffs into common df
+            thresholds = thresholds.append(diff)
+
+        # And finally, get the mean of all the pre-calculated thresholds over the folds
+        self.thresholds_ = self._final_thresholds(thresholds=thresholds)
```
epa095, line 136: This is a bit too subtle for me, that we have to run cross_validate on the same object before calling `anomaly` to have this variable set. Is there maybe some way we can make it a bit more explicit

gordo_components/model/anomaly/diff.py
milesgranger: In my head we were going to. :laughing:

gordo_components/model/anomaly/diff.py
milesgranger: Keep things interesting! :stuck_out_tongue_winking_eye: 
I fix, I fix...

gordo_components/model/anomaly/diff.py
milesgranger: #false You only get thresholds if `cross_validate` method was called. It will still work (calling `.anomally()` that is) without calling `cross_validate` 

IMO, it is worth the trade off of an ugly wrapping of `cross_validate` and then re-initializing the same Model depending on some output. Implementing an optional `cross_validate` as an optional method, keeps things quite clean in the `build_model` function.

The requirement of _this_ model needing access to folds is exceptionally specific to _this_ model.

milesgranger: otherwise, we can call `cross_validate` inside of `fit` if `thresholds_` is not defined?

flikka (approved):
Like it!
Regarding the closing of https://github.com/equinor/gordo-components/issues/650 - I guess there are things that are _not_ implemented in this PR, which I find perfectly fine - as long as we don't close it, but maybe "tick off" the checkboxes.

PR #674, opened 2019-11-18
Make global setting of model-server-resources mandatory

### LINKED ISSUE #665: Configfile: Remove per-model server resources
Today we support setting resource limit/requests in the configfile for server/builder/client in several ways:

- They can all be set in the globals section, and then they hold for all. 
- The server can (I think) also be set per model, but the client and model-builder can not. 

The latter made sense before when we had one server per model, but not now when we have only one server. So now it should not be possible to set it per model, but only in the globals section. Probably mostly involves removing tests and some special code.

ryanjdillon:
This PR does the following:
* adds a check that disallows setting of mode-server resources per machine.
* exposes ignore in `diffpatch.diff` for future events where we want to only update part of a patch (had we just wanted to ignore the user input and not raise an exception).
* Update tests to pytest while adding/update test cases

Will close #665

### DIFF
--- a/examples/config_crd.yaml
+++ b/examples/config_crd.yaml
@@ -10,27 +10,17 @@ spec:
   config:
     machines:
 
-      - dataset:
+      - name: ct-23-0001 #1st machine
+        dataset:
           tags: #list of tags for 1st machine
             - TAG 1
             - TAG 2
             - TAG 3
           train_start_date: 2016-11-07T09:11:30+01:00
           train_end_date: 2018-09-15T03:01:00+01:00
-        runtime:
-          server:
-            resources:
-              requests: # The resources we require reserved
-                memory: 1700 # Request 1.7G of memory
-                cpu: 2000 # Request two cores, while e.g. 250 would mean a quarter core
-              limits: # The resources we are limited to using.
-                memory: 2000 #Kill the server if it uses more than 2GB
-                cpu: 2000 #Throttle the server to never using more than 2 cores
-        metadata:
-            model-hierarchy: "1218 -\u003e -\u003e g's"
-        name: ct-23-0001 #1st machine
 
-      - dataset:
+      - name: ct-23-0002 #2nd machine
+        dataset:
           resolution: 2T
           tags: #list of tags for 2nd machine
             - TAG 1
@@ -39,28 +29,27 @@ spec:
           train_start_date: 2018-05-20T01:00:04+02:00
           train_end_date: 2019-05-10T15:05:50+02:00
         runtime:
-          server: #one can override just a single resource request as well
+          builder: #one can override just a single resource request as well
             resources:
               requests:
                 memory: 1000
           influx: # Dont push results for this single model to influx
             enable: False
-        name: ct-23-0002 #2nd machine
 
-      - dataset:
+      - name: ct-23-0003 #3rd machine
+        dataset:
           tags: #list of tags for 3rd machine
             - TAG 1
             - TAG 2
             - TAG 3
           train_start_date: 2018-09-15T13:03:04+02:00
           train_end_date: 2019-03-11T11:05:10+02:00
-        name: ct-23-0003 #3rd machine
 
     globals:
       runtime:
         influx: # Change to False to disable influx. Default value: True
           enable: True
-        builder: # The model builder can only be customized globally, any per-machine settings will be ignored.
+        builder: # The model builder can be customized per-machine, or set globally
           resources:
             requests:
               memory: 4000 #4 GB of memory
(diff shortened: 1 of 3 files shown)


examples/config_crd.yaml
```diff
@@ -1,7 +1,7 @@
 # Workflow generator configuration file compatible with `gordo`'s Kubernetes
 # custom resource defintion.  You may also refer to the `config.yaml` file in
 # the `gordo-test-project` repository
-apiVersion: equinor.com/v1
+apiVbuilder: equinor.com/v1
```
epa095: bugbug

tests/gordo_components/workflow/test_config_elements.py
```diff
@@ -20,203 +20,241 @@
...
+              requests:
+                memory: 5
+                cpu: 5
+              limits:
+                memory: 5
+                cpu: 5
         """
-        self.maxDiff = None
+    # Setting server runtime parameters by machine (i.e. not globals) should raise an error
+    with pytest.raises(ValueError):
```
epa095: Maybe extract into its own test? The two tests can share the `element_str` (by making it a constant in the file)

epa095 (commented):
looking good

examples/config_crd.yaml
ryanjdillon: Yikes....  I replaced `ersion` with `builder`? :thinking:

gordo_components/workflow/config_elements/machine.py
```diff
@@ -71,6 +71,11 @@ def from_config(  # type: ignore
         model = config.get("model") or config_globals.get("model")
 
         local_runtime = config.get("runtime", dict())
+        if "server" in local_runtime:
+            raise ValueError(
+                "Runtime settings for the model-server must be set as globals, not at machine level."
+            )
+
```
milesgranger: This _seems_ like it should rather be in the descriptor (`ValidMachineRuntime`) as that's where we're doing all validation of what constitutes a 'valid' runtime for a machine. 

Also, with that said, it seems to me that maybe this shouldn't raise an error, but a deprecation warning if anything? I can stick `foo` into the runtime as well, and it wouldn't care, right? Since it looks at specific keys not the keys actually contained. But maybe it does makes sense; making it painfully obvious for people upgrading versions. But even so, do we think anyone _actually_ sets the server runtime?

ryanjdillon: >This seems like it should rather be in the descriptor (ValidMachineRuntime) as that's where we're doing all validation of what constitutes a 'valid' runtime for a machine.

Make sense :+1: 

On :zap: Warning:zap:   VS :boom: Exception:boom:, @epa095 preferred hard compliance here, and after thinking about it a bit I think I'm on-board with that. If said warning is not noticed, it could make for some confusing debugging.

ryanjdillon: Tried out doing things as you suggested for the first one, but then the machine validator fails on the default server resources that get injected to the machine.

milesgranger: Then I suggest we fix that. Because doing a validation here, is not the correct place when we have all other verification of attributes in descriptors.

The possibilities I see here is that `DEFAULT_CONFIG_GLOBALS` should maybe be split into what is a _valid_ `Machine` global, and a valid config `global`, as it were, is a superset of a valid Machine global. Because it appears these are now two different things, but we're still treating them the same. And then update the tests to use the appropriate `global`/`Machine` default configs.

Ooorrrr. just ignore the key, like we would ignore _any other_ key which we don't care about and this whole problem goes away.

ryanjdillon:
@epa095 Miles and I discussed this again, and I'm on-board with his take on it, which is:

As it stands now, the server-runtime settings are read from the globals/default_globals, so the only thing that could happen here would be somebody setting the server resources for a machine, it does nothing, and then they are confused when the server resources don't reflect this. Miles checked with all users on Slack, and nobody is doing this, soooo it shouldn't be a problem.

**Short-term solution**
Just leave it

**Long-term solution**
Have separate configs for globals and (machine-globals, and avoid messy validation solutions with how things are currently patched in.

ryanjdillon:
removed error raising, left the rest, reeady for merge.

epa095 (approved):
LGTM, merge on will when rebased

gordo_components/workflow/workflow_generator/helpers.py
```diff
@@ -29,6 +39,6 @@ def patch_dict(original_dict: dict, patch_dictionary: dict) -> dict:
         `original_dict`
 
     """
-    diff = dictdiffer.diff(original_dict, patch_dictionary)
+    diff = dictdiffer.diff(original_dict, patch_dictionary, ignore=ignore)
```
milesgranger: All these changes in this file are no longer required, correct?

milesgranger (dismissed):

gordo_components/workflow/workflow_generator/helpers.py
milesgranger: just a note...I'm ok with it.. but just exposing parameters to expose them is slightly strange IMO if this is left in.

ryanjdillon: This isn't used any longer. I did some digging into the `dictdiffer` lib to find this, and while I'm not using it any longer, I thought that it is a useful thing to be aware of when doing this config patching. 

That said, I think the patching in general may be a confusing/problematic pattern to follow, so I'll just remove this with our discussion on changing things in the future.

PR #678, opened 2019-11-19
Add --ignore benchmarks & refactor pytest.ini

milesgranger:
Running `python -m pytest` will try loading benchmark tests. 
This adds the `--ignore benchmarks` flag and refactors the long line as well.


epa095 (approved):
Awesome. Would you also be so kind to change the order of the 3 last lines so the two "--cov" ones are at the end. It makes it easier to comment them out for us meere mortals which sometimes needs to debug stuff with pycharm ;-)

PR #683, opened 2019-11-19
Workflow: Clean up quoting data to builder

### LINKED ISSUE #682: Workflow: handle metadata containing single quotes
today we produce invalid yaml if the metadata contains a string with a single quote `'`.

epa095:
This standardizes on using the jinja2 json filter and yaml literal
quote to send data to the model-builder for all data. Earlier we relied
on python string-formatting the values in a sane way, which did not work
if the strings contained a single quote, since then python would switch to
using double quotes where it usually uses single quotes.

This closes #682

### DIFF
--- a/gordo_components/workflow/workflow_generator/resources/argo-workflow.yml.template
+++ b/gordo_components/workflow/workflow_generator/resources/argo-workflow.yml.template
@@ -1301,13 +1301,24 @@ spec:
         - name: model-builder-{{ machine.name }}
           template: model-builder
           arguments:
-            parameters: [{name: model-name, value: "{{ machine.name }}"},
-                         {name: model-config, value: "{{ machine.model }}"},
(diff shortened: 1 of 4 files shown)


gordo_components/workflow/workflow_generator/resources/argo-workflow.yml.template
```diff
@@ -1301,13 +1301,25 @@ spec:
...
+                  {{ machine.metadata | tojson }}
+              - name: data-config
+                value: |
+                  {{ machine.dataset.to_dict() | tojson }}
+              - name: data-provider
+                value: |
+                  {{ machine.data_provider.to_dict() | tojson }}
+              - name: evaluation-config
+                value: |
+                  {{ machine.evaluation | tojson}}
```
milesgranger, line 1321: Hawt! :fire:

milesgranger (approved):
Bravo!

PR #684, opened 2019-11-19
Add SensorTagNormalizationError with exit code 60

### LINKED ISSUE #677: Add exit code for ValueError
When we are not able to deduce the asset from a tag we return a ValueError. We should return a custom exit code for these, alternatively (and maybe preferably), we could sublcass ValueError and return that exception, and then have custom exit code for that specific error.

Example:
`ValueError: Unable to find asset for tag with name NO 6507/11-Y-1.OnStream `

epa095:
This closes #677

### DIFF
--- a/gordo_components/cli/cli.py
+++ b/gordo_components/cli/cli.py
@@ -8,6 +8,7 @@ import logging
 import sys
 import traceback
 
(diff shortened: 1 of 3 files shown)


milesgranger (approved):
Nice and clean. :bowtie:

PR #685, opened 2019-11-19
Workflow: delete pods when pod succeeds

epa095:
It was when the workflow succeeds, this will leave only failing pods.

[ci skip]


epa095:
Turns out argo 2.4 has some bugs we can not live with, so we downgrade to argo 2.3. Thus this PR is useless, and superseeded by #695

PR #689, opened 2019-11-20
Refactor model building.

### LINKED ISSUE #434: Refactor build in cli.py
The function handling build starts to get a bit overwhelming, and should perhaps be split into sub functions, to make it more readable and easier to maintain.

milesgranger:
Will close #434

### DIFF
--- a/gordo_components/builder/build_model.py
+++ b/gordo_components/builder/build_model.py
@@ -16,6 +16,7 @@ import pandas as pd
 import numpy as np
 import tensorflow as tf
 
+import sklearn
 from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn import metrics
 from sklearn.model_selection import cross_validate, TimeSeriesSplit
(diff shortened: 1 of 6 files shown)


gordo_components/builder/build_model.py
```diff
@@ -32,593 +33,577 @@
...
-        logger.debug("Starting cross validation")
-        start = time.time()
-        if hasattr(model, "predict"):
+                cv = cross_validate(
+                    model,
+                    X,
+                    y,
+                    scoring=metrics_dict,
+                    return_estimator=True,
+                    cv=TimeSeriesSplit(n_splits=3),
```
EinarJohnsen, line 168: Possbile to specify a different number of splits?

milesgranger: Oh nice. :+1:

gordo_components/builder/build_model.py
```diff
@@ -32,593 +33,577 @@
...
+        Returns
+        -------
+            dict
+        """
+        if isinstance(scaler, str):
+            scaler = serializer.pipeline_from_definition(scaler)
+
+        if scaler:
+            logger.debug("Fitting scaler for scoring purpose")
+            scaler.fit(y)
```
EinarJohnsen: Better to check if it is a str if it actually exists?
```
if scaler:
    logger.debug("Fitting scaler for scoring purpose")
    if isinstance(scaler, str): 
       scaler = serializer.pipeline_from_definition(scaler)
    scaler.fit(y)
```

milesgranger: meh, I _think_ it's functionally equivalent, because it'll at least be defined as `None` but I do like the suggestion as it's more concise and arguably more readable, as dealing the scaler logic is in a single place. :+1:

gordo_components/builder/build_model.py
milesgranger: I started, but I think maybe it's better to leave this until after this is merged. So far as I can see, maybe this should be tackled with the whole of the ability to define the splitter in the `evaluation` key. 

ie. 
```yaml
evaluation:
  cv:
    sklearn.model_selection.TimeSeriesSplit:
      n_splits: 5
```

gordo_components/builder/build_model.py
EinarJohnsen: Cool, I agree that it is functionally equivalent, and probably not computational heavy to check if it is a string. :sunglasses:

gordo_components/builder/build_model.py
EinarJohnsen: I agree, I think you have mentioned this before and it sounds like a good idea.

gordo_components/builder/build_model.py
```diff
@@ -486,7 +486,7 @@ def check_cache(self, model_register_dir: Union[os.PathLike, str]):
             )
             return None
 
-    def provide_saved_model(
+    def build_with_cache(
```
EinarJohnsen, line 492: Hmm. Do you feel this is explaining the function good enough? :thinking:

milesgranger: :man_shrugging: Slightly better than previous? Do you have a suggestion? :pray:

EinarJohnsen: Yes, I think it is better than the previous, but I am uncertain as it is a "if this else that" kind of function. But as I have no better to come up with, please ignore my nitpicking. :pick:

milesgranger: Ya, I agree. It feels like this outer check in the CLI of `if evaluation_config["cv_mode"] == ...` should be held inside of the `ModelBuilder`. But maybe that's ok actually? I don't really know. I implemented it as such inside the builder, but it looked ugly. Sooo, I've switched it back and will think more. Otherwise we can always do an additional (smaller) refactor after this is in. :question:

EinarJohnsen: Yeah, it is related to the the model building as it checks what to build/skip. But I think it is fine to do a smaller refactor later if we need it. It is quite a rewrite itself, so doing CLI in it's own is perhaps fine.

gordo_components/builder/build_model.py
```diff
@@ -32,593 +33,576 @@
...
+        """
+        Use the raw data from Gordo config file keys: name, model, dataset, metadata, and evalution
+        to build the final ML model.
+
+        Parameters
+        ----------
+        name: str
+            Name of model to be built
+        model_config: dict
+            Mapping of Model to initialize and any additional kwargs which are to be used in it's initialization.
```
epa095: long lines in the comments in generall

gordo_components/builder/build_model.py
```diff
@@ -32,593 +33,576 @@
...
+        """
+        out = model.predict(X) if hasattr(model, "predict") else model.transform(X)
+        return len(X) - len(out)
+
+    @staticmethod
+    def _save_model_for_workflow(
+        model: BaseEstimator, metadata: dict, output_dir: Union[os.PathLike, str]
+    ):
+        """
+        Save a model according to the expected Argo workflow procedure.
```
epa095: no longer "a model", now it is "the model" (same for all other "a model")

gordo_components/builder/build_model.py
```diff
@@ -32,593 +33,576 @@
...
         else:
-            cached_model_location = check_cache(model_register_dir, cache_key)
-            if cached_model_location:
+            logger.info(
+                f"Did not find the model with key {self.cache_key} in the register at "
+                f"{model_register_dir}."
+            )
+            return None
+
+    def build_with_cache(
```
epa095, line 492: I propose we change this so it returns the same as the `build` method (i.e. it loads the model/metadata if it exists in the cache, and returns that model/metadata instead of the location). Maybe they even can just become one function? Then cli.py can become much prettier!

Why was it like this? I think it was because earlier the caching worked by us writing the path to the special ugly file(!), while now we are anyway copying it and stuff.

epa095 (commented):
god start!

gordo_components/builder/build_model.py
milesgranger: Cool, we'll leave this for the follow up PR. It'll look dapper as dandy Daniel from DecadentDandyVille

epa095 (approved):
I like it more than it sounded like

PR #693, opened 2019-11-20
Add back PredictionForwarder callable interface

milesgranger:
Back by popular demand. @mrdeveloperdude :business_suit_levitating:


epa095 (approved):
LGTM, I think. Unless we want to take this opportunity to change it away from __call__?

milesgranger:
I don't think so, having it as `__call__` allows us to take a good ol' function should someone want to supply that.

PR #694, opened 2019-11-20
Dockerfiles: Improve caching of large packages

epa095:
This patch adds a step in the dockerfile which installs a few big
packages (tensorflow, pyarrow, scipy, and catboost) before the
requirements file is copied over. This allows docker to cache the layer
with those packages even if other requirements change.

Now with dependabot our requirements-file change all the time! And a few of our dependencies sum up to almost 700Mb extracted, so I moved them into a seperate step.

### DIFF
--- a/Dockerfile-Client
+++ b/Dockerfile-Client
@@ -10,11 +10,21 @@ RUN rm -rf /code/dist \
     && python setup.py sdist \
     && mv /code/dist/$(ls /code/dist | head -1) /code/dist/gordo-components-packed.tar.gz
 
+# Extract a few big dependencies which docker will cache even when other dependencies change
+RUN cat /code/requirements.txt | grep tensorflow== > /code/prereq.txt \
+    && cat /code/requirements.txt | grep pyarrow== >> /code/prereq.txt \
+    && cat /code/requirements.txt | grep scipy== >> /code/prereq.txt \
+    && cat /code/requirements.txt | grep catboost== >> /code/prereq.txt
+
 FROM python:3.6.9-slim-stretch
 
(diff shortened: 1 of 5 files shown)


Dockerfile-Client
```diff
@@ -10,11 +10,21 @@ RUN rm -rf /code/dist \
...
+RUN cat /code/requirements.txt | grep tensorflow== > /code/prereq.txt \
+    && cat /code/requirements.txt | grep pyarrow== >> /code/prereq.txt \
+    && cat /code/requirements.txt | grep scipy== >> /code/prereq.txt \
+    && cat /code/requirements.txt | grep catboost== >> /code/prereq.txt
+
 FROM python:3.6.9-slim-stretch
 
 # Install requirements separately for improved docker caching
+COPY --from=builder /code/prereq.txt .
+RUN pip install --no-deps -r prereq.txt --no-cache-dir
```
milesgranger, line 23: Why `--no-deps` ?

milesgranger (approved):
Looks pretty good to me.. always be optimizing as the old saying goes.
Just a single comment/question.

Dockerfile-Client
epa095: Becasue we dont want this step to pull inn (possibly unpinned) dependencies for these things. With this way here we are downloading and installing only these packages, and then installing the rest afterwards. 

We *could* drop --no-deps, but then things would get interesting in `pip install -r /code/requirements.txt`, since it would then potentially have to uninstal the installed-but-wrong version of something.

milesgranger: So flink, you've thought of everything. :+1:

PR #699, opened 2019-11-21
Add default_asset to sensortag and dataset

epa095:
The key `default_asset` specifies a fallback which is used as the asset
if the tagname-to-asset resolving fails. It compliments the `asset`
keyword, which bypasses the resolving and forces the asset to be
`asset`.

### DIFF
--- a/gordo_components/dataset/sensor_tag.py
+++ b/gordo_components/dataset/sensor_tag.py
@@ -33,10 +33,29 @@ TAG_TO_ASSET = [
 ]
 
 
-def _asset_from_tag_name(tag_name: str):
+def _asset_from_tag_name(tag_name: str, default_asset: str = None):
     """
     Resolves a tag to the asset it belongs to, if possible.
-    Returns None if it does not match any of the tag-regexps we know.
+
+    If `default_asset` is provided then it will be used as the fallback in case it
+    is impossible to resolve tag_name to an asset.
+
+    Parameters
+    ----------
+    tag_name : str
+        Tag to deduce the asset from.
+    default_asset : str
+        Asset to report if we are not able to resolve tag_name to an assets.
+
+    Returns
+    -------
+    str
+        The asset for provided tag.
+
+    Raises
+    ------
+    SensorTagNormalizationError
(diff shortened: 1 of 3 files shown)


gordo_components/dataset/sensor_tag.py
```diff
@@ -33,10 +33,11 @@
...
 
 
-def _asset_from_tag_name(tag_name: str):
+def _asset_from_tag_name(tag_name: str, default_asset: str = None):
     """
-    Resolves a tag to the asset it belongs to, if possible.
-    Returns None if it does not match any of the tag-regexps we know.
+    Resolves a tag to the asset it belongs to, if possible. Raises
+    SensorTagNormalizationError if it does not match any of the tag-regexps we know
+    and default_asset is not set.
```
milesgranger: Should this just follow the standard numpy docs if we're also adding what it can raise?

tests/gordo_components/dataset/test_sensor_tag.py
```diff
@@ -7,28 +7,30 @@
...
-                SensorTag(TAG_NAME1, asset_nonsense),
-                SensorTag(TAG_NAME2, asset_nonsense),
+                SensorTag(NON_RESOLVABLE_TAG_NAME1, asset_nonsense),
+                SensorTag(NON_RESOLVABLE_TAG_NAME2, asset_nonsense),
             ],
         ),
         (
             ["TRC-123", "GRA-214", "ASGB-212"],
             "ThisWillBeTheAsset",
+            "AndThisWIllBeIgnoredSinceAllTagsAreResolvable",
```
milesgranger: Nit pick of nit picks: `..Will..`

epa095: damn, you actually read those?

milesgranger: I read your code like it was written by the hand of god.

milesgranger (approved):
Approving, because you addressed 1 of 2 comments. :100:

flikka (approved):
The computer says yes

Issue #700, opened 2019-11-21
Add tag matching and asset to path for 1140, 1170, 1219


flikka:
1904 Johan Sverdrup
1140 Sleipner
1170 Heidrun
1219 Aasta Hansteen

EinarJohnsen:
Are we able to infer the tag matching before the paths are created?

flikka:
> Are we able to infer the tag matching before the paths are created?

I would guess that we could do the tag-to-asset, but not the asset-to-path.

Seems like "1904" is the prefix for 1904, 1140 sleipner has no prefix, 1170 Heidrun has "HD", and 1219 aasta has "1219".

EinarJohnsen:
Took a look for these beauties, but sadly I can not see them yet.

flikka:
@ryanjdillon this is related to https://github.com/equinor/gordo-infrastructure/issues/508  - closing this one

PR #701, opened 2019-11-22
Add new asset to path

EinarJohnsen:
Add new tag_to_asset for 1904, 1170 and 1219 (that has prefixes). Fixes part of #700


epa095 (commented):
But the paths, do we not have paths?

EinarJohnsen:
No, all these are missing from the datalake (unless they are very well hidden somewhere).

epa095 (approved):
Lovely!

epa095:
Maybe change commit message btw @EinarJohnsen ? Is it possible to get the assets nr in the main message and still keep it less than 50?

PR #702, opened 2019-11-22
Add NoSuitableDataProviderError

epa095:
To indicate that we have requested an tag which no dataproviders are
able to handle

### DIFF
--- a/gordo_components/data_provider/providers.py
+++ b/gordo_components/data_provider/providers.py
@@ -23,6 +23,10 @@ from gordo_components.dataset.sensor_tag import SensorTag
 logger = logging.getLogger(__name__)
 
 
+class NoSuitableDataProviderError(ValueError):
+    pass
+
+
 def load_series_from_multiple_providers(
     data_providers: typing.List[GordoBaseDataProvider],
     from_ts: datetime,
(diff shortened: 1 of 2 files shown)


gordo_components/data_provider/providers.py
```diff
@@ -23,6 +23,10 @@
 logger = logging.getLogger(__name__)
 
 
+class NoSuitableDataProviderError(ValueError):
+    pass
```
milesgranger, line 27: Do you think we should have a common collection for gordo exceptions?

milesgranger (approved):
Approving with single comment, take it or leave it. But also, consider reviewing some of my PRs?

gordo_components/data_provider/providers.py
epa095: By collection do you mean to put them somewhere together? IDK, I feel like errors like these belongs in only class which throws them, they are part of the API of this class. But more generall errors can definitely be collected somewhere.

milesgranger: Ya, I mean libs like `sklearn` have them in `sklearn.exceptions` and in general that seems fairly standard. But I do not have strong opinions about this, yet. :-)

PR #707, opened 2019-11-25
Add thread lock to data_provider get_client

ryanjdillon:
This PR adds a threading lock to the datalake provider method `_get_client`, so that  `n` threaded calls to the `load_series` method doesn't generate `n` attempts to authenticate to Azure.

### DIFF
--- a/gordo_components/data_provider/providers.py
+++ b/gordo_components/data_provider/providers.py
@@ -2,6 +2,7 @@
 import os
 import random
 import logging
+import threading
 import timeit
 
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


gordo_components/data_provider/providers.py
```diff
@@ -151,12 +153,16 @@ def load_series(
...
 
     def _get_client(self):
-        if not self.client:
-            self.client = create_adls_client(
-                storename=self.storename,
-                dl_service_auth_str=self.dl_service_auth_str,
-                interactive=self.interactive,
-            )
+        with self.lock:
+            logger.info("Acquiring threading lock for Datalake authentication.")
```
milesgranger: This should be above the `with` statement, as it's acquired the lock by this point.

epa095 (approved):
LGTM

PR #710, opened 2019-11-26
Restarting k8s watch on error events

epa095:


milesgranger (approved):
I'm going to conditionally approve this, because I think in general this whole module needs some TLC when it comes to testing.

PR #716, opened 2019-11-26
ModelBuilder: support optional caching in build

### LINKED ISSUE #709: Pretty up model builder CLI
`provide_saved_model`  in the model_builder can be changed to return the same as  `build` by just loading the model and returning it, and the two functions can maybe even become a single function. Then a lot of uflyness can be removed from cli.py

milesgranger:
Will close #709

### DIFF
--- a/gordo_components/builder/build_model.py
+++ b/gordo_components/builder/build_model.py
@@ -125,27 +125,23 @@ class ModelBuilder:
         """
         Always return a model and its metadata.
 
-        If ``output_dir`` is supplied without ``model_register_dir`` the model
-        and metadata will be deposited to this output directory.
-
-        However if ``model_register_dir`` is supplied, it will attempt to read
-        a previously cached model. Failing to find it will simply build the model.
-
-        If both ``model_register_dir`` and ``output_dir`` are specified, it will
-        check the cache and copy over to the output directory, otherwise build
-        and cache the model.
+        If ``output_dir`` is supplied, it will save the model there.
+        ``model_register_dir`` points to the model cache directory which it will
+        attempt to read the model from. Supplying both will then have the effect
+        of both; reading from the cache and saving that cached model to the new
+        output directory.
 
         Parameters
         ----------
         output_dir: Optional[Union[os.PathLike, str]]
-            A path to where the model will be deposited if it is built.
+            A path to where the model will be deposited.
         model_register_dir: Optional[Union[os.PathLike, str]]
-            A path to a register, see `:func:gordo_components.util.disk_registry`. If this is None
-            then always build the model, otherwise try to resolve the model from the
-            registry.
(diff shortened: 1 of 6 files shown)


flikka (commented):
Net removal of code-lines, so sweet.

milesgranger:
I don't know if this makes things better or not. :confused:

gordo_components/builder/build_model.py
```diff
@@ -109,7 +109,111 @@ def __init__(
...
+        output_dir: Optional[Union[os.PathLike, str]] = None,
+        model_register_dir: Optional[Union[os.PathLike, str]] = None,
+        replace_cache=False,
+    ) -> Tuple[Optional[sklearn.base.BaseEstimator], dict]:
+        """
+        Build the model and/or ensure that the desired model exists on disk
+        in `output_dir`, and returns the path to it. If `output_dir` exists we
+        assume the model is there (no validation), and return that path.
+
+        If ``output_dir`` is not given, it will simply build the model without
```
epa095: I think this whole logic was from the time when we copied and returned the path, not the actual models. Today I think it makes sense that the caching can be used without output_dir set (but it wont be able to write to the cache though) .  I think that can make it nicer. Maybe ditch the whole "copy model blabla" code and just try to read from the cache if it is there, and then serialize it to output_dir if that is set (so slightly less effecient since it will load/write instead of copy, but more effeciient)

epa095 (commented):
I think I see potential!

gordo_components/model/anomaly/diff.py
```diff
@@ -118,6 +118,8 @@ def cross_validate(
 
         thresholds = pd.DataFrame()
 
+        model_threshs = list()
```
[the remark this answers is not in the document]
milesgranger: :man_facepalming:

gordo_components/builder/build_model.py
```diff
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 
 import datetime
-from distutils.dir_util import copy_tree
```
epa095, line 4: :tada:

gordo_components/builder/build_model.py
```diff
@@ -109,7 +108,93 @@ def __init__(
...
+        Always return a model and its metadata.
+
+        If ``output_dir`` is supplied without ``model_register_dir`` the model
+        and metadata will be deposited to this output directory.
+
+        However if ``model_register_dir`` is supplied, it will attempt to read
+        a previously cached model. Failing to find it will simply build the model.
+
+        If both ``model_register_dir`` and ``output_dir`` are specified, it will
+        check the cache and copy over to the output directory, otherwise build
```
epa095: no copy no more

epa095 (approved):
I think it looks much nicer :-)

PR #732, opened 2019-11-27
Don't convert DataFrame to ndarray in cv

### LINKED ISSUE #731: DataFrame to ndarray conversion causes column dropper to fail

flikka:
Close #731


milesgranger (approved):
Nice catch, really saved our bacon. :bacon: :egg:

PR #733, opened 2019-11-27
Workflow: Delete old model objects

### LINKED ISSUE #717: Delete old models from k8s
Just as we delete other old resources, we should delete the old models.

https://github.com/equinor/gordo-components/blob/33504079dd17139e881cea543db6d00f2570ad79/gordo_components/workflow/workflow_generator/resources/argo-workflow.yml.template#L1258

epa095:
This closes #717


milesgranger (approved):
Nailed it.

PR #734, opened 2019-11-27
Change client dependencies in argo template

### LINKED ISSUE #718: Workflow: Client should depend on gordo-mdl-{{ machine.name }}
Earlier the client depended on the model-building step, which included the model notification step as well (causing watchman to gather the metadata for the model). Now the model-notification step is a seperate step on the top-level, so the client-step should explicitly depend on it, not on the model-builder step. 

So change the model-builder dependency in https://github.com/equinor/gordo-components/blob/fb5dfb5ee90f32847e221f7340ad480a571961fa/gordo_components/workflow/workflow_generator/resources/argo-workflow.yml.template#L1366  with `
gordo-mdl-{{ machine.name }}`

ryanjdillon:
This PR update the dependencies for the the client from using the model notification step (gordo-mdl) instead of the model builder step.

The test project completed successfully with this, with the clients kicking off with the completion of the mdl step. :checkered_flag: 

Closes #718

### DIFF
--- a/gordo_components/workflow/workflow_generator/resources/argo-workflow.yml.template
+++ b/gordo_components/workflow/workflow_generator/resources/argo-workflow.yml.template
@@ -1364,7 +1364,7 @@ spec:
           dependencies:
             - gordo-server
             - gordo-influx
-            - model-builder-{{ machine.name }}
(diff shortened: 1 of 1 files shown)


epa095 (approved):
LGTM

PR #735, opened 2019-11-27
Workflow: Remove gordo-mdl from the model-names

epa095:
I dont think they provide any value really, since all the `models` are gordo models.


EinarJohnsen (approved):
LGTM

Issue #737, opened 2019-11-27
Gordo production output: Evaluation of threshold and creation of health index


henrikny:
### Threshold evaluation

**Complete model**

For each model in production, apply the following scheme.

- [x] 1. Fetch the model thresholds (described in #736) from each of the three splits.
- [x] 2. Calculate the `average_threshold` over all splits.
- [x] 3. Fetch the `model_mse` (described in #650).
- [x] 4. Compare the vector against the known threshold and return a vector of health indicator [`model_mse` / `average threshold`].

**Individual input variable**

For every variable of every model in production, apply the following scheme:

- [x] 1. Fetch the individual input variable thresholds (described in #736) from each of the three splits.
- [x] 2. Calculate the `average_threshold` over all splits.
- [x] 3. Fetch the `tag_abs_res` (described in #650).
- [x] 4. Compare the vector against the known threshold and return a vector of health indicator [`tag_abs_res` / `average threshold`].

### Health indicator smoothing

The threshold evaluation is conveniently evaluated in Gordo for now, and is in practice a part of the Gordo output.
A smoothing of the health indicator is actually wanted, which applies a rolling look-back-window of size 6 which selects the minimum.
This should be in place for the complete model as well as for each individual reconstructed input variable.

To be able to create this smoothed health indicator by a look-back-window of six, additional points need to be fetched.

For given
- start timestamp t_1
- stop timestamp t_n and
- resolution r = delta[t_i] = t_i - t_[i-1] (e.g. 10min),
the vector length of the calculated metrics is (t_stop - t_start) / r = n (e.g. 24h / 10min = 144).

Fetch |T| = n + 5 data points needed at T = {t_-5, ... , t_-1, t_1, ..., t_n}.

It is unclear where this logic should reside for now, or if it should be considered for the November-2019 milestone.
In future implementations, this logic should reside outside Gordo, as more complicated factors might be added to the evaluation of the thresholds.

PR #738, opened 2019-11-28
DiffBasedAnomalyDetector: Add total-anomaly-confidence calculation

### LINKED ISSUE #736: Thresholds during model training
### Thresholds

These are the different thresholds that we want to define under model training.
The thresholds are used to create the health indicator under production.

As the thresholds are a part of the future live model evaluation, they need to be **logged** as well as kept as a part of the **meta-data**.

For both the individual sensors as well as for the complete model, a threshold is to be computed.

**Complete model**
For each validation set from the three time series cross-validation splits:

- [x] 1. Get the `model_mse` (described in #650) vector.
- [x] 2. Apply a rolling look-back-window of size 6 which selects the minimum.
- [x] 3. Return the scalar maximum value from the complete vector as the model threshold.

**Individual input variable**
For each validation set from the three time series cross-validation splits and for each variable:

- [x] 1. Get the `tag_abs_res` (described in #650) vector for each input variable.
- [x] 2. Apply a rolling look-back-window of size 6 which selects the minimum.
- [x] 3. Return the scalar maximum value from the complete vector as the variable threshold.

milesgranger:
#673 added support for feature-wise `anomaly-confidence` scores. 

This adds the remaining `total-anomaly-confidence` which is the aggregate anomaly score for the sample/timestep itself.

Part of #650 & #737  and will close #736

### DIFF
--- a/gordo_components/model/anomaly/diff.py
+++ b/gordo_components/model/anomaly/diff.py
@@ -62,13 +62,20 @@ class DiffBasedAnomalyDetector(AnomalyDetectorBase):
             return getattr(self.base_estimator, item)
 
     def get_metadata(self):
+        metadata = dict()
+
+        if hasattr(self, "feature_thresholds_"):
+            metadata["feature-thresholds"] = self.feature_thresholds_.tolist()
+        if hasattr(self, "aggregate_threshold_"):
+            metadata["aggregate-threshold"] = self.aggregate_threshold_
+
         if isinstance(self.base_estimator, GordoBase):
-            return self.base_estimator.get_metadata()
+            metadata.update(self.base_estimator.get_metadata())
         else:
-            return {
-                "scaler": str(self.scaler),
-                "base_estimator": str(self.base_estimator),
-            }
+            metadata.update(
+                {"scaler": str(self.scaler), "base_estimator": str(self.base_estimator)}
+            )
+        return metadata
 
     def score(
         self,
@@ -116,7 +123,8 @@ class DiffBasedAnomalyDetector(AnomalyDetectorBase):
 
         cv_output = cross_validate(self, X=X, y=y, **kwargs)
 
-        thresholds = pd.DataFrame()
+        feature_thresholds = pd.DataFrame()
+        scaled_mse_per_timestep = pd.Series()
 
         for i, ((test_idxs, _train_idxs), split_model) in enumerate(
             zip(kwargs["cv"].split(X, y), cv_output["estimator"])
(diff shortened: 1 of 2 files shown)


gordo_components/model/anomaly/diff.py
```diff
@@ -126,17 +134,49 @@ def cross_validate(
...
+            tag_thresholds_fold = self._feature_fold_thresholds(y_true, y_pred, fold=i)
+            feature_thresholds = feature_thresholds.append(tag_thresholds_fold)
 
-        # And finally, get the mean of all the pre-calculated thresholds over the folds
-        self.thresholds_ = self._final_thresholds(thresholds=thresholds)
+        # Calculate the final thresholds per feature based on the previous fold calculations
+        self.feature_thresholds_ = self._final_thresholds(thresholds=feature_thresholds)
+
+        # For the aggregate, use the accumulated mse of scaled residuals per timestep
+        self.aggregate_threshold_ = scaled_mse_per_timestep.rolling(6).min().max()
```
ryanjdillon: Maybe as a way of compartmentalizing this out you could do a context manager thing {SWIP-ey pseudo-code to follow), That way it might be easier to swap the method used here and have that reviewed/owned a little more externally.
```
Class AnomolyMonitor:
    feature_thresholds = list()
    aggregate_threshold = None
    def __enter__(self):
        return self
    def _final_thresholds():
        # do stuff
    def _scaled_mse_timestep():
        # do stuff
    def collect(true, pred, fold):
        self.feature_thresholds.append(_feature_fold_thresholds(true, pred, fold))
    def __exit__(self):
        # calc means
        # calc aggregate, etc.

class DiffBased
    def cv(self):
        with AnomolyMontior() as am:
            am.collect(true, pred, fold)
            self.aggregate = am.aggregate

**Review by ryanjdillon — APPROVED** [review]
Looks good to me :balloon:  Have a suggestion on abstraction, but this is likely something will have better ideas on for refinement later.

File: gordo_components/model/anomaly/diff.py [hunk 2]
```diff
@@ -237,8 +278,8 @@ def anomaly(
         data["total-anomaly"] = np.linalg.norm(data["tag-anomaly"], axis=1)
 
         # If we have `thresholds_` values, then we can calculate anomaly confidence
-        if hasattr(self, "thresholds_"):
-            confidence = tag_anomaly.values / self.thresholds_.values
+        if hasattr(self, "feature_thresholds_"):
+            confidence = tag_anomaly.values / self.feature_thresholds_.values
```
henrikny, line 283: The `tag_anomaly` is this class defined as the absolute scaled residuals.

For each input value, we want the absolute _unscaled_ residuals to be considered.

gordo_components/model/anomaly/diff.py
```diff
@@ -249,7 +290,16 @@ def anomaly(
             )
             data = data.join(anomaly_confidence_scores)
 
-        elif self.require_thresholds:
+        if hasattr(self, "aggregate_threshold_"):
+            scaled_mse = (scaled_diff ** 2).mean(axis=1)
+            data["total-anomaly-confidence"] = scaled_mse / self.aggregate_threshold_
```
henrikny, line 296: Sweet. For the total model, this is exactly what we want.

henrikny (commented):
### Regarding issue #650:

What is already in place:
- [x] `tag_pred`: Model predictions original scale (n x m). 
This is called `model-output` in the returned pandas data frame.
- [x] `tag_scaled_res`: Residuals based on scaled inputs and predictions (n x m). 
This is called `tag-anomaly` in the returned pandas data frame.
- [x] Current _total anomaly_ (scalar) based on the Frobenius (`numpy.linalg.norm`) norm (can be dropped).
This is called `total-anomaly in the returned pandas data frame.

What I cant find as output:
- [ ] `tag_abs_res`: Absolute residuals based on **unscaled** inputs and predictions (n x m).
- [x] `model_mse`: Based on the squared residuals from the **scaled** inputs and predictions, calculate the _average over all variables_ for _each point in time_ (n x 1).

The `model_mse` is evaluated, but is not a part of the returned pandas data frame, as far as I can see.


### Regarding issue #737 (which is also affected by this PR):

The individual input variables should be evaluated at their __original scale__ and not the scaled values.


### Regarding related issue #736, which this PR suggest closed:

The individual thresholds need to be based on the __original scale__, same as above.
This might even be the case, I just could not read it.
I'll come down for a chat and I'll approve it once I am on board on what this PR actually does :)

gordo_components/model/anomaly/diff.py
milesgranger: Right, this PR does not address that. We talked about how it use to be this way, (with unscaled) and then it was pushed that this should be _scaled_. We can flesh out what we're going to do there in another PR.

milesgranger:
Gave it a run through test project, all is well. :+1: 

image

PR #742, opened 2019-11-28
Change default log level to info

### LINKED ISSUE #657: Change default log-level to info and look over messages
Today the default log-level is debug, which leads to *a lot* of stuff in the logs. For production we should change to info. But then also try to run it and see if the stuff we have as info makes sense. Maybe some of the debug messages should be info, maybe some of the info messages should be debug? It should be possible to see the main steps with info, but it should not be super noisy.

EinarJohnsen:
Change log level from debug to info. Also change various log messages so the logs makes sense and shows the main steps without making to much noise. Closes #657

### DIFF
--- a/gordo_components/__init__.py
+++ b/gordo_components/__init__.py
@@ -53,14 +53,10 @@ except Exception:
     pass
 
 
-# Set log level, defaulting to DEBUG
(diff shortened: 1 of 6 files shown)


epa095 (approved):
Nice!

PR #751, opened 2019-12-03
Change KerasRaw lambda/staticmethod for serialization

ryanjdillon:
This PR performs a small fix on the `KerasRawModelRegressor`, as the `__init__` used a lambda function which couldn't be serialized.

I.E.:
```python
class KerasRawModelRegressor:
     def __init__(self, kind):
          self.kind= {}
          self.build_fn = lambda: self.keras_from_spec(self.kind)
       @staticmethod
       def keras_from_spec(spec):
          # do stuff with spec
          return model     
```
to
```python
class KerasRawModelRegressor:
     def __init__(self, kind):
          self.kind= {}
     def build_fn(self):
          # do stuff with self.kind
          return model
```

### DIFF
--- a/gordo_components/model/models.py
+++ b/gordo_components/model/models.py
@@ -291,9 +291,9 @@ class KerasRawModelRegressor(KerasAutoEncoder):
     """
 
     def __init__(self, kind: dict, **kwargs):
+
         self.kind = kind  # type: ignore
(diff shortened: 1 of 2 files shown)


epa095 (approved):
LGTM

PR #756, opened 2019-12-03
Add non-root user to docker images

ryanjdillon:
Adds non-root user to Docker images for security :lock:

### DIFF
--- a/Dockerfile-GordoDeploy
+++ b/Dockerfile-GordoDeploy
@@ -32,29 +32,49 @@ RUN cat /code/requirements.txt | grep tensorflow== > /code/prereq.txt \
     && cat /code/requirements.txt | grep scipy== >> /code/prereq.txt \
     && cat /code/requirements.txt | grep catboost== >> /code/prereq.txt
 
+
 FROM python:3.6.9-slim-stretch
 
-# Install requirements separately for improved docker caching
+# Nonroot user for running CMD
+RUN groupadd -g 999 gordo && \
+    useradd -r -u 999 -g gordo gordo
+
+ENV HOME "/home/gordo"
+ENV PATH "${HOME}/.local/bin:${PATH}"
+
+# Install requirements and gordo-components packaged in intermediate image 'python setup.py sdist'
+# perfomed separately for improved docker caching
 COPY --from=builder /code/prereq.txt .
 RUN pip install --no-deps -r prereq.txt --no-cache-dir
 
-COPY requirements.txt /code/
-RUN pip install -r /code/requirements.txt --no-cache-dir
+COPY --from=builder /code/dist/gordo-infrastructure-packed.tar.gz .
+RUN pip install gordo-infrastructure-packed.tar.gz
 
-COPY ./run_workflow_and_argo.sh /code/run_workflow_and_argo.sh
+COPY requirements.txt .
+RUN pip install -r requirements.txt --no-cache-dir
 
-COPY --from=builder /usr/local/bin/kubectl /usr/local/bin/kubectl
-COPY --from=builder /usr/local/bin/argo /usr/local/bin/argo
 RUN apt-get update && apt-get install -y \
     curl \
     jq \
  && rm -rf /var/lib/apt/lists/*
 
-COPY resources /code/resources
+COPY --from=builder /usr/local/bin/kubectl /usr/local/bin/kubectl
+COPY --from=builder /usr/local/bin/argo /usr/local/bin/argo
+COPY ./run_workflow_and_argo.sh ${HOME}/run_workflow_and_argo.sh
 
-# Install the current distribution of gordo-infrastructure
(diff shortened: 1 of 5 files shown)


Dockerfile-ModelServer
```diff
@@ -33,4 +33,9 @@ COPY --from=builder /code/dist/gordo-components-packed.tar.gz .
 # Install gordo-components, packaged from earlier 'python setup.py sdist'
 RUN pip install ./gordo-components-packed.tar.gz
 
+# Nonroot user for running CMD
+RUN groupadd -g 999 gordo && \
```
epa095: why 999?

Dockerfile-GordoDeploy
```diff
@@ -56,5 +56,9 @@ COPY resources /code/resources
 COPY --from=builder /code/dist/gordo-infrastructure-packed.tar.gz .
 RUN pip install ./gordo-infrastructure-packed.tar.gz
 
+# Nonroot user for running CMD
+RUN groupadd -g 999 gordo && \
+    useradd -r -u 999 -g gordo gordo
+USER gordo
 
 CMD ["bash", "/code/run_workflow_and_argo.sh"]
```
hjarraya: you probably want to `chown /code` folder also

Dockerfile-ModelServer
hjarraya: We use 1000

ryanjdillon: That was suggested in the Medium article you linked to. I figured he was using that as that would indicate the group/user is a dynamically generated system group/user.

Dockerfile-GordoDeploy
ryanjdillon: It has `rx` for `o` already
```bash
-rwxrwxr-x  1 rdi rdi    538 Oct 22 10:05 run_workflow_and_argo.sh
```
..but it does look like I need to `chown` more for the catboost library to do its thing in the `ModelBuilder` image.
```bash

File: Dockerfile-ModelServer (hunk 1 above)
Comment (ryanjdillon) [reply Dockerfile-ModelServer]: Should I use a 1000 as the uid/gid?

Reading [this](https://refspecs.linuxfoundation.org/LSB_3.0.0/LSB-Core-generic/LSB-Core-generic/uidrange.html), it seems that perhaps 499 may be better; though, I'm not sure what any real effect of doing one vs. the other could be.

File: Dockerfile-ModelBuilder [hunk 3]
```diff
@@ -18,24 +19,35 @@ RUN cat /code/requirements.txt | grep tensorflow== > /code/prereq.txt \
...
-# Install gordo-components, packaged from earlier 'python setup.py sdist'
-RUN pip install ./gordo-components-packed.tar.gz
+# Switch user and update needed env vars
+USER gordo
+ENV HOME "/home/gordo"
+ENV PATH "${HOME}/.local/bin:${PATH}"
 
-ADD build.sh /code/build.sh
+# Install requirements and gordo-components packaged in intermediate image 'python setup.py sdist'
+RUN pip install --no-deps -r /home/gordo/prereq.txt --no-cache-dir --user
```
epa095: We want this step before we even copy over `code/dist/gordo-components-packed.tar.gz`, so we better utilize docker caching.

epa095 (commented):
Looking better:-)

hjarraya:
You also want to look into the temp folder for your deployment descriptors see examples https://github.com/SeldonIO/seldon-core/issues/1143

Dockerfile-ModelBuilder
ryanjdillon: By *this step* you mean only the `prereq.txt`, right?

tests/conftest.py
```diff
@@ -48,7 +48,7 @@ def tmp_dir():
     """
     Fixture: Temporary directory, removed after test completion
     """
-    with tempfile.TemporaryDirectory() as tmp_dir:
+    with tempfile.TemporaryDirectory(dir="./") as tmp_dir:
```
milesgranger: What's this for? Is it also why the mounting of volumes didn't work?

ryanjdillon: That is so that `tmp_dir` gets created under the curent directory and not `/tmp`. I don't recall the mounting volumes problem you mention, but don't think so.

milesgranger: ok, but what's it changed for?

ryanjdillon: Ah, the volume bind problem was that I couldn't get the docker python client to bind the volume when using the `gordo` user. 

This here though is just something I noticed that while debugging things.  By doing this, if there is any cruft left over from a failed use `tempfile`, it is local to the repo and more obvious, rather than accumulating in `/tmp`.

ryanjdillon: Getting rid of this.

epa095 (approved):
LGTM!

Issue #744, opened 2019-11-29
EPIC: Project revision between client and server.


epa095:
**Problem:**
After a gordo project has been re-applied / recreated there are a time where we are inconsistent all over. 
- Watchman serves a project version, and also metadata. But it does currently not guarantee that the metadata it serves holds for the models of that exact version(!). This should be fixed. But for that to be fixed it needs to be able to communicate with the server what version if wants metadata for.
- The client has metadata it has fetched from watchman. If the previous point is fixed then it can know that the metadata is correct for that exact project version. But when it asks the server the server might or not be on the same version. If we are (unreasonably) lucky the signature of the model has changed so things will crash, and if we are unlucky we will just serve data with the wrong (ton old or too new) model.  So: The client should be able (forced?) to send a project version together with the request, and be guaranteed that if it gets a 200 OK then the data was applied on exactly that version of the model. 

**Subtasks:**
- [x] Server #745: Accept a project version parameter / header whatever with requests (to predict / anomaly / metadata), and if that is provided then only return 200 OK with a result if it can be served from that exact version. I propose we try to serve old versions in a "best effort", so we can serve old models, but if we are not able to resolve a version we return some other code than 500. Maybe a `410 Gone`. 
- [X] Server have endpoint for available revisions (#775)
- [X] Server have endpoint to list models by revision (#776)
- [x] Watchman #746: Ensure that the project version field in the metadata is consistent with the metadata collected/served.
- [x] Client #747:  Send the project_version together with the data to the server. Handle the possible 410 response code by updating its metadata from watchman.

PR #766, opened 2019-12-06
Remove watchman

### LINKED ISSUE #703: RFC: Ignore name/metadata in cache-key
Should we explicitly ignore e.g. the name and the user-provided metadata when calculating the cache-key, and then rather populate these field with whatever we are provided? That will make the cache-key more tied to just the serialized model, and give more cache-hits.

### LINKED ISSUE #746: Watchman: Ensure consistency between served metadata and project revision
Part of #744

Ensure that the project version field in the metadata is consistent with the metadata collected/served.
Probably depends on #745

milesgranger:
Part of #744 and #746 

Will close #703 

Switched to using the server itself as watchman's replacement. 

Update metadata to conform to reconstruct `config_elements.machine.Machine` for more predictable traversal, and consistent sharing between processes like the client and `server-to-sql` work. 

The last commit `Unify use of to/from_ts...` can be moved into a different PR as that is simply renaming all occurrences of `from_ts` -> `train_start_date` and `to_ts` -> `train_end_date`


gordo_components/__init__.py
```diff
@@ -51,12 +51,3 @@ def _parse_version(version: str) -> Tuple[int, ...]:
...
     pass
-
-
-# Set log level, defaulting to INFO
-log_level = os.getenv("LOG_LEVEL", "INFO").upper()
-
-logging.basicConfig(
-    level=getattr(logging, log_level),
-    format="[%(asctime)s] %(levelname)s [%(name)s.%(funcName)s:%(lineno)d] %(message)s",
-)
```
flikka: Do you want this change in this PR? Seems to be related to the logging cleanup...

EinarJohnsen: Why was this removed?

milesgranger: No, you are correcto!

milesgranger: Because, I've failed to move it into a different PR. But the rub is that having it hear isn't good. For one reason if tests with use logging fail, we get an IO error caused by confliction with pytest logging because having it here overrides their logging. Anyhow, I will move this to another PR post haste

milesgranger: but, I will keep it here until #774 is merged because it's a right pain to debug when tests fail. After that's in, the diff will be gone.

flikka (commented):
Feels good to get rid of things. When this was made, I guess we had 1) All the models having their own server and 2) No gordo CRD. So it makes so much sense to thin it out.

If it all works, fine by me - some testing is probably in place, or already done perhaps? The whole thing with the Postgres and Grafana too. 

A question: Does anyone use the watchman endpoint? I don't think, except from Lennart, but I guess he uses the client? Leif Ivar might, but I can drop him a mail.

Another question: Is there a (public) endpoint that exposes the list of models and the model's metadata? A bit confused about the endpoints now :-) Has the controller replaced the watchman content, or only partially?

Yet another question: The controllers exposure of a gordo, that is now what the client uses, right? Also this is what is fed into the postgres? Does this include metadata from building too, like cross val scores, training times etc?

milesgranger:
> If it all works, fine by me - some testing is probably in place, or already done perhaps? The whole thing with the Postgres and Grafana too.

Not yet, as the talk with @epa095 yesterday eludes that perhaps we should use the server itself to see what models it's capable of serving; changing this further to query the server instead of the controller.

> A question: Does anyone use the watchman endpoint? I don't think, except from Lennart, but I guess he uses the client? Leif Ivar might, but I can drop him a mail.

I am not sure, I would expect @mrdeveloperdude does not use watchman directly.

> Another question: Is there a (public) endpoint that exposes the list of models and the model's metadata? A bit confused about the endpoints now :-) Has the controller replaced the watchman content, or only partially?

Not yet, but this would go into point one, we could extend the current server to list the models it's capable of serving if/when we go that route instead of the current standing here with controller querying.

> Yet another question: The controllers exposure of a gordo, that is now what the client uses, right? Also this is what is fed into the postgres? Does this include metadata from building too, like cross val scores, training times etc?

Not, and probably another good point for using the server instead of the controller for this control flow.

mrdeveloperdude:
I only use client like a gentleman 😇

gordo_components/__init__.py
ryanjdillon: Were these going into the cli on this PR as per #773?

gordo_components/client/forwarders.py
```diff
@@ -9,25 +9,22 @@
...
-from gordo_components.client.utils import influx_client_from_uri, EndpointMetadata
+from gordo_components.client.utils import influx_client_from_uri
+from gordo_components.workflow.config_elements.machine import Machine
 
 
 """
 Module contains objects which can be made into generators which take
-and EndpointMetadata and metadata (dict) when instantiated and are sent
+and Machine and metadata (dict) when instantiated and are sent
 prediction dataframes as the prediction client runs::
```
ryanjdillon: This is a confusing read. Maybe reword?

gordo_components/workflow/config_elements/machine.py
```diff
@@ -66,6 +66,10 @@ def from_config(  # type: ignore
     ):
         if config_globals is None:
             config_globals = dict()
+        else:
+            config_globals = config_globals.copy()
+
+        config = config.copy()
```
ryanjdillon: Myabe `config_globals_copy = ...` and `config_copy = ...` etc. for clarity.

gordo_components/workflow/controller_to_sql/controller_to_sql.py
```diff
@@ -62,21 +60,21 @@ def endpoint_to_machine_data(ep):
...
+    controller_address: str,
     sql_host: str,
     sql_port: int = 5432,
     sql_database: str = "postgres",
     sql_username: str = "postgres",
     sql_password: str = None,
 ):
     """
-    Connects to watchman, downloads all metadata, and proceeds to push them to postgres.
+    Connects to controller, downloads all metadata, and proceeds to push them to postgres.
```
ryanjdillon: This kind of makes it sound like 'all metadata' is on the controller.

tests/gordo_components/client/test_client.py
```diff
@@ -443,90 +440,52 @@ def test_ml_server_dataframe_to_dict_and_back(tags: typing.List[str]):
         assert np.allclose(df[top_lvl_name].values, df_clone[top_lvl_name].values)
 
 
+def _machine(name: str) -> Machine:
+    """
+    Helper to build a basic Machine, only defining its name
+    """
+    # TODO: Fix this from changes from EndpointMetadata being removed
```
ryanjdillon, line 443: Are these fixes still needing to be done?

tests/gordo_components/client/test_client.py
```diff
@@ -443,90 +440,52 @@ def test_ml_server_dataframe_to_dict_and_back(tags: typing.List[str]):
...
-        (
-            [_endpoint_metadata("t1", True), _endpoint_metadata("t2", False)],
-            "t2",
-            True,
-            ValueError,
-        ),
+        # Two machines, no target, should give two machines
+        ([_machine("t1"), _machine("t2")], None, [_machine("t1"), _machine("t2")]),
+        # One machine target should filter down to that machine
+        ([_machine("t1"), _machine("t2")], "t2", [_machine("t2")]),
```
ryanjdillon, line 464: Sweet. Would be cool to have some advanced querying/regex matching in the future.

ryanjdillon (dismissed):
Looks like a sweet improvement :+1: . Approving, but of coarse defaulting to the seniority approval.

gordo_components/__init__.py
milesgranger: Yep yep, just haven't rebased as it was addressed in #774

gordo_components/workflow/controller_to_sql/controller_to_sql.py
milesgranger: Resolving this, as it's not irrelevant, just that it is given the new direction this PR will go (ie. no longer querying controller)

tests/gordo_components/client/test_client.py
milesgranger: Same as previous comment, soon will not be relevant.

gordo_components/server/views/base.py
```diff
@@ -233,6 +232,19 @@ def get(self):
...
+    """
+
+    @api.doc(description="List the metadata for all models capable of being served.")
+    def get(self, gordo_project: str):
+        collection_dir = os.environ[current_app.config["MODEL_COLLECTION_DIR_ENV_VAR"]]
+        available_models = os.listdir(collection_dir)
+        return jsonify(available_models)
+
+
+api.add_resource(ModelListView, "/gordo/v0/<gordo_project>/models")
```
epa095, line 247: Should we just start putting in the revision here already? E.g.  `/gordo/v0/<gordo_project>/<revision>/models`, where `latest` is a legal revision?

milesgranger: I could add that. The PR doesn't have enough changes for you already? :stuck_out_tongue_winking_eye:

milesgranger: Actually, perhaps it may be best to forego this into another PR. Adding this will require updates in the tests to a somewhat larger degree. We'll have to train more than one model for different revisions and all that.

gordo_components/builder/build_model.py
```diff
@@ -233,9 +233,9 @@ def _build(self) -> Tuple[Optional[sklearn.base.BaseEstimator], dict]:
             metrics_list = self.metrics_from_list(self.evaluation_config.get("metrics"))
 
             # Cross validate
-            logger.debug("Starting cross validation")
-            start = time.time()
             if hasattr(model, "predict"):
+                logger.debug("Starting cross validation")
```
epa095, line 245: :+1:

gordo_components/builder/build_model.py
```diff
@@ -268,44 +268,43 @@ def _build(self) -> Tuple[Optional[sklearn.base.BaseEstimator], dict]:
...
                     scores.update({metric: val})
+
+                cv_duration_sec = time.time() - start
             else:
                 logger.debug("Unable to score model, has no attribute 'predict'.")
 
-            cv_duration_sec = time.time() - start
-
-            # If cross_val_only, return the cv_scores and empty model.
-            if self.evaluation_config["cv_mode"] == "cross_val_only":
```
epa095, line 277: is this gone on purpose? Seems like it is no longer possible to only do CV without building the whole model?

gordo_components/builder/local_build.py
```diff
@@ -75,6 +74,8 @@ def local_build(
     Iterable[Tuple[Union[BaseEstimator, None], dict]]
         A generator yielding tuples of models and their metadata.
     """
+    from gordo_components.builder.mlflow_utils import mlflow_context, log_metadata
```
epa095, line 77: This is a new style, any reason why we are doing this only(?) here and not everywhere? Are these imports particularly problematic?

gordo_components/cli/client.py
```diff
@@ -29,7 +29,6 @@
 @click.option("--host", help="The host the server is running on", default="localhost")
 @click.option("--port", help="Port the server is running on", default=443)
 @click.option("--scheme", help="tcp/http/https", default="https")
-@click.option("--gordo-version", help="Version of gordo", default="v0")
```
epa095, line 32: Why remove this?

gordo_components/client/client.py
```diff
@@ -125,90 +120,92 @@ def __init__(
...
-            raise IOError(f"Failed to get endpoints: {repr(resp.content)}")
-        return [EndpointMetadata(data) for data in resp.json()["endpoints"]]
+            raise IOError(f"Failed to get machines: {repr(resp.content)}")
+        else:
+            model_names = resp.json()
+            with ThreadPoolExecutor(max_workers=self.parallelism) as executor:
+                machines = executor.map(self._machine_from_metadata, model_names)
+                return list(machines)
+
+    def _machine_from_metadata(self, name: str) -> Machine:
```
epa095: Wellwel, both the name and docstring is maybe a bit missleading, since it actually *fetches* the fricking metadata

gordo_components/builder/local_build.py
milesgranger: OH ya! The test won't even run if it's where it was... :man_shrugging: ML flow confusion.

gordo_components/client/client.py
```diff
@@ -125,90 +120,92 @@ def __init__(
...
         """
-        resp = self.session.get(endpoint)
+        resp = self.session.get(f"{self.server_endpoint}/models")
         if not resp.ok:
-            raise IOError(f"Failed to get endpoints: {repr(resp.content)}")
-        return [EndpointMetadata(data) for data in resp.json()["endpoints"]]
+            raise IOError(f"Failed to get machines: {repr(resp.content)}")
+        else:
+            model_names = resp.json()
+            with ThreadPoolExecutor(max_workers=self.parallelism) as executor:
```
epa095, line 181: Probably makes sense to have this customizable as seperate parallelism, as fetching metadata is quite different from applying predictions.

epa095: But we can do that later

gordo_components/builder/build_model.py
milesgranger: eh, ya.. that's true. It was a bit confusing getting the metadata built up in an intermediate fashion, but I overlooked the fact this would result in building the whole model. I fix, I fix.

gordo_components/cli/client.py
milesgranger: because, as I saw it, running a given version of gordo client against a different major version would intrinsically be a bad idea. Maybe I am wrong in that assumption.

gordo_components/model/anomaly/diff.py
```diff
@@ -297,6 +297,7 @@ def anomaly(
             # Dataframe of % abs_diff is of the thresholds
             anomaly_confidence_scores = pd.DataFrame(
                 confidence,
+                index=tag_anomaly_scaled.index,
```
epa095, line 300: this came a bit out of the blue, is it related to the other stuff or just another bug you squashed while walking around?

gordo_components/workflow/server_to_sql/server_to_sql.py
```diff
@@ -0,0 +1,110 @@
...
+    """
+    address: ParseResult = urlparse(server_address)
+
+    client = Client(
+        project=project_name,
+        host=address.netloc,
+        port=address.port or 443 if address.scheme == "https" else 80,
+        scheme=address.scheme,
+    )
+    return client.machines
```
epa095, line 120: Reasonable. Reasonable

resources/grafana/dashboards/machines.json
```diff
@@ -941,8 +941,8 @@
             "measurement": "total-anomaly",
             "orderByTime": "ASC",
             "policy": "default",
-            "query": "SELECT mean(/^$Sensor$/) as \"field\" FROM \"original-input\" WHERE (\"machine\" =~ /^$Machine$/) AND $timeFilter GROUP BY time($__interval) fill(null)",
-            "rawQuery": false,
+            "query": "SELECT mean(*) FROM \"total-anomaly-scaled\" WHERE (\"machine\" =~ /^$Machine$/) AND $timeFilter GROUP BY time($__interval) fill(null)",
```
epa095: IDK if this is intended. Talk to @flikka , he is an expert on grafana-influx-query-shitups

setup.cfg
```diff
@@ -19,7 +19,6 @@ testmodel        = pytest --addopts "tests/gordo_components/model"
 testserializer   = pytest --addopts "tests/gordo_components/serializer"
 testserver       = pytest --addopts "tests/gordo_components/server"
 testutil         = pytest --addopts "tests/gordo_components/util"
-testwatchman     = pytest --addopts "tests/gordo_components/watchman"
```
epa095, line 22: :tada:

gordo_components/model/anomaly/diff.py
milesgranger: Yessir, there be monsters lurking in this detector; what with all the strange calculations. :sweat_smile:

resources/grafana/dashboards/machines.json
milesgranger: That's probably on me, as I merely _looked_ at a query in influx and broke this. :laughing:

gordo_components/workflow/config_elements/machine.py
```diff
@@ -106,6 +106,9 @@ def from_config(  # type: ignore
     def __str__(self):
         return yaml.dump(self.to_dict())
 
+    def __eq__(self, other):
```
epa095, line 109: :+1:

gordo_components/workflow/server_to_sql/server_to_sql.py
milesgranger: So pluggable!

gordo_components/workflow/config_elements/validators.py
```diff
@@ -239,7 +239,7 @@ def __set__(self, instance, value):
         if (
             len(value) == 0
             or not isinstance(value, list)
-            or not isinstance(value[0], str)
+            or not any(isinstance(value[0], inst) for inst in (str, dict))
```
epa095: sensortag ?

gordo_components/workflow/server_to_sql/server_to_sql.py
```diff
@@ -0,0 +1,110 @@
...
+                machine_config_to_machine_data(machine)
+            ).on_conflict_ignore().execute()
+    return True
+
+
+def get_machines_from_server(
+    project_name: str, server_address: str
+) -> List[MachineConfig]:
+    """
+    TODO: Parameters
```
epa095: TODO still? Now you have your name on this, so its your problem!!

tests/gordo_components/builder/test_builder.py
```diff
@@ -456,7 +469,7 @@ def test_provide_saved_model_caching_handle_existing_different_register(tmp_dir)
     "should_be_equal,metadata,tag_list,replace_cache",
     [
         (True, None, None, False),
-        (False, {"metadata": "something"}, None, False),
+        (True, {"metadata": "something"}, None, False),
```
epa095, line 472: wait, what happened here? Is the ` {"metadata": "something"}` available in `second_metadata` ? Because it should be

milesgranger: Because we no longer count metadata as part of the cache key, right? Or at least that's what we were _going_ to do, and anyway the addition of `build-metadata` as a subkey into `metadata` would break this, so it worked nicely with the removal of the metadata as part of the cache key.

epa095: Closing #703 is nice, but then I propose you also add a test (or change this) to verify that even when the model is cached the second loading of the model gets the user-specified metadata attached as expected.

gordo_components/workflow/server_to_sql/server_to_sql.py
```diff
@@ -0,0 +1,109 @@
...
+    sql_host: str,
+    sql_port: int = 5432,
+    sql_database: str = "postgres",
+    sql_username: str = "postgres",
+    sql_password: str = None,
+):
+    """
+    Connects to server, downloads all metadata, and proceeds to push them to postgres.
+
+    Parameters
```
ryanjdillon, line 56: Periods after all parameter definitions here.

gordo_components/workflow/server_to_sql/server_to_sql.py
```diff
@@ -0,0 +1,109 @@
...
+            logger.debug(f"Inserting machine {machine['name']} in sql")  # type: ignore
+            Machine.insert(machine).on_conflict_ignore().execute()
+    return got_all_machines
+
+
+def get_machines_from_server(
+    project_name: str, server_address: str
+) -> List[MachineConfig]:
+    """
+    TODO: Parameters
```
ryanjdillon: What Erik said.

tests/gordo_components/workflow/test_server_to_sql/test_server_to_sql.py
```diff
@@ -0,0 +1 @@
+# TODO: Add relevant tests
```
ryanjdillon: :laughing:

tests/gordo_components/workflow/test_server_to_sql/test_server_to_sql.py
```diff
@@ -1 +1,10 @@
-# TODO: Add relevant tests
```
ryanjdillon: I see what's going on here.

tests/gordo_components/workflow/test_server_to_sql/test_server_to_sql.py
```diff
@@ -1 +1,10 @@
...
+import tests.utils as tu
+from gordo_components.workflow.server_to_sql import server_to_sql as sts
+
+
+def test_get_machines_from_server(ml_server):
+    machines = sts.get_machines_from_server(
```
ryanjdillon, line 6: This seems to unnecessarily add a bit of strain looking over there tests. I don't know what the other's opinions are on this namespacing stuff, but I am not really a fan of `np` or `pd` either.

gordo_components/workflow/server_to_sql/server_to_sql.py
```diff
@@ -56,7 +56,10 @@ def server_to_sql(
 
     Parameters
     ----------
-    TODO: FIll in missing params
+    project_name: str
+        Name of the project to run against.
+    server_address:
+        Address of the server to connect to. ie. https://myhost.com
     sql_host: str
         Host where postgres resides, e.g. "localhost"
```
ryanjdillon: e.g. - exempli gratia ‘for the sake of example’ ... used when giving an example.
i.e. - id est ‘that is’ ... i.e. 'to put it another way' or 'specifically the thing I am referring to'

And punktums here and below.

gordo_components/workflow/server_to_sql/server_to_sql.py
```diff
@@ -36,9 +35,9 @@ def __repr__(self):
 def machine_config_to_machine_data(machine: MachineConfig):
     return dict(
         name=machine.name,
-        metadata=machine.metadata["machine-metadata"],
-        train_start_date=machine.dataset.train_start_date,
-        train_end_date=machine.dataset.train_end_date,
+        dataset=machine.dataset.to_dict(),
+        model=machine.model,
+        metadata=machine.metadata,
```
ryanjdillon: I like this. That makes much more sense.

gordo_components/builder/build_model.py
```diff
@@ -223,6 +223,13 @@ def _build(self) -> Tuple[Optional[sklearn.base.BaseEstimator], dict]:
 
         cv_duration_sec = None
 
+        metadata: Dict[Any, Any] = dict(
```
ryanjdillon, line 227: Will the metadata keys ever not be strings?

gordo_components/workflow/server_to_sql/server_to_sql.py
```diff
@@ -97,7 +97,18 @@ def get_machines_from_server(
...
     """
-    TODO: Parameters
+    Get a list of machines from the server
+
+    Parameters
+    ----------
+    project_name: str
+        Project name to query machines for
+    server_address: str
+        The address of the server ie. https://myhost.com
```
ryanjdillon: *i.e.* to *e.g.* and periods

tests/gordo_components/workflow/test_server_to_sql/test_server_to_sql.py
milesgranger: What's that?

tests/gordo_components/workflow/test_server_to_sql/test_server_to_sql.py
milesgranger: Looking at 'sts' is more straining than a free function? To me, much like the well accepted practice of shortnaming pandas and numpy, looking at a prefix immediately shows me this function is from another module. Thus I don't have to think "is this a function defined in the current file?"

gordo_components/workflow/server_to_sql/server_to_sql.py
milesgranger: Thanks for the grammar lesson.

tests/gordo_components/workflow/test_server_to_sql/test_server_to_sql.py
milesgranger: This was because there was a single test in here, but the function it tested was removed from refactoring, and no other functions in `server_to_sql` had tests to begin with.

gordo_components/builder/build_model.py
```diff
@@ -516,7 +528,6 @@ def cache_key(self) -> str:
                 "name": self.name,
                 "model_config": self.model_config,
                 "data_config": self.data_config,
-                "user-defined": self.metadata,
```
epa095, line 519: Should also remove the  `if self.metadata is None:`... above since it was only there because of this line you now removed.

gordo_components/builder/build_model.py
```diff
@@ -223,6 +224,13 @@ def _build(self) -> Tuple[Optional[sklearn.base.BaseEstimator], dict]:
 
         cv_duration_sec = None
 
+        metadata: Dict[Any, Any] = dict(
+            name=self.name,
+            dataset=dataset.get_metadata(),
+            metadata={"user-metadata": self.metadata},
```
epa095: Before it was `user-defined`, now it is `user-metadata`. On purpose?

tests/gordo_components/builder/test_builder.py
```diff
@@ -519,45 +532,19 @@ def test_provide_saved_model_caching(
...
-        model_config=model_config,
-        data_config=data_config,
-        evaluation_config=evaluation_config,
-    ).build()
-    if should_be_equal:
-        assert model is not None
-    else:
-        assert model is None
+    if metadata is not None:
+        assert str(metadata) in str(second_metadata)
```
epa095: Can you, to make me feel just a tiny bit safer, have this test that `metadata` is at the expected place in `second_metadata`? Which I guess is `second_metadata["metadata"]["user-metadata"]` ?

epa095 (commented):
looking gooder

gordo_components/builder/build_model.py
milesgranger: ah, no, oversight. I'll check it back

tests/gordo_components/builder/test_builder.py
milesgranger: Safety third!

gordo_components/builder/build_model.py
milesgranger: Nice catch! :baseball:

epa095 (approved):
When tests pass I am happy!

milesgranger:
Cool, I'll start one more run through the test project to be sure things are still ok.

PR #772, opened 2019-12-09
Handle empty timeseries input to GordoDataset

### LINKED ISSUE #686: join_timeseries: handle empty timeseries
Currently we crash with 
```
  File "/usr/local/lib/python3.6/site-packages/gordo_components/dataset/base.py", line 77, in join_timeseries
    tz=series.index[0].tzinfo
  File "/usr/local/lib/python3.6/site-packages/pandas/core/indexes/datetimes.py", line 1220, in __getitem__
    result = self._data.__getitem__(key)
  File "/usr/local/lib/python3.6/site-packages/pandas/core/arrays/datetimelike.py", line 426, in __getitem__
    val = getitem(key)
IndexError: index 0 is out of bounds for axis 0 with size 0
```
https://github.com/equinor/gordo-components/blob/cc942329f58d0721216267d158ca5d52879f9cf8/gordo_components/dataset/base.py#L77

Maybe we should either raise a better error, or handle it so we return an empty dataframe (which is maybe the right thing to do?). If we end up doing the latter we should probably make sure we have some tests somewhere so we dont end up training models on no data.

ryanjdillon:
This PR does the following:
* Converts `tests/gordo_components/dataset/test_dataset.py` to pytest format
* Raises a new `InsufficientDataError` if not enough row in a sensor `Series`.

closes #686

### DIFF
--- a/gordo_components/dataset/datasets.py
+++ b/gordo_components/dataset/datasets.py
@@ -21,6 +21,10 @@ from gordo_components.dataset.sensor_tag import normalize_sensor_tags
 logger = logging.getLogger(__name__)
 
 
+class InsufficientDataError(ValueError):
+    pass
+
+
 def compat(init):
     """
     __init__ decorator for compatibility where the Gordo config file's ``dataset`` keys have
@@ -61,6 +65,7 @@ class TimeSeriesDataset(GordoBaseDataset):
         row_filter_buffer_size: int = 0,
         asset: Optional[str] = None,
         default_asset: Optional[str] = None,
+        n_samples_threshold: int = 0,
         **_kwargs,
     ):
         """
@@ -111,6 +116,8 @@ class TimeSeriesDataset(GordoBaseDataset):
         default_asset: Optional[str]
             Asset which will be used if `asset` is not provided and the tag is not
             resolvable to a specific asset.
+        n_samples_threshold: int = 0
+            The threshold at which the generated DataFrame is considered to have too few rows of data.
         _kwargs
         """
         self.from_ts = self._validate_dt(from_ts)
@@ -131,6 +138,7 @@ class TimeSeriesDataset(GordoBaseDataset):
         self.aggregation_methods = aggregation_methods
         self.row_filter_buffer_size = row_filter_buffer_size
         self.asset = asset
+        self.n_samples_threshold = n_samples_threshold
 
         if not self.from_ts.tzinfo or not self.to_ts.tzinfo:
             raise ValueError(
@@ -157,15 +165,26 @@ class TimeSeriesDataset(GordoBaseDataset):
 
         # Resample if we have a resolution set, otherwise simply join the series.
         if self.resolution:
-            data = self.join_timeseries(
-                series_iter,
-                self.from_ts,
-                self.to_ts,
-                self.resolution,
-                aggregation_methods=self.aggregation_methods,
-            )
+            try:
+                data = self.join_timeseries(
+                    series_iter,
+                    self.from_ts,
+                    self.to_ts,
+                    self.resolution,
+                    aggregation_methods=self.aggregation_methods,
+                )
+            except IndexError:
+                raise InsufficientDataError(
+                    f"One or more series in the tag list is missing data for the specified period."
+                )
         else:
             data = pd.concat(series_iter, axis=1, join="inner")
+
+        if len(data) <= self.n_samples_threshold:
+            raise InsufficientDataError(
+                f"The length of the generated DataFrame ({len(data)}) does not exceed the "
+                f"specified required threshold for number of rows ({self.n_samples_threshold})."
+            )
 
         if self.row_filter:
             data = pandas_filter_rows(
--- a/gordo_components/cli/cli.py
+++ b/gordo_components/cli/cli.py
@@ -11,6 +11,7 @@ import traceback
 
 from gordo_components.data_provider.providers import NoSuitableDataProviderError
 from gordo_components.dataset.sensor_tag import SensorTagNormalizationError
+from gordo_components.dataset.datasets import InsufficientDataError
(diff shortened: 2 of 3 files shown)


gordo_components/dataset/base.py
```diff
@@ -73,6 +74,12 @@ def join_timeseries(
         resampled_series = []
 
         for series in series_iterable:
+            if not series.size:
+                dt_index = pd.date_range(
+                    start=resampling_startpoint, end=resampling_endpoint, freq="24H"
+                )
+                series = pd.Series([np.NaN] * len(dt_index), index=dt_index, tz=UTC)
```
epa095: Nha, it needs to be empty instead. Note as the documentation to this function says, it is guaranteed to return a dataframe without NaNs.

gordo_components/dataset/base.py
```diff
@@ -73,6 +74,12 @@ def join_timeseries(
         resampled_series = []
 
         for series in series_iterable:
+            if not series.size:
```
epa095: I guess you rely on 0 beeing false here? I must admit I prefer explicit check like `==0` if it is guaranteed to be 0. Or can it sometimes be None?

epa095 (commented):
good start :-)

gordo_components/dataset/base.py
```diff
@@ -70,78 +79,87 @@ def join_timeseries(
...
+                    f"Series {repr(series.name)} size ({series.size}) is less than or equal to the minimum required ({row_threshold})."
                 )
 
-            elif series.index[0] < resampling_startpoint:
-                msg = (
-                    f"Error - for {series.name}, first timestamp "
-                    f"{series.index[0]} is before the resampling start point "
-                    f"{startpoint_sametz}"
+            # Resample if we have a resolution set, otherwise simply join the series.
+            if bool(resolution):
```
milesgranger: A small thing, but what case are we trying to cover by explicitly calling `bool` here?

gordo_components/dataset/base.py
```diff
@@ -70,78 +79,87 @@ def join_timeseries(
...
+                    resampled.columns = pd.MultiIndex.from_product(
+                        [[series.name], resampled.columns],
+                        names=["tag", "aggregation_method"],
+                    )
+                filled = resampled.fillna(method="ffill")
+                _series_iterable.append(filled)
+            else:
+                _series_iterable.append(series)
+
+        data = pd.concat(_series_iterable.__iter__(), axis=1, join="inner")
```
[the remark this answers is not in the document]
ryanjdillon: :+1:  That's a space-cadet-residual I missed after debugging something.

gordo_components/dataset/datasets.py
```diff
@@ -155,17 +155,13 @@ def get_data(self) -> Tuple[pd.DataFrame, Optional[pd.DataFrame]]:
...
-            data = self.join_timeseries(
-                series_iter,
-                self.from_ts,
-                self.to_ts,
-                self.resolution,
-                aggregation_methods=self.aggregation_methods,
-            )
-        else:
-            data = pd.concat(series_iter, axis=1, join="inner")
+        data = self.join_timeseries(
```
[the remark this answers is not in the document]
ryanjdillon: I personally like this in that it keeps all of the series join logic in one method. The nan dropping I can move within the `if resolution` block.

gordo_components/dataset/base.py
ryanjdillon: :+1:  That's another residual from debugging this weird error I was getting.

gordo_components/dataset/base.py
ryanjdillon: After some discussion with Miles, I'm now catching the index error in `base.GordoDataset.get_data`, so this has been removed.

gordo_components/dataset/datasets.py
```diff
@@ -61,6 +65,7 @@ def __init__(
         row_filter_buffer_size: int = 0,
         asset: Optional[str] = None,
         default_asset: Optional[str] = None,
+        row_threshold: int = 0,
```
milesgranger: Could we rename this to something like `n_samples_threshold` or something?

gordo_components/dataset/datasets.py
```diff
@@ -157,13 +165,23 @@ def get_data(self) -> Tuple[pd.DataFrame, Optional[pd.DataFrame]]:
...
+                    self.from_ts,
+                    self.to_ts,
+                    self.resolution,
+                    aggregation_methods=self.aggregation_methods,
+                )
+                if len(data) <= self.row_threshold:
+                    raise InsufficientDataError(
+                        f"The length of the generated DataFrame ({len(data)}) does not exceed the "
+                        f"specified required threshold for number of rows ({self.row_threshold})."
+                    )
```
milesgranger: I was thinking rather this should be outside of here and after the possible joining without resolution. That way both are checked.

milesgranger (commented):
I like this approach better then the previous work. :+1:

gordo_components/dataset/datasets.py
```diff
@@ -21,6 +21,10 @@
 logger = logging.getLogger(__name__)
 
 
+class InsufficientDataError(ValueError):
```
epa095, line 24: Also remember to add it to this handy map: https://github.com/equinor/gordo-components/blob/50d5a9e96550b2c9c4eda02bcc07c6f725da7cca/gordo_components/cli/cli.py#L31

ryanjdillon: Oooh... right.

gordo_components/dataset/datasets.py
ryanjdillon: OK, got it. See what you think now :+1:

milesgranger (approved):
Nice! :clap:  Not sure if @epa095 has any extra comments or not though.

epa095 (approved):
LGTM

PR #774, opened 2019-12-10
Move logging config to CLI entry point

### LINKED ISSUE #773: Move logging basicConfig setup into the cli
Turns out that we should not set up logging on import (as we do today), but rather on program startup. The easiest is probably to make a function doing the configuration, and hooking click up to call it when anything is run.

milesgranger:
Will close #773 

Problem :clinking_glasses: 
---
As it stands now, tests which fail while attempting to log _after_ the test fails, like anything which uses the `influxdb` feature will create a massive stack trace of logging io errors. Causes by us setting the logging config on import when it should rather be set on some process, like our CLI.

Solution :milk_glass: 
---
So we move it into the CLI entry point so logging will only be configured to our desired level when using the project via CLI.

### DIFF
--- a/gordo_components/__init__.py
+++ b/gordo_components/__init__.py
@@ -51,12 +51,3 @@ try:
 except Exception:
     warnings.warn(f"Failed to fix absl logging bug {traceback.format_exc()}")
     pass
-
-
-# Set log level, defaulting to INFO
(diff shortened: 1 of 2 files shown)


epa095 (approved):
LGTM

PR #781, opened 2019-12-10
Cache 25000 metadata entries in the server

epa095:
Uses pickle and compression to lower the memory requirements.

### DIFF
--- a/gordo_components/server/utils.py
+++ b/gordo_components/server/utils.py
@@ -2,8 +2,11 @@
 
 import logging
 import functools
+import zlib
 import os
 import io
+import pickle
+
 import dateutil
 import timeit
 from datetime import datetime
@@ -337,7 +340,6 @@ def load_model(directory: str, name: str) -> BaseEstimator:
     return model
 
 
-@lru_cache(maxsize=20)
 def load_metadata(directory: str, name: str) -> dict:
     """
     Load metadata from a directory for a given model by name.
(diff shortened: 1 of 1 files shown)


gordo_components/server/utils.py
```diff
@@ -353,9 +355,22 @@ def load_metadata(directory: str, name: str) -> dict:
...
     -------
     dict
+    """
+    compressed_metadata = _load_compressed_metadata(directory, name)
+    return pickle.loads(zlib.decompress(compressed_metadata))
+
+
+@lru_cache(maxsize=25000)
+def _load_compressed_metadata(directory: str, name: str):
+    """Loads the metadata and returns it as a lzma compressed pickle, to use
```
milesgranger: Very small point, most of the time we have a return under the `"""`  but the main thing is if we want to do our numpy docs here?

milesgranger (commented):
Is there an echo in here? _"Maybe we compress the metadata..."_ :stuck_out_tongue_winking_eye:

gordo_components/server/utils.py
epa095: Tja, IDK. Since it is an hidden function I think it is less important to have a full documentation, but it is nice to follow the same style as the rest, so I tried to make it numpy compliant.

milesgranger (approved):
Neat! :sparkles:

Issue #649, opened 2019-11-13
Parameter logging during model training


henrikny:
These are the different model parameters we want to log under training.

### Metrics
For each scaled variable and for each time series cross-validation split:
- [x] Mean squared errors
- [x] Explained variance
- [x] R^2

### Metadata
- [x] Start and end date for each fold.
- [x] Gordo model name (tilstandomatic_model_id + tilstandomatic_model_tag) = model_id
- [x] Unique model version (model_id + hash of model config?)
- [x] Complete config information (parsed yaml)
- [x] Data query duration
- [x] Training time duration

ryanjdillon:
Start and end date to CV folds added in #785, so this one should be all wrapped up. Closing.

PR #785, opened 2019-12-11
Add cv fold timestamps and n points to metadata

### LINKED ISSUE #783: Start and end date for each fold
As per one of the ticks in #649, this issue hereby pertains to the adding of start and end timestamps for each dataset used in each fold!

ryanjdillon:
# Problem :fire: :european_castle:  :fire: :dragon_face: 
Folks want some information on the start/end of each CV fold

# Solution :bow_and_arrow:  :crown: :crossed_swords: 
Add the following to the metadata:
* start/end timestamps per set (train, test), per fold
* n_samples per set (train, test), per fold

closes #783
also the last tickbox on #649

### DIFF
--- a/gordo_components/builder/build_model.py
+++ b/gordo_components/builder/build_model.py
@@ -10,7 +10,7 @@ import os
 import time
 import random
 from pathlib import Path
-from typing import Union, Optional, Dict, Any, Tuple, List, Callable
+from typing import Union, Optional, Dict, Any, Tuple, Type, List, Callable
 
 import pandas as pd
 import numpy as np
@@ -19,7 +19,7 @@ import tensorflow as tf
 import sklearn
 from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn import metrics
-from sklearn.model_selection import cross_validate, TimeSeriesSplit
+from sklearn.model_selection import BaseCrossValidator, cross_validate, TimeSeriesSplit
 from sklearn.pipeline import Pipeline
 
 from gordo_components.util import disk_registry
@@ -186,6 +186,7 @@ class ModelBuilder:
             evaluation=self.machine.evaluation,
         )
 
+        split_metadata: Dict[str, Any] = dict()
         scores: Dict[str, Any] = dict()
         if self.machine.evaluation["cv_mode"].lower() in (
             "cross_val_only",
@@ -205,12 +206,12 @@ class ModelBuilder:
                 scaler = self.machine.evaluation.get("scoring_scaler")
                 metrics_dict = self.build_metrics_dict(metrics_list, y, scaler=scaler)
 
+                # Generate metadata about CV train, test splits
+                split_obj = TimeSeriesSplit(n_splits=3)
+                split_metadata = ModelBuilder.build_split_dict(X, split_obj)
+
                 cv_kwargs = dict(
-                    X=X,
-                    y=y,
-                    scoring=metrics_dict,
-                    return_estimator=True,
-                    cv=TimeSeriesSplit(n_splits=3),
+                    X=X, y=y, scoring=metrics_dict, return_estimator=True, cv=split_obj
                 )
                 if hasattr(model, "cross_validate"):
                     cv = model.cross_validate(**cv_kwargs)
@@ -245,6 +246,7 @@ class ModelBuilder:
                         "cross-validation": {
                             "cv-duration-sec": cv_duration_sec,
                             "scores": scores,
+                            "splits": split_metadata,
                         }
                     }
                 }
@@ -269,6 +271,7 @@ class ModelBuilder:
                 "cross-validation": {
                     "cv-duration-sec": cv_duration_sec,
                     "scores": scores,
+                    "splits": split_metadata,
                 },
             }
         )
@@ -284,12 +287,42 @@ class ModelBuilder:
         random.seed(seed)
 
     @staticmethod
+    def build_split_dict(X: pd.DataFrame, split_obj: Type[BaseCrossValidator]) -> dict:
+        """
+        Get dictionary of cross-validation training dataset split metadata
+
+        Parameters
+        ----------
+        X: pd.DataFrame
+            The training dataset that will be split during cross-validation.
+        split_obj: Type[sklearn.model_selection.BaseCrossValidator]
+            The cross-validation object that returns train, test indices for splitting.
+
+        Returns
+        -------
+        split_metadata: Dict[str,Any]
+            Dictionary of cross-validation train/test split metadata
+        """
+        split_metadata: Dict[str, Any] = dict()
+        for i, (train_ind, test_ind) in enumerate(split_obj.split(X)):
+            split_metadata.update(
+                {
+                    f"fold-{i+1}-train-start": X.index[train_ind[0]],
+                    f"fold-{i+1}-train-end": X.index[train_ind[-1]],
+                    f"fold-{i+1}-test-start": X.index[test_ind[0]],
+                    f"fold-{i+1}-test-end": X.index[test_ind[-1]],
+                }
+            )
+            split_metadata.update({f"fold-{i+1}-n-train": len(train_ind)})
+            split_metadata.update({f"fold-{i+1}-n-test": len(test_ind)})
+        return split_metadata
+
+    @staticmethod
     def build_metrics_dict(
         metrics_list: list,
         y: pd.DataFrame,
         scaler: Optional[Union[TransformerMixin, str]] = None,
     ) -> dict:
-
         """
         Given a list of metrics that accept a true_y and pred_y as inputs this returns a
         dictionary with keys in the form '{score}-{tag_name}' for each given target tag
--- a/gordo_components/builder/mlflow_utils.py
+++ b/gordo_components/builder/mlflow_utils.py
@@ -262,11 +262,20 @@ def get_batch_kwargs(metadata: dict) -> dict:
     model_keys = ["model-creation-date", "model-builder-version", "model-offset"]
     param_list += get_params(build_metadata["model"], model_keys)
 
+    # Parse cross-validation split metadata
+    try:
+        splits = build_metadata["model"]["cross-validation"]["splits"]
+        param_list += get_params(splits, splits.keys())
+    except KeyError:
+        logger.debug(
+            "Key 'metadata.model.cross-validation.splits' not found found in metadata."
+        )
+
     # Parse cross-validation metrics
-    tag_list = normalize_sensor_tags(
-        metadata["dataset"]["tag_list"], asset=metadata["dataset"]["asset"]
-    )
-    if build_metadata["model"].get("cross-validation", {}).get("scores", None):
+    try:
+        tag_list = normalize_sensor_tags(
+            metadata["dataset"]["tag_list"], asset=metadata["dataset"]["asset"]
+        )
         scores = build_metadata["model"]["cross-validation"]["scores"]
         keys = sorted(list(scores.keys()))
         subkeys = ["mean", "max", "min", "std"]
@@ -280,16 +289,20 @@ def get_batch_kwargs(metadata: dict) -> dict:
             # Summary stats per metric
             for sk in subkeys:
                 metric_list.append(
-                    Metric(f"{k}_{sk}", scores[k][f"fold-{sk}"], epoch_now(), 0)
+                    Metric(f"{k}-{sk}", scores[k][f"fold-{sk}"], epoch_now(), 0)
                 )
             # Append value for each fold with increasing steps
             metric_list += [
                 Metric(k, scores[k][f"fold-{i+1}"], epoch_now(), i)
                 for i in range(n_folds)
             ]
+    except KeyError:
+        logger.debug(
+            "Key 'metadata.model.cross-validation.scores' not found found in metadata."
+        )
 
     # Parse fit metrics
-    if build_metadata["model"].get("history", None):
+    try:
         meta_params = build_metadata["model"]["history"]["params"]
 
         metric_list += get_metrics(
@@ -305,6 +318,8 @@ def get_batch_kwargs(metadata: dict) -> dict:
                 Metric(m, float(x), timestamp=epoch_now(), step=i)
                 for i, x in enumerate(data)
             ]
+    except KeyError:
+        logger.debug("Key 'metadata.model.history.params' not found found in metadata.")
 
     return {"metrics": metric_list, "params": param_list}
 
--- a/gordo_components/builder/mlflow_utils.py
+++ b/gordo_components/builder/mlflow_utils.py
@@ -265,11 +265,12 @@ def get_batch_kwargs(metadata: dict) -> dict:
     # Parse cross-validation split metadata
     try:
         splits = build_metadata["model"]["cross-validation"]["splits"]
-        param_list += get_params(splits, splits.keys())
     except KeyError:
         logger.debug(
-            "Key 'metadata.model.cross-validation.splits' not found found in metadata."
+            "Key 'build-metadata.model.cross-validation.splits' not found found in metadata."
         )
+    else:
+        param_list += get_params(splits, splits.keys())
 
     # Parse cross-validation metrics
     try:
@@ -277,6 +278,11 @@ def get_batch_kwargs(metadata: dict) -> dict:
             metadata["dataset"]["tag_list"], asset=metadata["dataset"]["asset"]
         )
         scores = build_metadata["model"]["cross-validation"]["scores"]
+    except KeyError:
+        logger.debug(
+            "Key 'build-metadata.model.cross-validation.scores' not found found in metadata."
+        )
+    else:
         keys = sorted(list(scores.keys()))
         subkeys = ["mean", "max", "min", "std"]
 
@@ -296,15 +302,15 @@ def get_batch_kwargs(metadata: dict) -> dict:
                 Metric(k, scores[k][f"fold-{i+1}"], epoch_now(), i)
                 for i in range(n_folds)
             ]
-    except KeyError:
-        logger.debug(
-            "Key 'metadata.model.cross-validation.scores' not found found in metadata."
-        )
 
     # Parse fit metrics
     try:
         meta_params = build_metadata["model"]["history"]["params"]
-
+    except KeyError:
+        logger.debug(
+            "Key 'build-metadata.model.history.params' not found found in metadata."
+        )
+    else:
         metric_list += get_metrics(
             build_metadata["model"],
             ["data-query-duration-sec", "model-training-duration-sec"],
@@ -318,8 +324,6 @@ def get_batch_kwargs(metadata: dict) -> dict:
                 Metric(m, float(x), timestamp=epoch_now(), step=i)
                 for i, x in enumerate(data)
             ]
-    except KeyError:
-        logger.debug("Key 'metadata.model.history.params' not found found in metadata.")
 
     return {"metrics": metric_list, "params": param_list}
 
--- a/tests/gordo_components/builder/test_mlflow_utils.py
+++ b/tests/gordo_components/builder/test_mlflow_utils.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+import datetime
 
 import mlflow
 from mlflow.entities import Metric, Param
@@ -166,9 +166,9 @@ def test_get_run_id_external_calls(
 @pytest.mark.parametrize(
     "x,expected",
     [
-        (datetime(1970, 1, 1), 0),
-        (datetime(1970, 1, 1, 0, 0, 1), 1000),
-        (datetime(1970, 1, 1, 0, 0, 0, 1000), 1),
+        (datetime.datetime(1970, 1, 1), 0),
+        (datetime.datetime(1970, 1, 1, 0, 0, 1), 1000),
+        (datetime.datetime(1970, 1, 1, 0, 0, 0, 1000), 1),
     ],
 )
 def test_datetime_to_ms_since_epoch(x, expected):
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -16,26 +16,32 @@ from nbconvert.exporters import PythonExporter
 
 from gordo_components.machine.dataset.data_provider.providers import DataLakeProvider
 from gordo_components.machine.dataset.datasets import TimeSeriesDataset
+from gordo_components.machine.dataset.sensor_tag import SensorTag
 
 logger = logging.getLogger(__name__)
 
+CONFIG = dict(
+    train_start_date=dateutil.parser.isoparse("2014-07-01T00:10:00+00:00"),
+    train_end_date=dateutil.parser.isoparse("2015-01-01T00:00:00+00:00"),
+    tag_list=[SensorTag(name=f"Tag {i}", asset=None) for i in range(10)],
+)
+
 
 def _fake_data():
-    data = pd.DataFrame({f"sensor_{i}": np.random.random(size=100) for i in range(10)})
+    data = pd.DataFrame(
+        {
+            f"sensor_{i}": np.random.random(size=100)
+            for i in range(len(CONFIG["tag_list"]))
+        }
+    )
     return data, data
 
 
 @mock.patch.object(TimeSeriesDataset, "get_data", return_value=_fake_data())
 def test_faked_DataLakeBackedDataset(MockDataset):
 
-    config = dict(
-        train_start_date=dateutil.parser.isoparse("2014-07-01T00:10:00+00:00"),
-        train_end_date=dateutil.parser.isoparse("2015-01-01T00:00:00+00:00"),
-        tag_list=["asgb.19ZT3950%2FY%2FPRIM", "asgb.19PST3925%2FDispMeasOut%2FPRIM"],
-    )
-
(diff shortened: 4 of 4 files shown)


gordo_components/builder/build_model.py
```diff
@@ -240,12 +241,23 @@ def _build(self) -> Tuple[Optional[sklearn.base.BaseEstimator], dict]:
                 scaler = self.evaluation_config.get("scoring_scaler")
                 metrics_dict = self.build_metrics_dict(metrics_list, y, scaler=scaler)
 
+                # Generate metadata about CV train, test splits
+                tssplit = TimeSeriesSplit(n_splits=3)
+                for i, (train_ind, test_ind) in enumerate(tssplit.split(X)):
+                    prefix = f"fold-{i+1}"
+                    splits.update(
+                        {
+                            f"{prefix}_train_start": X.index[train_ind[0]].isoformat(),
```
milesgranger: I don't know if the indexes are _always_ going to be datetime instances, we definitely don't assume that in other places

gordo_components/builder/build_model.py
```diff
@@ -240,12 +241,23 @@ def _build(self) -> Tuple[Optional[sklearn.base.BaseEstimator], dict]:
...
+                            f"{prefix}_train_end": X.index[train_ind[-1]].isoformat(),
+                            f"{prefix}_test_start": X.index[test_ind[0]].isoformat(),
+                            f"{prefix}_test_end": X.index[test_ind[-1]].isoformat(),
+                            f"{prefix}_n_train": len(train_ind),
+                            f"{prefix}_n_test": len(test_ind),
+                        }
+                    )
+
```
milesgranger: The method is already quite long, what do you think about putting this into a different method/function?

gordo_components/builder/build_model.py
```diff
@@ -240,12 +241,23 @@ def _build(self) -> Tuple[Optional[sklearn.base.BaseEstimator], dict]:
...
+                for i, (train_ind, test_ind) in enumerate(tssplit.split(X)):
+                    prefix = f"fold-{i+1}"
+                    splits.update(
+                        {
+                            f"{prefix}_train_start": X.index[train_ind[0]].isoformat(),
+                            f"{prefix}_train_end": X.index[train_ind[-1]].isoformat(),
+                            f"{prefix}_test_start": X.index[test_ind[0]].isoformat(),
+                            f"{prefix}_test_end": X.index[test_ind[-1]].isoformat(),
+                            f"{prefix}_n_train": len(train_ind),
+                            f"{prefix}_n_test": len(test_ind),
```
milesgranger: As this is pretty well meant for json serving, we've been preferring `-` to `_`

gordo_components/builder/mlflow_utils.py
```diff
@@ -257,7 +257,12 @@ def get_batch_kwargs(metadata: dict) -> dict:
 
     metric_list: List[Metric] = list()
 
-    # Parse cross-validation metrics
+    # Parse cross-validation scoring metrics
+    if metadata["model"].get("cross-validation", {}).get("splits", None):
+        splits = metadata["model"]["cross-validation"]["splits"]
+        param_list += get_params(splits, splits.keys())
```
milesgranger: Is this a thing we're going to have to make a habit of? As in, each time a new parameter is added to the metadata we'll have to update this too? :fearful: 

Also, after working on removing watchman and using the server instead, I've found working in this module, as it relates to the `metadata` quite difficult/tedious to debug. IMO, we're making too many assumptions about the structure of the metadata; not sure this is the right place to handle it, but making mention of it anyway. :sweat_smile:

ryanjdillon: Yeah, I agree this is very gross and terrible. It was a concession made when the metadata couldn't be parsed in a more general way for mlflow, and one that I think needs to change.

gordo_components/builder/build_model.py
ryanjdillon: I had done that originally and then saw the metric keys were named in this way.  That said I think we could re-think the structure/naming in general inside the metadata.

gordo_components/builder/build_model.py
```diff
@@ -314,13 +320,39 @@ def set_seed(self, seed: int):
...
+    @staticmethod
+    def build_split_dict(X: pd.DataFrame, tssplit: TimeSeriesSplit) -> dict:
+        """
+        Get dictionary of cross-validation training dataset split metadata
+
+        Parameters
+        ----------
+        X: pd.DataFrame
+            The training dataset that will be split during cross-validation.
+        tssplit: sklearn.model_selection.TimeSeriesSplit
```
milesgranger: I _think_ one could generalize this to any splitter

gordo_components/builder/build_model.py
```diff
@@ -314,13 +320,39 @@ def set_seed(self, seed: int):
...
+            Dictionary of cross-validation train/test split metadata
+        """
+        split_metadata: Dict[str, Any] = dict()
+        for i, (train_ind, test_ind) in enumerate(tssplit.split(X)):
+            for name, ind in zip(["train", "test"], [train_ind, test_ind]):
+                for pos, idx in zip(["start", "end"], [0, -1]):
+                    split_metadata.update(
+                        {f"fold_{i+1}_{name}_{pos}": X.index[ind[idx]]}
+                    )
+                split_metadata.update({f"fold_{i+1}_n_{name}": len(ind)})
```
milesgranger: So the method below this uses hyphens, maybe this should to since that's the better practice when things are destined for json?

tests/test_examples.py
```diff
@@ -19,23 +19,27 @@
 
 logger = logging.getLogger(__name__)
 
+CONFIG = dict(
+    from_ts=dateutil.parser.isoparse("2014-07-01T00:10:00+00:00"),
+    to_ts=dateutil.parser.isoparse("2015-01-01T00:00:00+00:00"),
+    tag_list=["asgb.19ZT3950%2FY%2FPRIM", "asgb.19PST3925%2FDispMeasOut%2FPRIM"],
```
milesgranger: need fake tags in this great open-source project of ours.

ryanjdillon: This was a refactor, but you mean make this "faktag1", etc., right?

milesgranger: Ok, I just assumed `asgb.19ZT...` refers to an actual tag. Make it whatever you want.

gordo_components/builder/build_model.py
ryanjdillon: It does for those that return indices. For example, this wouldn't work with [train_test_split](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.train_test_split.html)

milesgranger: No, any cross validation splitter.. `TimeSeriesSplit`, `KFold..` blah blah

milesgranger: https://scikit-learn.org/stable/modules/classes.html#splitter-classes

ryanjdillon: Right, so there are "Splitter Classes" and "splitter functions". So should I just create a custom type hint class for sklearn splitter classes?

milesgranger: `sklearn.model_selection.BaseCrossValidator`

gordo_components/builder/build_model.py
```diff
@@ -314,13 +320,39 @@ def set_seed(self, seed: int):
...
+    @staticmethod
+    def build_split_dict(X: pd.DataFrame, tssplit: Type[BaseCrossValidator]) -> dict:
+        """
+        Get dictionary of cross-validation training dataset split metadata
+
+        Parameters
+        ----------
+        X: pd.DataFrame
+            The training dataset that will be split during cross-validation.
+        tssplit: Type[sklearn.model_selection.BaseCrossValidator]
```
milesgranger: Sorry, I don't mean to be pedantic here, but now `tssplit` doesn't make sense, maybe just `splitter`?

gordo_components/builder/build_model.py
```diff
@@ -368,11 +400,11 @@ def _score_per_tag(y_true, y_pred):
         metrics_dict = {}
         for metric in metrics_list:
             for index, col in enumerate(y.columns):
-                metric_str = metric.__name__.replace("_", "-")
+                metric_str = metric.__name__.replace("_", "_")
                 metrics_dict.update(
                     {
                         metric_str
-                        + f'-{col.replace(" ", "-")}': metrics.make_scorer(
+                        + f'_{col.replace(" ", "_")}': metrics.make_scorer(
```
milesgranger: What? Why do this? My comment read that _your_ new method should adhere to how _this_ method use hyphens and not underscores.

gordo_components/builder/build_model.py
ryanjdillon: No, that makes sense :+1:

gordo_components/builder/build_model.py
ryanjdillon: I think I was confused by what you meant by *this*. So back to how I had it the first time ;)

tests/test_examples.py
```diff
@@ -16,26 +16,34 @@
 
 from gordo_components.data_provider.providers import DataLakeProvider
 from gordo_components.dataset.datasets import TimeSeriesDataset
+from gordo_components.dataset.sensor_tag import SensorTag
 
 logger = logging.getLogger(__name__)
 
+CONFIG = dict(
+    from_ts=dateutil.parser.isoparse("2014-07-01T00:10:00+00:00"),
+    to_ts=dateutil.parser.isoparse("2015-01-01T00:00:00+00:00"),
```
milesgranger: Should rather be `train_start/end_date` the only reason they still work with `TimeSeriesDataset` is because we have a compatibility decorator on it's `init`

tests/test_examples.py
```diff
@@ -16,26 +16,34 @@
...
+    tag_list=[SensorTag(name=f"Tag {i}", asset=None) for i in range(10)],
+)
+
 
 def _fake_data():
-    data = pd.DataFrame({f"sensor_{i}": np.random.random(size=100) for i in range(10)})
+    index = pd.date_range(CONFIG["from_ts"], CONFIG["to_ts"], freq="12T")
```
milesgranger: Is this because it fails when they the index is _not_ a datetime? Do you think it's worth it to make it such that it'll still work without datetime indexes?

ryanjdillon: Yes, correct-a-mundo. It appeared to me as though they were intended to have datetime indices. Is that not the case?

ryanjdillon: Now it runs with a regular 'ole index.

gordo_components/builder/build_model.py
```diff
@@ -326,13 +329,39 @@ def set_seed(self, seed: int):
...
+        split_metadata: Dict[str,Any]
+            Dictionary of cross-validation train/test split metadata
+        """
+        split_metadata: Dict[str, Any] = dict()
+        for i, (train_ind, test_ind) in enumerate(split_obj.split(X)):
+            for name, ind in zip(["train", "test"], [train_ind, test_ind]):
+                for pos, idx in zip(["start", "end"], [0, -1]):
+                    split_metadata.update(
+                        {f"fold-{i+1}-{name}-{pos}": X.index[ind[idx]]}
+                    )
```
milesgranger: I like the params, but by the third for loop, my brain gets slightly tired.
Is something like this easier on the eyes? 
```python
split_metadata.update(
    {
        f"fold-{i+1}-train-start": X.index[train_ind[0]],
        f"fold-{i+1}-train-end": X.index[train_ind[-1]],
        f"fold-{i+1}-test-start": X.index[test_ind[0]],
        f"fold-{i+1}-test-end": X.index[test_ind[-1]]
    }
)
```

tests/gordo_components/builder/test_mlflow_utils.py
```diff
@@ -14,143 +15,213 @@
...
                                 {
-                                    "gordo_components.model.models.KerasAutoEncoder": {
-                                        "kind": "feedforward_hourglass",
-                                        "compression_factor": 0.5,
-                                        "encoding_layers": 2,
-                                        "func": "tanh",
-                                        "out_func": "linear",
-                                        "epochs": 3,
+                                    "sklearn.multioutput.MultiOutputRegressor": {
+                                        "estimator": "sklearn.linear_model.base.LinearRegression"
```
milesgranger: I guess all this is a copy-paste update?

tests/gordo_components/builder/test_mlflow_utils.py
```diff
@@ -14,143 +15,213 @@
...
+                    "fold-3-test-end": Timestamp(
+                        "2019-03-01 00:00:00+0000", tz="UTC", freq="10T"
+                    ),
+                    "fold-3-n-test": 2115,
                 },
             },
+            "feature-thresholds": [0.5609058195202434, 0.5959327804053265],
+            "aggregate-threshold": 1.2813764724006529,
+            "scaler": "RobustScaler(copy=True, quantile_range=(25.0, 75.0), with_centering=True,\n             with_scaling=True)",
+            "base_estimator": "Pipeline(memory=None,\n         steps=[('step_0',\n                 PCA(copy=True, iterated_power='auto', n_components=None,\n                     random_state=None, svd_solver='auto', tol=0.0,\n                     whiten=False)),\n                ('step_1',\n                 MultiOutputRegressor(estimator=LinearRegression(copy_X=True,\n                                                                 fit_intercept=True,\n                                                                 n_jobs=None,\n                                                                 normalize=False),\n                                      n_jobs=None))],\n         verbose=False)",
```
milesgranger: Is all this _really_ necessary? Not to say it should be fixed in this PR, but the fact that so much updating is required for a 'small' change like simply inserting start/end dates per fold requires all this adjustment for ML flow stuff to work seems excessive... :fearful:

tests/gordo_components/builder/test_mlflow_utils.py
ryanjdillon: Correct.

gordo_components/builder/mlflow_utils.py
EinarJohnsen: When we upgrade to Python 3.8, at least we can do: 
```
if splits := metadata["model"].get("cross-validation", {}).get("splits", None):
```
:heart_eyes:

gordo_components/builder/build_model.py
ryanjdillon: That's what I done did this first time, so I dooo agree. Should I change it back? I think both are reasonably readable on that note, though. 

I just generalized things while I abstracted it out to a function.

milesgranger: I find the nested loops has my brain doing a lot of extra work remapping names to variables. Primarily because what is being iterated over isn't a 'big' iterator, just two values. :man_shrugging:

tests/gordo_components/builder/test_mlflow_utils.py
milesgranger: Is it needed? Sorry, I'm just a bit skeptical of this mlflow utils test stuff.. Are we just copying and pasting because the tests would fail otherwise or?

tests/gordo_components/builder/test_mlflow_utils.py
ryanjdillon: I made an issue for it here #791

gordo_components/builder/build_model.py
ryanjdillon: Ok, changed it back now :+1:

tests/gordo_components/builder/test_mlflow_utils.py
milesgranger: Cool, I think that'll work nicely when we sort that out.. maybe I'll be less afraid of doing stuff in this module. :wink:

tests/test_examples.py
```diff
@@ -16,26 +16,32 @@
...
+
 
 def _fake_data():
-    data = pd.DataFrame({f"sensor_{i}": np.random.random(size=100) for i in range(10)})
+    data = pd.DataFrame(
+        {
+            f"sensor_{i}": np.random.random(size=100)
+            for i in range(len(CONFIG["tag_list"]))
+        }
+    )
```
milesgranger, line 36: The commit message no longer reflects what happened here.

gordo_components/builder/mlflow_utils.py
```diff
@@ -257,7 +257,12 @@ def get_batch_kwargs(metadata: dict) -> dict:
 
     metric_list: List[Metric] = list()
 
-    # Parse cross-validation metrics
+    # Parse cross-validation split metadata
+    if metadata["model"].get("cross-validation", {}).get("splits", None):
+        splits = metadata["model"]["cross-validation"]["splits"]
```
milesgranger: I _think_ this will always result in `None` because the `cross-validation` is stored under `build-metadata -> model`, no? 

---
Edit, I saw in the test's metadata that it wouldn't be the case there, but I think in the 'real-world' it would result in None, just to clarify that.

gordo_components/builder/build_model.py
```diff
@@ -248,12 +249,12 @@ def _build(self) -> Tuple[Optional[sklearn.base.BaseEstimator], dict]:
                 scaler = self.evaluation_config.get("scoring_scaler")
                 metrics_dict = self.build_metrics_dict(metrics_list, y, scaler=scaler)
 
+                # Generate metadata about CV train, test splits
+                split_obj = TimeSeriesSplit(n_splits=3)
```
EinarJohnsen, line 210: Do we have a way of specifying the number of splits in the evaluation config? Or should we make that into an issue if not? I do not like the hardcoded way of setting the number of splits.

tests/gordo_components/builder/test_mlflow_utils.py
```diff
@@ -14,143 +15,213 @@
...
             ],
-            "train_start_date": datetime(2017, 12, 25, 6, 0, tzinfo=UTC),
-            "train_end_date": datetime(2017, 12, 30, 6, 0, tzinfo=UTC),
+            "train_start_date": datetime.datetime(2019, 1, 1, 0, 0, tzinfo=tzutc()),
+            "train_end_date": datetime.datetime(2019, 3, 1, 0, 0, tzinfo=tzutc()),
             "resolution": "10T",
             "filter": "",
             "row_filter_buffer_size": 0,
             "data_provider": {"type": "RandomDataProvider"},
-            "asset": None,
```
milesgranger: Ok, maybe my misunderstanding is here, because this isn't actually the structure of the metadata returned from the builder, because there is no `build-metadata` key here? Should we wait for #795 to get in maybe?

ryanjdillon: Yeah, let's do that. Feels a bit unnecessary to fix this.

gordo_components/builder/build_model.py
ryanjdillon: I agree that should be changed, but I didn't think that should be for this PR.

gordo_components/builder/mlflow_utils.py
ryanjdillon: As per comment below, I'll delay this one for #795 as it's fixed there.

gordo_components/builder/build_model.py
EinarJohnsen: Yeah, I also see it as a separate issue, just shining some light at it.

gordo_components/builder/mlflow_utils.py
EinarJohnsen: Probably also a separate issue that can be discussed. But after some reading about how to best retrieving/checking for sub-keys from a nested dictionary, people say that using try/catch might be better than using the .get().get() thingy. It's easy to shoot oneself in the foot, if one of the get's return something else than a dict (or something that cannot be called get() on). 
Perhaps we should think about this at least. If the structure does not change, and we are guaranteed to always nest it in this way, then we will probably not have an issue, but it might be more error prone if things change.

ryanjdillon: Nice :+1: All that said, I think generalizing this method (and removing this cruft) once some dust has settled and we are sure we are keeping all of this would be prudent and wise.

tests/gordo_components/builder/test_mlflow_utils.py
ryanjdillon: Ok, this one is resolved now :+1:

gordo_components/builder/mlflow_utils.py
ryanjdillon: I decided to go with Einar's suggested here on this go around, since I needed to merge these changes anyway :+1:

gordo_components/builder/mlflow_utils.py
```diff
@@ -262,11 +262,20 @@ def get_batch_kwargs(metadata: dict) -> dict:
     model_keys = ["model-creation-date", "model-builder-version", "model-offset"]
     param_list += get_params(build_metadata["model"], model_keys)
 
+    # Parse cross-validation split metadata
+    try:
+        splits = build_metadata["model"]["cross-validation"]["splits"]
+        param_list += get_params(splits, splits.keys())
+    except KeyError:
+        logger.debug(
+            "Key 'metadata.model.cross-validation.splits' not found found in metadata."
```
milesgranger: This error isn't correct is it, referencing `metadata.model.cross-validation` when it would have been under `build-metadata`

gordo_components/builder/mlflow_utils.py
```diff
@@ -262,11 +262,20 @@ def get_batch_kwargs(metadata: dict) -> dict:
     model_keys = ["model-creation-date", "model-builder-version", "model-offset"]
     param_list += get_params(build_metadata["model"], model_keys)
 
+    # Parse cross-validation split metadata
+    try:
+        splits = build_metadata["model"]["cross-validation"]["splits"]
+        param_list += get_params(splits, splits.keys())
```
milesgranger: minor detail, but if the `try` block is ensuring that the previous line doesn't error, then the second should be in the `else` section. If this line, for whatever reason in the future also raises a `KeyError` then the logged error is not actually relevant.

gordo_components/builder/mlflow_utils.py
```diff
@@ -280,16 +289,20 @@ def get_batch_kwargs(metadata: dict) -> dict:
             # Summary stats per metric
             for sk in subkeys:
                 metric_list.append(
-                    Metric(f"{k}_{sk}", scores[k][f"fold-{sk}"], epoch_now(), 0)
+                    Metric(f"{k}-{sk}", scores[k][f"fold-{sk}"], epoch_now(), 0)
```
milesgranger, line 298: Nice catch. :+1:

gordo_components/builder/mlflow_utils.py
```diff
@@ -280,16 +289,20 @@ def get_batch_kwargs(metadata: dict) -> dict:
...
                 )
             # Append value for each fold with increasing steps
             metric_list += [
                 Metric(k, scores[k][f"fold-{i+1}"], epoch_now(), i)
                 for i in range(n_folds)
             ]
+    except KeyError:
+        logger.debug(
+            "Key 'metadata.model.cross-validation.scores' not found found in metadata."
```
milesgranger: same as above comments, regarding this message and there are _lots_ in here which should not be in `try`

gordo_components/builder/mlflow_utils.py
```diff
@@ -305,6 +318,8 @@ def get_batch_kwargs(metadata: dict) -> dict:
                 Metric(m, float(x), timestamp=epoch_now(), step=i)
                 for i, x in enumerate(data)
             ]
+    except KeyError:
+        logger.debug("Key 'metadata.model.history.params' not found found in metadata.")
```
milesgranger: same same

tests/test_examples.py
```diff
@@ -16,26 +16,32 @@
...
 from gordo_components.machine.dataset.datasets import TimeSeriesDataset
+from gordo_components.machine.dataset.sensor_tag import SensorTag
 
 logger = logging.getLogger(__name__)
 
+CONFIG = dict(
+    train_start_date=dateutil.parser.isoparse("2014-07-01T00:10:00+00:00"),
+    train_end_date=dateutil.parser.isoparse("2015-01-01T00:00:00+00:00"),
+    tag_list=[SensorTag(name=f"Tag {i}", asset=None) for i in range(10)],
+)
```
milesgranger, line 27: Do you think this should be a fixture? :man_shrugging:

gordo_components/builder/mlflow_utils.py
ryanjdillon: Fixed here and for the others :+1:

tests/test_examples.py
ryanjdillon: I did it as a global since it is used in a helper function that is passed as the return value to a `mock.patch`.  I don't THINK a fixture could be used here.

ryanjdillon: Just tried trying this out in a MWE, If I were to make that helper function a fixture, then use that as a return value I get this:
```
Fixture "fn" called directly. Fixtures are not meant to be called directly,
but are created automatically when test functions request them as parameters.
See https://docs.pytest.org/en/latest/fixture.html for more information about fixtures, and
https://docs.pytest.org/en/latest/deprecations.html#calling-fixtures-directly about how to update your code.
```

milesgranger (approved):
LGTM - But running it through the test project might be a good measure as well before merging. :+1:

ryanjdillon:
Test project looks good. Merging :+1:

PR #787, opened 2019-12-13
Remove references to ambassador in workflow

epa095:


milesgranger:
It's a WiP so maybe your not to it yet, but will we also want to remove _all_ ambassador references? See there is some in `cli/workflow_generator.py` and `server.py`

epa095:
tested on test-project and seems to work!

milesgranger (approved):
OMG, free at last, not bad! :clap:

PR #794, opened 2019-12-13
Make `Machine` a first class citizen

milesgranger:
After #766 it seemed rather clear there was duplication between what we have as `Dataset` as a config element and what we have as existing datasets. When merging those two, we're left with just a `Machine` element. 

We naturally reflect all of these in the config file during the process, `dataset`, `model`, etc. it seems `Machine` should be the composition of all of these throughout the same process. One example of this is when we then started using the `Client` to deal in terms of `Machine`s rather than "endpoint" metadata as it did before. 

Now we can simply pass the `Machine` around, to include inside the workflow template generation, rather than separating and passing around _dataset_, _metadata_, _model_, and all that just to "re-combine" them again in the CLI build.

Furthermore, we can transform `ModelBuilder` -> `ModelBuilderMixin` for the `Machine` and it will be able to build a `Machine` directly rather than passing the machine, as it is now, to the `ModelBuilder`.

---

Apologies for the number of file changes, I think the majority is documentation changes and import renamings.


milesgranger:
Can confirm it passed the gordo-test-project. :+1:

gordo_components/builder/build_model.py
```diff
@@ -199,18 +154,16 @@ def _build(self) -> Tuple[Optional[sklearn.base.BaseEstimator], dict]:
         -------
             Tuple[Optional[sklearn.base.BaseEstimator], dict]
         """
```
EinarJohnsen: I know this is not part of the changes in this PR, but as a side note: Is the return signature here correct, and the model really optional? For me it seems like we are under the assumption that the model are going to exist, and that it will always return a "valid" model.

milesgranger: I fix, I fix! :+1:

gordo_components/util/utils.py
```diff
@@ -0,0 +1,52 @@
+import functools
+import inspect
+from typing import Callable
+
+
+def capture_args(method: Callable):
+    """
+    Decorator to capture ``args`` and ``kwargs`` passed to a given method
+    Assumed there is a ``self`` to which it assigned the attribute to as ``_params``
+    as one dict of kwargs
```
ryanjdillon: Maybe reword to something like:
>   Decorator that captures ``args`` and ``kwargs`` passed to a given method.
>   This assumes the decorated method has a ``self`` , which has a dict of kwargs  assigned as an attribute named ``_params`` .

gordo_components/util/utils.py
```diff
@@ -0,0 +1,52 @@
...
+
+    Parameters
+    ----------
+    method: Callable
+        Some method of an object, with 'self' as the first parameter
```
ryanjdillon: Periods here and below

gordo_components/util/utils.py
```diff
@@ -0,0 +1,52 @@
...
+    -------
+    Any
+        Returns whatever the original method would return
+    """
+
+    @functools.wraps(method)
+    def wrapper(self, *args, **kwargs):
+        """
+        Need to map args to the parameters in the method and then update that
+        param dict with explicitly provided kwargs and assign that to _params
```
ryanjdillon: This sounds like a pseudo-code definition leftover?

gordo_components/machine/dataset/base.py
```diff
@@ -2,22 +2,66 @@
...
 import pandas as pd
 import numpy as np
 
+
 logger = logging.getLogger(__name__)
 
 
 class GordoBaseDataset:
+
+    _params: Dict[Any, Any] = dict()  # provided by @capture_args on child's __init__
```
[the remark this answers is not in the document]
ryanjdillon: About this, could you maybe do something like this?
```python
class Base:
    @abstractmethod
    def __init__(self, *args, **kwargs):
        self._params = self.capture_args(*args, **kwargs)
```

gordo_components/machine/dataset/base.py
```diff
@@ -2,22 +2,66 @@
...
+            raise AttributeError(
+                f"Failed to lookup init parameters, ensure the "
+                f"object's __init__ is decorated with 'capture_args'"
+            )
+        # Update dict with the class
+        params = self._params
+        params["type"] = self.__class__.__name__
+        for key, value in params.items():
+            if hasattr(value, "to_dict"):
+                params[key] = value.to_dict()
```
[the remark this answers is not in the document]
milesgranger: Oh, good catch!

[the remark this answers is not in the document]
milesgranger: oh, actually, all the params are already there, this just updates any with a `to_dict()` method.

gordo_components/cli/cli.py
```diff
@@ -172,68 +104,42 @@ def build(
...
         model, metadata = builder.build(output_dir, model_register_dir)
 
         if print_cv_scores:
             for score in get_all_score_strings(metadata):
                 print(score)
-
         # If enabled, configure remote logging to AzureML, otherwise logs locally
-        if enable_remote_logging:
-            metadata.update({"project-name": project_name})
+        if machine.runtime and machine.runtime["builder"]["remote_logging"]["enable"]:
```
EinarJohnsen: Try/Catch as we discussed

gordo_components/cli/workflow_generator.py
```diff
@@ -230,13 +230,14 @@ def unique_tag_list_cli(machine_config: str, output_file_tag_list: str):
     machines = NormalizedConfig(yaml_content, project_name="test-proj-name").machines
 
     tag_list = set(tag for machine in machines for tag in machine.dataset.tag_list)
+
     if output_file_tag_list:
         with open(output_file_tag_list, "w") as output_file:
-            for item in tag_list:
-                output_file.write("%s\n" % item)
+            for tag in tag_list:
+                output_file.write("%s\n" % tag.name)
```
EinarJohnsen: Should we change this to a f-string, as it seems like we overall lean towards using them?
 ```
output_file.write(f"{tag.name}\n")
```

File: gordo_components/machine/dataset/data_provider/azure_utils.py [hunk 9]
```diff
@@ -29,12 +30,16 @@ def get_datalake_token(
     lib.DataLakeCredential
         A lib.DataLakeCredential which can be used to authenticate towards the datalake
     """
-
+    dl_service_auth_str = (
+        os.environ.get("DL_SERVICE_AUTH_STR")
+        if str(dl_service_auth_str) == "None"
+        else dl_service_auth_str
```
EinarJohnsen, line 36: Is the string casting necessary? Coulden't we just check if it is None?

milesgranger: Ya, at some point I got literally `"None"` and thought this would cover both aspects ok enough.. probably should make an issue later to sort it out?

EinarJohnsen: Sounds a bit funny getting "None" :laughing:, so perhaps we should look into that, however this should cover both cases as you say :+1:

tests/conftest.py
```diff
@@ -193,7 +196,8 @@ def ml_server(
         targets=targets,
         model_location=trained_model_directory,
     ):
-        yield
+        with patch.object(sensor_tag, "_asset_from_tag_name", return_value="default"):
```
ryanjdillon, line 201: Same as above. I would maybe expect the default behaviour to be setting the asset to "default" if not defined, which makes me thing the test project tag_list sets assets, and we for some reason want them to be assigned a default asset instead for this test.

tests/gordo_components/builder/test_builder.py
```diff
@@ -379,9 +381,10 @@ def test_output_scores_metadata():
             """
 
     model_config = yaml.load(raw_model_config, Loader=yaml.FullLoader)
-    model, metadata = ModelBuilder(
-        name="model-name", model_config=model_config, data_config=data_config
-    ).build()
+    machine = Machine(
+        name="model-name", dataset=data_config, model=model_config, project_name="test"
+    )
+    model, metadata = ModelBuilder(machine).build()
```
ryanjdillon, line 387: Another "another PR" comment, but maybe `ModelBuilder` > `MachineBuilder`.

gordo_components/util/utils.py
```diff
@@ -0,0 +1,49 @@
...
+            if value.default is not inspect.Parameter.empty and param != "self"
+        }
+
+        # Update with arg values provided
+        arg_map = dict()
+        for arg_val, arg_key in zip(
+            args, (arg for arg in inspect.getfullargspec(method).args if arg != "self")
+        ):
+            arg_map[arg_key] = arg_val
+
```
EinarJohnsen, line 41: Not necessarily more readable, but this could be a dict comprehension. 
```
arg_map = {arg_key: arg_val for arg_key, arg_val in zip(
            args, (arg for arg in inspect.getfullargspec(method).args if arg != "self"))
          }  
```

milesgranger: ha, ya, I had changed it from the original for to the comprehension, but changed it back. I can definitely change it though. I have no strong opinions.

tests/gordo_components/builder/test_builder.py
```diff
@@ -430,10 +433,10 @@ def test_provide_saved_model_caching_handle_existing_same_dir(tmp_dir):
     data_config = get_random_data()
     output_dir = os.path.join(tmp_dir, "model")
     registry_dir = os.path.join(tmp_dir, "registry")
-
-    builder = ModelBuilder(
-        name="model-name", model_config=model_config, data_config=data_config
+    machine = Machine(
+        name="model-name", dataset=data_config, model=model_config, project_name="test"
```
ryanjdillon, line 437: Small thing, but sometimes it is:
```python
name="model-name", project_name="test"
```
and sometimes:
```python
name="test", project_name="test"
```
maybe make these uniform.

gordo_components/machine/dataset/datasets.py
```diff
@@ -122,11 +138,17 @@ def __init__(
...
+            raise ValueError(
+                f"train_end_date ({self.train_end_date}) must be after train_start_date ({self.train_start_date})"
+            )
+
         self.tag_list = normalize_sensor_tags(list(tag_list), asset, default_asset)
         self.target_tag_list = (
             normalize_sensor_tags(list(target_tag_list), asset, default_asset)
             if target_tag_list
-            else []
+            else self.tag_list.copy()
```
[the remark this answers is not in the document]
milesgranger: :man_shrugging: yaaa. This is mostly because by combining the config `Dataset` and the associated validators, it would error if this was empty. We could copy a bit the logic of the model side and set `_strict=False` but I dunno, certainly seems like an edge case.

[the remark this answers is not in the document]
EinarJohnsen: I see, I have no particular opinion about it, was more a general wondering :man_shrugging:

gordo_components/util/utils.py
ryanjdillon: I vote the comprehension, but also very not strongly opinionated about this.

gordo_components/machine/dataset/datasets.py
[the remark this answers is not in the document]
milesgranger: Maybe, maybe not.. it's a python list which is scary to share. :man_shrugging:

gordo_components/machine/dataset/datasets.py
```diff
@@ -217,10 +246,11 @@ def get_metadata(self):
 class RandomDataset(TimeSeriesDataset):
     """
     Get a TimeSeriesDataset backed by
-    gordo_components.data_provider.providers.RandomDataProvider
+    gordo_components.machine.dataset.data_provider.providers.RandomDataProvider
```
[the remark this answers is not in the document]
milesgranger: you mean `gordo_components.machine.dataset.data_provider import RandomDataProvider`
I can get on board with that. :+1:

gordo_components/util/utils.py
EinarJohnsen: Well, I don't have a strong opinion either. There is no additional lookups, so go for the gut feeling and just leave it then :leaves:

tests/gordo_components/builder/test_builder.py
milesgranger: Oh ya, definitely agree there! :+1:

gordo_components/machine/model/anomaly/diff.py
```diff
@@ -10,10 +10,10 @@
 from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.model_selection import TimeSeriesSplit, cross_validate
 
-from gordo_components.model.base import GordoBase
-from gordo_components.model import utils as model_utils
-from gordo_components.model.models import KerasAutoEncoder
-from gordo_components.model.anomaly.base import AnomalyDetectorBase
+from gordo_components.machine.model.base import GordoBase
+from gordo_components.machine.model import utils as model_utils
```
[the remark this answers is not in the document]
milesgranger: ya, if we change this, it's another PR as Erik and I discussed this earlier. As `utils` exists _inside_ the `model` module it seemed redundant naming.. but ya.. we could discuss this after as this is outside the scope of this PR.

tests/gordo_components/builder/test_builder.py
```diff
@@ -512,20 +515,25 @@ def test_provide_saved_model_caching(
...
     output_dir = os.path.join(tmp_dir, "model")
     registry_dir = os.path.join(tmp_dir, "registry")
-
-    _model, first_metadata = ModelBuilder(
-        name="model-name", model_config=model_config, data_config=data_config
-    ).build(output_dir=output_dir, model_register_dir=registry_dir)
+    machine = Machine(
+        name="model-name", dataset=data_config, model=model_config, project_name="test"
+    )
+    _model, first_metadata = ModelBuilder(machine).build(
```
EinarJohnsen: Not your doing, but this is not a private variable, but rather an unused variable. Might change it to "_" as it is not used.

tests/gordo_components/builder/test_builder.py
```diff
@@ -512,20 +515,25 @@ def test_provide_saved_model_caching(
...
+        output_dir=output_dir, model_register_dir=registry_dir
+    )
 
     if tag_list:
         data_config["tag_list"] = tag_list
 
     new_output_dir = os.path.join(tmp_dir, "model2")
     _model, second_metadata = ModelBuilder(
```
EinarJohnsen: ^

tests/gordo_components/builder/test_builder.py
EinarJohnsen: I see we do this at multiple places. Change it if you want to, otherwise just ignore me.

gordo_components/machine/dataset/data_provider/providers.py
```diff
@@ -231,6 +232,8 @@ def __init__(
                     dataframe_client=True,
                 )
             else:
+                if "type" in kwargs:
+                    kwargs.pop("type")
```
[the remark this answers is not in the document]
milesgranger: bleh, i take it back.. other things do care about the `type` between from_dict and here when done through higher levels like the `Machine` or `Dataset` and i can't modify the `init` of the DataFrameClient as that's in another lib.

tests/conftest.py
milesgranger: I think we wanted to explicitly raise an error during an actual workflow, which is why we don't have any default asset for a tag name. But agree with making comments here.

flikka (commented):
A quickish lookthrough: I find a lot of this sensible, simplifcations and gatherings so good. Great stuff!

One concept I struggle with. It seems like the "machine" module has been now a swallower of everything? I don't quite see why just "A (in this case machine) always contains a B (for example a model or a dataset)" implies that B must be a submodule of A? For example dataset, or model. Could they not live outside of a machine? A lot of deep paths that I am not quite convinced I find necessary... Could these things not continue to live next to eachother? Makes navigation easier I think. 

A scikit-learn pipeline "Pipeline" contains a lot of stuff. But that doesn't mean that all that stuff must live in the sklearn.pipeline module, or?

tests/gordo_components/builder/test_builder.py
milesgranger: meh, I think a leading `_` is only "private" if it's an attribute, otherwise it would be pointless to have a private free variable, because no one could actually use it. Anyhow, I can definitely change it, gave it a name only to easily know _what_ it is.

milesgranger (commented):
Valid points. IMO, I found it rather easier to traverse what's in components by have less "components" in the top level. 

I understand `dataset` and the like are useful _outside_ of Gordo, but within Gordo, we use these things in the context of a `Machine` which implies they belong as sub-modules to it. 

In this way, it's not similar to sklearn pipelines, because the transformers themselves _are_ used outside of a pipeline in the sklearn world. When running Gordo to deploy models, things like `dataset` only make sense in the context of a `Machine`.. so that's at least my rationalization of the restructuring of modules.

flikka:
> I understand `dataset` and the like are useful _outside_ of Gordo, but within Gordo, we use these things in the context of a `Machine` which implies they belong as sub-modules to it.

I acknowledge your arguments, and will not press this! Maybe I will even agree once it grows on me a bit. Renaming gordo_components to gordo would also make it more smooth to type. 

Overall a very nice changeset! I suggest we do a bit of a discussion/voting tomorrow, and merge it (or test it more first) - and rather keep an eye out for missing things the next days, such a bit change should probably not brew for to long 👍

tests/gordo_components/builder/test_builder.py
EinarJohnsen: I agree, a free private variable makes little sense here anyway, if it was a module it would perhaps be different. It's all convention anyhow, so it really does not matter unless you do import * on a module with "private" functions or something like that. 
So just leave it if you want, does not really matter much.

milesgranger: I've fallen in line with your way of thinking, tis now `_` :+1:

ryanjdillon:
About moving the dataset/model under machine, I think it makes the context of the modules more intuitive. That said, I think it makes sense to move the dataset/provider to it's own repo, and then whatever gordo specific routines could live where Miles has them now.

gordo_components/cli/cli.py
```diff
@@ -172,70 +104,50 @@ def build(
...
+    if model_parameter and isinstance(machine["model"], str):
+        model_parameter = dict(model_parameter)
+        machine["model"] = expand_model(machine["model"], model_parameter)
 
-    # Set default data provider for data config
-    # TODO: This is for backwards compatibility, as the `data_provider` param should
-    # TODO: be provided in the `data_config` itself
-    if data_provider is not None:
-        data_config["data_provider"] = data_provider
+    machine = Machine.from_config(machine, project_name=machine["project_name"])
```
flikka: I find it a bit confusing to read code (in general also, yes LOL) that reuses variables, in particular when they change type and "meaning". Something like machine_config or machine_dict perhaps? It's just when reading the code, jumping a bit up and down, when it has changed to something else in-the-middle it confuses me.

Also, type hints would be cool-o-yeah. Since you are changing it anyways, and already have it in the docstring. Or is there some reason for omitting it I'm missing here perhaps?

milesgranger: Ya, I agree. I fix, I fix!

flikka (approved):
What could possibly go wrong?

milesgranger:
One last time for QA. 
image

PR #795, opened 2019-12-16
Add metadata fixture

### LINKED ISSUE #791: Reduce the size of example metadata for testing mlflow_utils
And while at it... there are probably some simpler ways of avoiding having example metadata for the tests altogether.

ryanjdillon:
# Problema :hungary: 
The metadata fixture in `test_mlflow_utils.py` is static, takes up space, and requires updating everying the metadata structure is updated. 

# Solución :taco: 
Make a metadata fixture that uses the `trained_model_directory` fixture, loading the metadata from a model trained with the latest and greatest. 

In doing so the following were also needed and done...
* the `collection_dir` was abstracted out to another fixture
* the `tmp_dir` fixture's scope was changed to session so these build outputs persist across tests
* fixes some errors in tests and mlflow_utils metadata parsing.

closes #791


tests/conftest.py
```diff
@@ -42,10 +42,10 @@ def sensors():
     return tu.SENSORTAG_LIST
 
 
-@pytest.fixture
+@pytest.fixture(scope="session")
```
milesgranger: This seems unsafe, one temp directory for the entire test session?

tests/gordo_components/builder/test_mlflow_utils.py
```diff
@@ -4,155 +4,8 @@
...
-                    "epochs": 3,
-                    "steps": 23,
-                    "samples": 708,
-                    "verbose": 0,
-                    "do_validation": False,
-                    "metrics": ["loss", "accuracy"],
-                },
-            },
-        },
-    }
```
milesgranger: Sweet!

tests/gordo_components/builder/test_mlflow_utils.py
```diff
@@ -330,22 +183,22 @@ def test_get_batch_kwargs(metadata):
     Test that dicts are correctly converted to MLflow types or errors raised
     """
 
-    def _test_convert_metadata(metadata):
+    def _test_mlflow_batch_arg_types(metadata):
```
milesgranger, line 186: This renaming seems unrelated to adding the metadata fixture, another commit?

tests/conftest.py
ryanjdillon: For some tests this is useful. The pytest builtin `tmpdir` fixture is function scoped, so we could use that in instances where we want things to get flushed. From what I've seen, this doesn't seem to be problem with existing tests, but I can make doubly sure :+1:

milesgranger: ya, but that's nothing to preclude it from being a weird problem in the future. Is it possible to just add the tempdir on creation at the specific fixture which requires it for the session?

ryanjdillon: Definitely. This just seemed a tad bit cleaner, but I'm down with that if that seems best if it may be problematic.

milesgranger: Cool, and I think us implementing our own `tmp_dir` was merely an oversight from the pytest builtin :+1:

tests/conftest.py
```diff
@@ -47,7 +47,7 @@ def sensors():
 @pytest.fixture
 def tmp_dir():
     """
-    Fixture: Temporary directory, removed after test completion
+    Fixture: Temporary directory, removed after session completion
```
milesgranger: outdated now

tests/conftest.py
```diff
@@ -74,58 +74,68 @@ def base_route(api_version, gordo_project, gordo_name):
...
+
+    builder = local_build(config_str=config, enable_mlflow=False)
+    model, metadata = next(builder)  # type: ignore
+    serializer.dump(model, model_dir, metadata=metadata)
+    yield model_dir
+
+
+@pytest.fixture
+def metadata(trained_model_directory):
+    return serializer.load_metadata(trained_model_directory)
```
milesgranger, line 144: Nice!

tests/gordo_components/builder/test_mlflow_utils.py
```diff
@@ -4,155 +4,8 @@
...
-                    "epochs": 3,
-                    "steps": 23,
-                    "samples": 708,
-                    "verbose": 0,
-                    "do_validation": False,
-                    "metrics": ["loss", "accuracy"],
-                },
-            },
-        },
-    }
```
milesgranger, line 155: :wave: :+1:

milesgranger (commented):
Looking nice!

gordo_components/builder/mlflow_utils.py
```diff
@@ -242,29 +242,36 @@ def get_batch_kwargs(metadata: dict) -> dict:
         instances.
     """
 
+    metric_list: List[Metric] = list()
+    build_metadata = metadata["metadata"]["build-metadata"]
+
+    # Projecti/machine parameters
```
flikka: Projecti?

gordo_components/builder/mlflow_utils.py
```diff
@@ -242,29 +242,36 @@ def get_batch_kwargs(metadata: dict) -> dict:
...
-    metric_list: List[Metric] = list()
+    # Model parameters
+    model_keys = ["model-creation-date", "model-builder-version", "model-offset"]
+    param_list += get_params(build_metadata["model"], model_keys)
 
     # Parse cross-validation metrics
-    if metadata["model"].get("cross-validation", {}).get("scores", None):
-        scores = metadata["model"]["cross-validation"]["scores"]
+    tag_list = normalize_sensor_tags(
+        metadata["dataset"]["tag_list"], asset=metadata["dataset"]["tag_list"]
```
flikka: Looks strange? 
`asset=metadata["dataset"]["tag_list"]` - don't think that makes sense.

flikka (commented):
Looks good, I just have two comments, then I would say GO!

tests/conftest.py
```diff
@@ -74,58 +74,74 @@ def base_route(api_version, gordo_project, gordo_name):
     return f"/gordo/{api_version}/{gordo_project}/{gordo_name}"
 
 
+@pytest.fixture(scope="session")
+def model_collection_directory(gordo_project: str):
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        collection_dir = os.path.join(tmp_dir, gordo_project)
+        os.makedirs(collection_dir, exist_ok=True)
+        yield collection_dir
```
milesgranger, line 82: Would this maybe be a good opportunity to use pytest's built in tmpdir?

tests/conftest.py
```diff
@@ -74,58 +74,74 @@ def base_route(api_version, gordo_project, gordo_name):
...
+
+
+@pytest.fixture(scope="session")
+def config_str(gordo_name: str, sensors: List[SensorTag]):
+    """
+    Fixture: Default config for testing
+    """
+    yield f"""
```
milesgranger: Is there a functional difference between using `yield` vs `return` here?

tests/conftest.py
```diff
@@ -74,58 +74,74 @@ def base_route(api_version, gordo_project, gordo_name):
...
+                    - {sensors[2].name}
+                    - {sensors[3].name}
+                  target_tag_list:
+                    - {sensors[0].name}
+                    - {sensors[1].name}
+                    - {sensors[2].name}
+                    - {sensors[3].name}
+                  train_start_date: '2019-01-01T00:00:00+00:00'
+                  train_end_date: '2019-10-01T00:00:00+00:00'
+                  asset: asgb
```
milesgranger, line 105: No strong feelings here, but should we do one of three things here:
1. Change this to `{sensors[0].asset}` <- assuming all other tags are the same asset, which I guess is pretty safe.
2. Update things like `tags` elements to: `{dict(sensors[0]._asdict())}` which would then ensure each element had an asset defined and delete this `asset` param.
3. Leave it, because I guess it doesn't _really_ matter. :sweat_smile:

milesgranger (commented):
Sorry just a couple more comments, but this is going to be nice to have! :clap:

tests/conftest.py
ryanjdillon: I have an Issue I opened on going through and doing all tmpdir swaps in one fell swoop.

tests/conftest.py
ryanjdillon: Not to my knowledge.

tests/conftest.py
ryanjdillon: Currently, all of those sensors have no asset set I believe, it's getting set here in this config, so 1. and 2. would break in some cases when normalize_sensors is called -- without patching the asset somehow-- I think.

gordo_components/builder/mlflow_utils.py
ryanjdillon: :disappointed:

gordo_components/builder/mlflow_utils.py
ryanjdillon: It would if we wanted to set the `tag_list` to the `asset`, buuuut... I suppose we don't.

tests/conftest.py
milesgranger: ya, but that doesn't mean we have to repeat the error until _that_ issue is fixed, right?

ryanjdillon: Sorry, looking at this in the code again... we shouldn't/can't use the pytest built-in, as it's function scoped, and this is session scoped. That was why I session scoped the "custom" tmp_dir, figuring we could use that one when we wanted a session-scoped temp dir.

ryanjdillon: It seems like my last comment didn't get posted here. 

Looking at this again, the reason I didn't do that is that the pytest fixture is function scoped, which was why I session-scoped our "custom" tmp_dir fixture, so we could use that when we needed a session scoped tmp_dir, such as here.

milesgranger (approved):

PR #798, opened 2019-12-17
Fix artifact logging

ryanjdillon:
# Problems :hole: 
The configs/metadata are not currently logged as artifacts

# Solution :moneybag: 
Fix logging of artifacts to AzureML with mlflow client. With an integration test to make sure all's well that logs well.

### DIFF
--- a/gordo_components/builder/mlflow_utils.py
+++ b/gordo_components/builder/mlflow_utils.py
@@ -12,8 +12,10 @@ from azureml.core.authentication import (
     InteractiveLoginAuthentication,
     ServicePrincipalAuthentication,
 )
+import mlflow
 from mlflow.entities import Metric, Param
 from mlflow.tracking import MlflowClient
+import numpy as np
 from pytz import UTC
 
 from gordo_components.machine.dataset.sensor_tag import normalize_sensor_tags
@@ -103,6 +105,8 @@ def get_mlflow_client(
             workspace_kwargs["auth"] = InteractiveLoginAuthentication(force=True)
 
         tracking_uri = Workspace(**workspace_kwargs).get_mlflow_tracking_uri()
+
+    mlflow.set_tracking_uri(tracking_uri)
 
     return MlflowClient(tracking_uri)
 
@@ -398,46 +402,26 @@ def get_spauth_kwargs() -> dict:
     )
 
 
-class DatetimeEncoder(json.JSONEncoder):
+class MetadataEncoder(json.JSONEncoder):
     """
     Encode datetime.datetime objects as strings
 
     Example
     -------
-    >>> s = json.dumps({"now":datetime.now(tz=UTC)}, cls=DatetimeEncoder, indent=4)
+    >>> s = json.dumps({"now":datetime.now(tz=UTC)}, cls=MetadataEncoder, indent=4)
     >>> s = '{"now": "2019-11-22 08:34:41.636356+"}'
     """
 
     def default(self, obj):
         if isinstance(obj, datetime):
             return obj.strftime("%Y-%m-%d %H:%M:%S.%f+%z")
+        # Typecast builtin and numpy ints and floats to builtin types
+        elif np.issubdtype(type(obj), np.floating):
+            return float(obj)
+        elif np.issubdtype(type(obj), np.integer):
+            return int(obj)
         else:
             return json.JSONEncoder.default(self, obj)
-
-
-def send_artifacts(client: MlflowClient, run_id: str, metadata: dict):
-    """
-    Send artifacts
-
-    Parameters
-    ----------
-    client: MlflowClient
-        MLflow tracking client.
-    run_id: str
-        MLflow Run ID.
-    metadata: dict
-        Metadata produced by :func:`~gordo_components.builder.build_model.build_model`
-        to extract configs from.
-    """
-    with tempfile.TemporaryDirectory() as tmp_dir:
-        for name, obj in [
-            ("model-config", metadata["model"]["model-config"]),
-            ("dataset-config", metadata["dataset"]),
-            ("user-config", metadata["user-defined"]),
-        ]:
-            fp = os.path.join(tmp_dir, f"{name}.json")
-            json.dump(obj, open(fp, "w"), cls=DatetimeEncoder)
-        client.log_artifacts(run_id, local_dir=tmp_dir)
 
 
 @contextmanager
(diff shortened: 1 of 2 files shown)


gordo_components/builder/mlflow_utils.py
```diff
@@ -474,20 +474,9 @@ def log_metadata(mlflow_client, run_id, metadata: dict):
...
-    # Log error without exiting the build
-    # FIX: when a solution is found when logging artifacts wit azureml lib,
-    # this can raise an error
+        with tempfile.TemporaryDirectory(dir="./") as tmp_dir:
+            fp = os.path.join(tmp_dir, f"metadata.json")
+            json.dump(metadata, open(fp, "w"), cls=DatetimeEncoder)
+            mlflow_client.log_artifacts(run_id, tmp_dir)
     except Exception as e:
-        logger.error(e)
+        raise e
```
milesgranger: Is there a point now to the try block?

gordo_components/builder/mlflow_utils.py
```diff
@@ -474,20 +477,9 @@ def log_metadata(mlflow_client, run_id, metadata: dict):
...
-            ]:
-                fp = os.path.join(tmp_dir, f"{name}.json")
-                json.dump(obj, open(fp, "w"), cls=DatetimeEncoder)
-            mlflow_client.log_artifacts(run_id, local_dir=tmp_dir)
-    # Log error without exiting the build
-    # FIX: when a solution is found when logging artifacts wit azureml lib,
-    # this can raise an error
+        with tempfile.TemporaryDirectory(dir="./") as tmp_dir:
+            fp = os.path.join(tmp_dir, f"metadata.json")
+            json.dump(metadata, open(fp, "w"), cls=DatetimeEncoder)
```
milesgranger: Is it safe not to use a context manager when writing to this file, do you know if the internals of `json.dumps` will close the file if it cannot complete serialization process?

gordo_components/builder/mlflow_utils.py
```diff
@@ -474,20 +477,9 @@ def log_metadata(mlflow_client, run_id, metadata: dict):
...
+            mlflow_client.log_artifacts(run_id, tmp_dir)
     except Exception as e:
-        logger.error(e)
+        raise MlflowLoggingError(e)
```
milesgranger, line 494: Why are we doing this? Seems it will make things harder to debug, in that if say it errors serializing with a TypeError, we don't get a `TypeError` at all. Is it not better to simply remove the try catch?

milesgranger: I see a bit of a trade off there. :thinking: Maybe a simple comment to mention why remapping the error is necessary, Otherwise it just looks strange. IMO

gordo_components/builder/mlflow_utils.py
ryanjdillon: What do you mean by serialization process?

milesgranger: if `json.dumps` (the process of converting `metadata` to json, serialization) fails, there is nothing in there that will close the handle to this file; and in general [opening files like this is an anti-pattern](https://docs.quantifiedcode.com/python-anti-patterns/maintainability/not_using_with_to_open_files.html)

ryanjdillon: As discussed, rolling this one into #795.

gordo_components/builder/mlflow_utils.py
ryanjdillon: Roger. Commented :heavy_check_mark:

gordo_components/builder/mlflow_utils.py
```diff
@@ -398,48 +401,36 @@ def get_spauth_kwargs() -> dict:
...
+    @staticmethod
+    def reduce_bit_type(obj):
+        for type_cls, type_str in [(float, "float"), (int, "int")]:
+            if type_str in str(type(obj)):
+                return type_cls(obj)
+
     def default(self, obj):
         if isinstance(obj, datetime):
             return obj.strftime("%Y-%m-%d %H:%M:%S.%f+%z")
+        # Typecast floats and ints to 8bit python number objects
```
milesgranger: This is just wrong. You're saying you want to cast the number to 8bits, using python's `int` or `float`? 8 bits is 0-255 for unsigned and -128-127 for signed; definitely not what `int` or `float` is.

gordo_components/builder/mlflow_utils.py
```diff
@@ -398,48 +401,36 @@ def get_spauth_kwargs() -> dict:
...
+    @staticmethod
+    def is_int_or_float(obj):
+        return any([type_str in str(type(obj)) for type_str in ("float", "int")])
+
+    @staticmethod
+    def reduce_bit_type(obj):
+        for type_cls, type_str in [(float, "float"), (int, "int")]:
+            if type_str in str(type(obj)):
+                return type_cls(obj)
+
```
milesgranger: We certainly don't want this... 

What I _think_ you're after is basically checking if the number object is a numpy instance which is not json serializable. ie.
```python
a = numpy.int32(1) 
json.dumps({"my-number": a})  # This will fail.
```

So what I think you want, is more something like a numpy to python conversion:
```python
if isinstance(a, np.float):
    return float(a)
elif isinstance(a, np.integer):
    return int(a)
else:
    return a
```

gordo_components/builder/mlflow_utils.py
ryanjdillon: Ok... ok.

gordo_components/builder/mlflow_utils.py
ryanjdillon: That's correct-a-mundo. Do you think there may be other float/int types outside of numpy that might not be covered by `json`?

gordo_components/builder/mlflow_utils.py
```diff
@@ -398,48 +402,40 @@ def get_spauth_kwargs() -> dict:
...
     def default(self, obj):
+        np_float_types = (np.float, np.float16, np.float32, np.float64, np.float128)
+        np_int_types = (np.int, np.int0, np.int8, np.int16, np.int32, np.int64)
         if isinstance(obj, datetime):
             return obj.strftime("%Y-%m-%d %H:%M:%S.%f+%z")
+        # Convert numpy floats and ints to built-in python types
+        elif any([isinstance(obj, t) for t in np_float_types]):
+            return float(obj)
+        elif any([isinstance(obj, t) for t in np_int_types]):
+            return int(obj)
```
milesgranger: Possible to just use `np.integer` and `np.floating`? Don't they cover all subtypes?

gordo_components/builder/mlflow_utils.py
```diff
@@ -398,48 +402,40 @@ def get_spauth_kwargs() -> dict:
...
+    @staticmethod
+    def is_int_or_float(obj):
+        return any([type_str in str(type(obj)) for type_str in ("float", "int")])
+
+    @staticmethod
+    def reduce_bit_type(obj):
+        for type_cls, type_str in [(float, "float"), (int, "int")]:
+            if type_str in str(type(obj)):
+                return type_cls(obj)
+
```
milesgranger: Not needed anymore, right?

gordo_components/builder/mlflow_utils.py
ryanjdillon: Ah.. maybe that's why. I was trying `np.float` and `np.int` and that didn't do it.

gordo_components/builder/mlflow_utils.py
ryanjdillon: right :+1:

gordo_components/builder/mlflow_utils.py
ryanjdillon: Bummer, doesn't work.

ryanjdillon: Solution using `np.issubtype` now pushed.

gordo_components/builder/mlflow_utils.py
```diff
@@ -398,48 +402,28 @@ def get_spauth_kwargs() -> dict:
...
     """
 
     def default(self, obj):
         if isinstance(obj, datetime):
             return obj.strftime("%Y-%m-%d %H:%M:%S.%f+%z")
+        # Typecast builtin and numpy ints and floats to builtin types
+        elif np.issubdtype(type(obj), np.floating):
+            return float(obj)
+        elif np.issubdtype(type(obj), np.integer):
+            return int(obj)
```
milesgranger, line 422: Nice! :+1:

milesgranger (approved):

Issue #775, opened 2019-12-10
Server: Add endpoint giving revisions it can serve


epa095:
Add an `GET` endpoint which lists out the versions of the project it can find on the disk. It should be as simple as listing the directory one up of `MODEL_COLLECTION_DIR`. 
Also consider changing it a bit so instead of taking in `MODEL_COLLECTION_DIR` which is `/gordo/models/{{project_name}}/{{project_version}}` the server takes in `/gordo/models/{{project_name}}` and then a second parameter which is `default_version` ? Dealers choice. 

Feel free to put a TTL cache with e.g `2s` timeout on it.

epa095:
Maybe the response should be two things, a list of revisions *and* a "latest:blala" key giving which revision is the latest

PR #804, opened 2019-12-18
Use project revision instead of version

milesgranger:
prelude to #775

Gordo controller (and discussions with @epa095 ) we've agreed that the term 'revision' better suits what this is, similar to a git sha. 

Before moving onto supporting giving revisions the server can facilitate, best to square away the same use of terms.


flikka (approved):
Nice and simple. Vesion is such a nice word, a shame letting it go.

PR #805, opened 2019-12-18
Add revision list view to server

### LINKED ISSUE #775: Server: Add endpoint giving revisions it can serve

milesgranger:
Will close #775 

List project revisions the server can serve, including the latest

### DIFF
--- a/gordo_components/server/views/base.py
+++ b/gordo_components/server/views/base.py
@@ -252,6 +252,27 @@ class ModelListView(Resource):
         return jsonify(available_models)
 
 
+class RevisionListView(Resource):
+    """
+    List the available revisions the model can serve.
+    """
+
+    @api.doc(description="Available revisions of the project that can be served.")
+    def get(self, gordo_project: str):
+        collection_dir = os.environ[current_app.config["MODEL_COLLECTION_DIR_ENV_VAR"]]
+        try:
+            available_revisions = os.listdir(os.path.join(collection_dir, ".."))
+        except FileNotFoundError:
+            logger.error(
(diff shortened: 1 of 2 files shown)


gordo_components/server/views/base.py
```diff
@@ -252,6 +252,27 @@ def get(self, gordo_project: str):
...
+        try:
+            available_revisions = os.listdir(os.path.join(collection_dir, ".."))
+        except FileNotFoundError:
+            logger.error(
+                f"Attempted to list directories above {collection_dir} but failed with: {traceback.format_exc()}"
+            )
+            available_revisions = [collection_dir]
+
+        return jsonify(
+            {"latest": collection_dir, "available-revisions": available_revisions}
```
epa095, line 272: 👍

milesgranger: :hankey: :toilet: ?

epa095: :-1: :tv: :baby:

EinarJohnsen (approved):
LGTM :snowman_with_snow:

PR #810, opened 2019-12-19
Server: use revision to serve metadata/models

### LINKED ISSUE #745: Server: Handle project revision
Related task: #744
Accept a project version parameter / header whatever with requests (to predict / anomaly / metadata), and if that is provided then only return 200 OK with a result if it can be served from that exact version. I propose we try to serve old versions in a "best effort", so we can serve old models, but if we are not able to resolve a version we return some other code than 500. Maybe a `410 Gone`.

milesgranger:
Will close #745

### DIFF
--- a/tests/gordo_components/server/test_gordo_server.py
+++ b/tests/gordo_components/server/test_gordo_server.py
@@ -192,7 +192,8 @@ def test_models_by_revision_list_view(caplog, tmpdir, revision_to_models):
             )
             assert resp.status_code == 410
             assert resp.json == {
-                "error": "Revision 'revision-does-not-exist' not found."
+                "error": "Revision 'revision-does-not-exist' not found.",
+                "revision": "revision-does-not-exist",
             }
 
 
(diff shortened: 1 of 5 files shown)


ryanjdillon (commented):
Looking good :cowboy_hat_face:

milesgranger:
Ran this through the test project without any trouble just now, if that adds any confidence for reviewers. :+1:

tests/gordo_components/server/test_gordo_server.py
```diff
@@ -109,8 +111,9 @@ def test_list_revisions(tmpdir, revisions: List[str]):
         client = app.test_client()
         resp = client.get("/gordo/v0/test-project/revisions")
 
-    assert set(resp.json.keys()) == {"latest", "available-revisions"}
-    assert resp.json["latest"] == model_dir
+    assert set(resp.json.keys()) == {"latest", "available-revisions", "revision"}
+    assert resp.json["latest"] == os.path.basename(model_dir)
+    assert resp.json["revision"] == os.path.basename(model_dir)
```
ryanjdillon, line 119: Is there a scenario where the client could not be *latest*, but it is aware of *latest*? If so, maybe a test for that combo to be sure it's working.

ryanjdillon (approved):
LGTM :tada:

tests/gordo_components/server/test_gordo_server.py
milesgranger: Extended the test to do as you suggest. :+1:

PR #816, opened 2019-12-23
Client: use revision

### LINKED ISSUE #747: Client: Pass project revision to server and handle response
Part of #744 

Send the project_version together with the data to the server. Handle the possible 410 response code by updating its metadata from watchman.

milesgranger:
Depends on #810 

Close #747

### DIFF
--- a/gordo_components/client/client.py
+++ b/gordo_components/client/client.py
@@ -114,8 +114,6 @@ class Client:
         self.data_provider = data_provider
         self.use_parquet = use_parquet
         self.project_name = project
-        self.session = session or requests.Session()
-        self.revision = revision
 
         # Default, failing back to /prediction on http code 422
         self.prediction_path = "/anomaly/prediction"
@@ -125,6 +123,9 @@ class Client:
         self.n_retries = n_retries
         self.query = f"?format={'parquet' if use_parquet else 'json'}"
         self.target = target
+        self.session = session or requests.Session()
+        self.revision = revision
+
         self.machines = self.get_machines()
 
     @property
@@ -187,6 +188,7 @@ class Client:
 
         # Update session headers, to send the revision we want.
         self.session.headers.update({"revision": self._revision})
+        self.get_metadata(force_refresh=True)
 
     def get_machines(self) -> List[Machine]:
         # Thread safe single access and updating of machines.
--- a/gordo_components/client/client.py
+++ b/gordo_components/client/client.py
@@ -6,6 +6,7 @@ import requests
 import logging
 import itertools
 import typing
+from functools import wraps
 from time import sleep
 from threading import Lock
 from typing import Dict, Any, List, Callable, Optional
@@ -56,7 +57,7 @@ class Client:
         forward_resampled_sensors: bool = False,
         n_retries: int = 5,
         use_parquet: bool = False,
-        session: requests.Session = requests.Session(),
+        session: Optional[requests.Session] = None,
         revision: Optional[str] = None,
     ):
         """
@@ -99,7 +100,7 @@ class Client:
             Pass the data to the server using the parquet protocol. Default is True
             and recommended as it's more efficient for larger batch sizes. If False JSON
             is used for sending the data back and forth.
-        session: requests.Session
+        session: Optional[requests.Session]
             The http session object to use for making requests.
         revision: Optional[str]
             Specific project revision to use when connecting to the server.
@@ -112,8 +113,8 @@ class Client:
         self.prediction_forwarder = prediction_forwarder
         self.data_provider = data_provider
         self.use_parquet = use_parquet
-        self.session = session
         self.project_name = project
+        self.session = session or requests.Session()
         self.revision = revision
 
         # Default, failing back to /prediction on http code 422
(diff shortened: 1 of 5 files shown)


gordo_components/server/utils.py
```diff
@@ -384,8 +384,7 @@ def metadata_required(f):
 
     @wraps(f)
     def wrapper(*args: tuple, gordo_project: str, gordo_name: str, **kwargs: dict):
-        collection_dir = os.environ[current_app.config["MODEL_COLLECTION_DIR_ENV_VAR"]]
-        g.metadata = load_metadata(directory=collection_dir, name=gordo_name)
+        g.metadata = load_metadata(directory=g.collection_dir, name=gordo_name)
```
ryanjdillon: This is sweet.

tests/gordo_components/server/test_gordo_server.py
```diff
@@ -177,12 +180,66 @@ def test_models_by_revision_list_view(caplog, tmpdir, revision_to_models):
...
+            resp = client.get(
+                f"/gordo/v0/test-project/{model_name}/metadata?revision={revision}"
+            )
+            assert resp.status_code == 200
+            assert resp.json["revision"] == revision
+
+            # Verify the server read the metadata.json file we had overwritten
+            assert resp.json["metadata"] == {"revision": revision, "model": model_name}
+
+        # Asking for a revision which doesn't exit gives a 410 Gone.
```
ryanjdillon: > which doesn't exit

tests/gordo_components/server/test_gordo_server.py
```diff
@@ -177,12 +180,66 @@ def test_models_by_revision_list_view(caplog, tmpdir, revision_to_models):
...
+                f"/gordo/v0/test-project/models?revision=revision-does-not-exist"
+            )
+            assert resp.status_code == 410
+            assert resp.json == {
+                "error": "Revision 'revision-does-not-exist' not found."
+            }
+
+
+@pytest.mark.parametrize("revisions", (("123", "456"), ("123",)))
+def test_request_specific_revision(trained_model_directory, tmpdir, revisions):
```
ryanjdillon: Looks good :ok_hand:

gordo_components/client/client.py
```diff
@@ -100,6 +101,9 @@ def __init__(
             is used for sending the data back and forth.
         session: requests.Session
             The http session object to use for making requests.
+        revision: Optional[str]
+            Specific project revision to use when connecting to the server.
+            Defaults to asking the server for the latest revision its capable of serving.
```
ryanjdillon, line 107: small thing, but should be 
> it's capable of server

gordo_components/client/client.py
```diff
@@ -121,6 +126,39 @@ def __init__(
...
+    @revision.setter
+    def revision(self, revision: str):
+        """
+        Verifies the server can satisfy this revision
+        """
+        resp = self.session.get(
+            f"{self.base_url}/gordo/v0/{self.project_name}/revisions"
+        )
+        if resp.ok:
+            if revision:
```
ryanjdillon: This looks like it would always evaluate since it is a required arg to this method, but not familiar with property setting decorators.

ryanjdillon: Ah, it of coarse wouldn't if some null thing was passed, but in that case, should it maybe just error rather than setting the revision to latest?

tests/conftest.py
```diff
@@ -64,6 +64,11 @@ def gordo_project():
     return tu.GORDO_PROJECT
 
 
+@pytest.fixture(scope="session")
+def gordo_revision():
+    return tu.GORDO_REVISION
```
ryanjdillon, line 69: Would be good as a fixture to follow suite with the rest of the test project configuration.

gordo_components/client/client.py
```diff
@@ -126,6 +127,33 @@ def __init__(
...
+        """
+        Wrap a request function to check for 410 status codes, and update
+        the current revision if so.
+        """
+
+        @wraps(f)
+        def _wrapper(*args, **kwargs):
+            resp = f(*args, **kwargs)
+            if resp.status_code == 410:
+                self.revision = None  # Trigger updating revision
```
ryanjdillon: This seems like it may have to do with the injecting of the revision in the headers of success responses. Maybe we could have some detailed docs on what revision business we expect in the various response headers.

ryanjdillon:
LGTM. Up to you on what of the things I commented on should be revised :+1:

ryanjdillon (approved):

gordo_components/server/server.py
```diff
@@ -117,6 +120,36 @@ def build_app():
...
 
+    @app.before_request
+    def _set_revision_and_collection_dir():
+        g.collection_dir = os.environ[
+            current_app.config["MODEL_COLLECTION_DIR_ENV_VAR"]
+        ]
+        g.current_revision = os.path.basename(g.collection_dir)
+
+        # If a specific revision was requested, update collection_dir
+        g.revision = request.args.get("revision") or request.headers.get("revision")
```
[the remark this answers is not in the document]
milesgranger: Good point. Always walking that line of being too strict.. I suppose since we return the revision which was used, it's _maybe_ the user's fault if two were supplied? :man_shrugging:

gordo_components/server/server.py
```diff
@@ -117,6 +120,36 @@ def build_app():
...
+        if g.revision:
+            g.collection_dir = os.path.join(g.collection_dir, "..", g.revision)
+            try:
+                os.listdir(g.collection_dir)  # List dir to ensure it exists
+            except FileNotFoundError:
```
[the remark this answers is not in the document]
milesgranger: Could have done that, but I think we're leaning slightly towards the python's "asking forgiveness rather than permission" style. But I have no strong feelings towards this, aside from that it makes it slightly more explicit in the tests to verify what has happened in certain circumstances.

gordo_components/server/server.py
```diff
@@ -117,6 +120,36 @@ def build_app():
...
+                return make_response(
+                    jsonify({"error": f"Revision '{g.revision}' not found."}), 410
+                )
+        else:
+            g.revision = g.current_revision
+
+    @app.after_request
+    def _revision_used(response):
+        if 200 <= response.status_code <= 299:
```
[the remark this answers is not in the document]
milesgranger: I don't actually remember why I did this.. maybe I didn't want to mess with error messages, but I guess that's a bit silly.

gordo_components/client/client.py
milesgranger: ya.. I'm inclined to agree, a bit too magical that setting `self.revision = None` will result in finding the latest; at best passing 'latest' or method to update to latest. I'd guess I'll opt for the former.

milesgranger:
Ran through test project JIC. :-)

PR #819, opened 2019-12-24
Add dataclasses for machine.metadata

milesgranger:


gordo_components/client/client.py
```diff
@@ -407,6 +410,7 @@ def _send_prediction_request(
             }
 
         # Start attempting to get predictions for this batch
+
```
EinarJohnsen: Newlines are the best :sunglasses:

EinarJohnsen:
I like this!

gordo_components/builder/build_model.py
```diff
@@ -121,15 +126,16 @@ def build(
...
 
             # Otherwise build and cache the model
             else:
-                model, metadata = self._build()
+                model, machine = self._build()
                 self.cached_model_path = self._save_model(
-                    model=model,
-                    metadata=metadata,
-                    output_dir=output_dir,  # type: ignore
+                    model=model, metadata=machine, output_dir=output_dir  # type: ignore
```
ryanjdillon, line 141: Maybe a bit much for this PR, but if `machine` is now synonymous with `metadata`, perhaps renaming things accordingly would be good here, if not just the internal methods for now.

gordo_components/builder/build_model.py
```diff
@@ -121,15 +126,16 @@ def build(
             if self.cached_model_path:
                 model = serializer.load(self.cached_model_path)
                 metadata = serializer.load_metadata(self.cached_model_path)
-                metadata["metadata"]["user-defined"] = self.machine.metadata
+                metadata["metadata"][
+                    "user_defined"
+                ] = self.machine.metadata.user_defined
```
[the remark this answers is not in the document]
ryanjdillon: Like in L184-L191, wouldn't need to do that without new *user_defined* metadata.

gordo_components/machine/metadata/metadata.py
```diff
@@ -0,0 +1,42 @@
...
+@dataclass
+class CrossValidationMetaData:
+    scores: Dict[str, Any] = field(default_factory=dict)
+    cv_duration_sec: Optional[float] = field(default=None)
+    splits: Dict[str, Any] = field(default_factory=dict)
+
+
+@dataclass_json
+@dataclass
+class ModelBuildMetadata:
```
ryanjdillon, line 20: It seems like once when we create these they could be immutable after that, in which case you could use the [`frozen`](https://docs.python.org/3/library/dataclasses.html#module-level-decorators-classes-and-functions) arg to the decorator.

gordo_components/machine/metadata/metadata.py
```diff
@@ -0,0 +1,42 @@
...
+        default_factory=CrossValidationMetaData
+    )
+    model_training_duration_sec: Optional[float] = field(default=None)
+    model_meta: Dict[str, Any] = field(default_factory=dict)
+
+
+@dataclass_json
+@dataclass
+class BuildMetadata:
+    model: ModelBuildMetadata = ModelBuildMetadata()
```
flikka: Should this use field(default_factory=ModelBuildMetadata)? Asking for a friend.

gordo_components/machine/metadata/metadata.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+@dataclass_json
+@dataclass
+class Metadata:
+    user_defined: Dict[str, Any] = field(default_factory=dict)
+    build_metadata: BuildMetadata = BuildMetadata()
```
flikka: Same question as above... (default_factory?)

gordo_components/builder/build_model.py
milesgranger: ya.. I was on the fence about renaming as well and trying to minimize changes.. :man_shrugging:

flikka (changes requested):
Looks nice, just had a couple of questions around the dataclass field things and how / when the defaults are created.

gordo_components/builder/build_model.py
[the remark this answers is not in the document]
milesgranger: It seems, if that is the case, that's a larger change outside of this PR. As user defined metadata has always been part of the concept of "metadata" in the project. Not saying it shouldn't be, but I'm not clear on if _that_ change should be within the scope of converting metadata as a dict to a dataclass as we're doing here.

gordo_components/machine/metadata/metadata.py
milesgranger: As we talked about, lower branches of the metadata such as `CrossValMetadata` and similar could maybe, in some way, benefit from `frozen` but it wasn't at the top level nodes because we naturally (as of now) inject updates to the metadata as it flows through the process.

But I have no real opinion or grievance about adding it for other parts.

gordo_components/machine/metadata/metadata.py
milesgranger: Nice catch! :fishing_pole_and_fish:

gordo_components/builder/build_model.py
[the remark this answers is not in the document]
ryanjdillon: Agreed :+1:

gordo_components/machine/metadata/metadata.py
ryanjdillon: Though I'm not strongly opinionated about it either, it make sense to me. I would think that we probably don't want to alter the metadata for the cross-validation after it's occurred (for example), so if there is something that needs to be mutable down the line it would be good to have it made so explicitly. We can always revisit this later.

gordo_components/builder/build_model.py
ryanjdillon: I'll make an issue for it.

gordo_components/machine/metadata/metadata.py
milesgranger: Yep, like we talked about, seems dataclass inherit their owner's frozen setting. Maybe we could revisit this later if we really want to make the whole thing frozen.

milesgranger:
Ran through test project without trouble. :rainbow: :rabbit:

flikka (approved):

PR #821, opened 2019-12-30
Replace CircleCI with Github Actions

milesgranger:
Should be enough to make the switch, but can be made more fancy by pulling out the deploy steps, tests, image and docs building into their own workflows later if we want.

Benefits :unicorn: 
-----------
- Super easy matrix config, can easily run tests on different OS, python version, ...
- Better concurrency... all tests for all components start right away.. unlike CircleCI limit of 4 containers
- Less YAML

Cons :carousel_horse: 
-------
- CodeCov **cannot** evaluate coverage on PRs b/c running on platforms other than CircleCI, AppVeyor or TravisCI requires a token which cannot be shared with forks.
  - Releated issue: https://github.com/codecov/codecov-action/issues/29
  - > Note: Token not required for public repositories uploading from Travis, CircleCI or AppVeyor.

Suppose we can store this PR until it is supported?


.github/workflows/main.yml
```diff
@@ -0,0 +1,140 @@
...
+      - uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
+          architecture: 'x64'
+
+      - uses: actions/cache@v1
+        if: startsWith(runner.os, 'Linux')
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-${{ matrix.python-version }}-pip-${{ hashFiles('**/requirements.txt') }}
```
epa095: When does this save the cache? Magically after the steps are done? And should we add the test req file as well?

epa095 (commented):
Definitely nicer than the circle setup

.github/workflows/main.yml
milesgranger: It does auto add the save step after the last job completes successfully. 
Add a wildcard for all `*requirements.txt` files to be included.

EinarJohnsen:
Looks cool!

flikka (approved):
Modern

PR #822, opened 2020-01-02
Replace tests/utils globals with fixtures

### LINKED ISSUE #803: Replace test_utils imports with fixtures
Currently there are tests that import the default test proeject parameters from `tests/test_utils.py`, while those are also fixture'ized in `test/conftest.py`. 

Ideally the tests would all use the fixtures and `test_utils` would only be imported in `conftest.py`.

ryanjdillon:
# Problem :no_mouth: 
Parameters used for the configuration of a test project, used across a number of tests, are currently defined as globals in `tests/utils.py` and imported into a few test modules.

# Solution :lips: 
To make this improvement, the following must be done:

* Make new fixtures that return these parameters
* Replace the use of the globals to the new fixtures
* Refactor tests that used the globals in a `parametrize` decorator for use with the fixtures
* Remove redundant fixtures and methods

closes #803

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,25 +1,35 @@
 # -*- coding: utf-8 -*-
 
 import asyncio
-import os
+import io
+import json
 import logging
+import os
+import re
 import tempfile
+from threading import Lock
 from typing import List
 from unittest.mock import patch
 
+import docker
+from flask import Request
 import pytest
+import responses
 
 from gordo_components import serializer
 from gordo_components.machine.dataset.sensor_tag import SensorTag
-from gordo_components.machine.dataset.data_provider.providers import InfluxDataProvider
+
 from gordo_components.server import server
 from gordo_components.builder.local_build import local_build
 from gordo_components.machine.dataset import sensor_tag
+from gordo_components.machine.dataset.sensor_tag import to_list_of_strings
+from gordo_components.server import server as gordo_ml_server
 
 from tests import utils as tu
 
-
 logger = logging.getLogger(__name__)
+
+TEST_SERVER_MUTEXT = Lock()
 
 
 def pytest_collection_modifyitems(items):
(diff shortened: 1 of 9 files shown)


tests/gordo_components/server/test_base_view.py
```diff
@@ -8,42 +8,46 @@
...
+
+    if data_generation_method != "np_list":
+        df = pd.DataFrame(data_to_post, columns=sensors_str)
+        if data_generation_method == "dataframe_records":
+            data_to_post = df.to_dict("records")
+        elif data_generation_method == "dataframe_list":
+            data_to_post = df.to_dict("list")
+        elif data_generation_method == "dataframe_dict":
+            data_to_post = df.to_dict("dict")
+
```
milesgranger: I'm for the change, but maybe just pass the actual argument for `to_dict(...)`, and have that variable be `None` if it the data should be left as a list.

tests/gordo_components/server/test_base_view.py
```diff
@@ -69,19 +73,23 @@ def test_prediction_endpoint_post_ok(
...
     ],
 )
 def test_prediction_endpoint_post_fail(
-    caplog, base_route, sensors, gordo_ml_server_client, data_to_post
+    caplog, base_route, sensors, gordo_ml_server_client, error_generator
 ):
     """
     Test expected failures when posting certain types of data
     """
+    data_to_post = {"X": error_generator(np.random.random((len(sensors), 10)))}
```
milesgranger: I find this harder to read, and it takes more lines..

milesgranger (commented):
Looking nice!

tests/gordo_components/server/test_base_view.py
ryanjdillon: I liked it a bit better before too. This was done to address not being able to pass fixtures to the `parametrize` decorater.

ryanjdillon: My thinking here was that we'd start with a valid `X` and some mutation to it yields the erroneous data being tested.

milesgranger: IMO it 'breaks' part of the test, for example 'No X' had a _key_ other than `X` now it puts the _value_ as `True`. I still don't see why this is necessary.

ryanjdillon: This is necessary (but could of course be wrong here) because I can no longer call the global `tu.SENSORS_STR_LIST` to the `pytest.mark.parametrize` decorator.  I need to now pass the `sensors_str` fixture to the test, and then generate the data in the test. 

I can fix the key bit easily though.

milesgranger: you don't _need_ to have `tu.SENSORS_STR_LIST` though, right? I mean, having an empty list is clearly going to be 'not enough features' or even make it 10,000 features, because we surely won't have that either.

ryanjdillon: Ok, have a simple solution that just omits the parametrize...

milesgranger: Cool, LGTM, I'll approve once tests are passing. :+1:

tests/conftest.py
```diff
@@ -116,21 +170,21 @@ def config_str(gordo_name: str, sensors: List[SensorTag]):
...
+    model_collection_directory: str, config_str: str, gordo_single_target: str
 ):
     """
     Fixture: Train a basic AutoEncoder and save it to a given directory
     will also save some metadata with the model
     """
 
     # Model specific to the model being trained here
-    model_dir = os.path.join(model_collection_directory, gordo_name)
+    model_dir = os.path.join(model_collection_directory, gordo_single_target)
```
ryanjdillon: One thing worth mentioning here... `gordo_single_target` is just the gordo name, but when used in the client/server tests makes sense contextually to be called `gordo_single_target`, but in other places this feels to me to be slightly ambiguous. 

I decided to just use `gordo_single_target` throughout, but perhaps we want to change it to `gordo_name`.

flikka: Yeah, I found this name change a bit confusing maybe, for the other (non client/server) places.

milesgranger (approved):
Most excellent! :clap:

tests/conftest.py
ryanjdillon: Do you think I should maybe alias `gordo_name` to `gordo_single_target`, or swap out `gordo_single_target` for `gordo_name?

ryanjdillon: I've now added the `gordo_name` fixture and an alias to a `gordo_single_target` fixture :+1:

PR #823, opened 2020-01-03
Use try block in reading metadata

### LINKED ISSUE #780: load_metadata: Try loading and handle exception
Today we do
```python
        if path.isfile(possible_path):
            with open(possible_path, "r") as f:
                return simplejson.load(f)
```
but this is a bit unpythonic, it is more common to just try to read and handle the exception. For our specific case that also has the advantage that we get fewer IOPs, which is something we need to limit a bit.

milesgranger:
Will close #780


EinarJohnsen (approved):
LGTM, as long as tests pass :passport_control:

PR #827, opened 2020-01-06
Change remaining GORDO_PROJECT_VERSION -> GORDO_PROJECT_REVISION

milesgranger:
It seems #804 forgot to update a few `GORDO_PROJECT_VERSION` references in the workflow. They appear to just be used for passing env var to scripts/args so had no real affect.


EinarJohnsen (approved):
Good catch, LGTM :100:

PR #840, opened 2020-01-08
Add testformatting and allelse

EinarJohnsen:
Currently we are not testing formatting and allelse in the workflow on Github actions, this adds those to the matrix.


milesgranger (approved):
Nice catch! Merge when tests pass :crossed_fingers:

PR #849, opened 2020-01-09
Use pytest tmpdir fixture

### LINKED ISSUE #799: Replace `tmp_dir` fixture for pytest's `tmpdir` fixture
Pytest's built-in `tmp_path` and `tmpdir` fixtures can be used for the same purposes as ours and are also function scoped.

Using them instead will remove a few lines of code.

milesgranger:
Will close #799
Remove own implementation in favor of pytest's

### DIFF
--- a/tests/gordo/cli/test_cli.py
+++ b/tests/gordo/cli/test_cli.py
@@ -48,40 +48,40 @@ def machine():
 
 
 @pytest.fixture
-def runner(tmp_dir):
-    mlflow.set_tracking_uri(f"file:{tmp_dir}")
+def runner(tmpdir):
+    mlflow.set_tracking_uri(f"file:{tmpdir}")
     yield CliRunner()
 
 
-def test_build_env_args(runner, tmp_dir, machine):
+def test_build_env_args(runner, tmpdir, machine):
     """
     Instead of passing OUTPUT_DIR directly to CLI, should be able to
     read environment variables
     """
     logger.info(f"MODEL_CONFIG={json.dumps(MODEL_CONFIG)}")
 
-    with temp_env_vars(MACHINE=json.dumps(machine.to_dict()), OUTPUT_DIR=tmp_dir):
+    with temp_env_vars(MACHINE=json.dumps(machine.to_dict()), OUTPUT_DIR=str(tmpdir)):
         result = runner.invoke(cli.gordo, ["build"])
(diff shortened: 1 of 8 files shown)


tests/gordo/cli/test_cli.py
```diff
@@ -241,11 +243,7 @@ def test_build_exit_code(exception, exit_code, runner, tmp_dir, machine):
     [(True, {"cv_mode": "full_build"}), (False, {"cv_mode": "cross_val_only"})],
 )
 def test_build_cv_mode(
-    runner: CliRunner,
-    should_save_model: bool,
-    cv_mode: str,
-    tmp_dir: str,
-    machine: Machine,
+    tmpdir, runner: CliRunner, should_save_model: bool, cv_mode: str, machine: Machine
```
EinarJohnsen, line 246: Could have a type hint for the squint :squid:

tests/gordo/cli/test_cli.py
```diff
@@ -283,11 +281,11 @@ def test_build_cv_mode(
     ],
 )
 def test_build_cv_mode_cross_val_cache(
+    tmpdir,
```
EinarJohnsen, line 284: ^ :squid:

EinarJohnsen (approved):
Just a typehint comment, otherwise it looks good :accept:

flikka (changes requested):
Should this line in the conftest also use tmpdir perhaps? I just saw it because I wondered why the `import tempfile` was still in the conftest.py file.

This line being: https://github.com/equinor/gordo/blob/ab14ca8f7167be528096df1ee0eef2972e804e84/tests/conftest.py#L134

milesgranger:
@flikka 
1. Good eye
2. That is on purpose, as _that_ fixture is valid for the session, whereas `tmpdir` is valid per test ran, so they cannot be used together.

tests/gordo/cli/test_cli.py
milesgranger: blah, I purposely didn't do that because it's not in `typing` but rather import from pathlib I think, and would require importing that in every file which uses `tmpdir` as a fixture.. maybe we should though. Your thoughts @flikka ?

flikka:
Aha, I see

flikka (approved):

PR #851, opened 2020-01-13
Precheck workflow generate produces valid yaml

### LINKED ISSUE #712: Check generated argo tempelate is valid YAML
Argo lint didn't catch a fairly hard to debug error I made in the argo workflow template. It would be great to have things setup so that:

1. the argo template fields are populated and saved to yaml in a `tmp_dir`
2. `assert yaml.load(tmp_dir/tmp.yml)` to check that the YAML can be converted to JSON
3. Then build the docker image and reference that workflow template for testing `gordo-components workflow`

**the mistake**
```yaml
              - name: enable-remote-logging
                value: |
                  {{ machine.runtime.builder.remote_logging.enable | tojson}}
```
Which yields `false`, not `"false"` or `False`, which is not caught by argo lint., in addition to there being an erroneous inserted tab. This resulted in 
**the error**
```bash
FATA[0000] /tmp/out.yaml failed to parse: error converting YAML to JSON: yaml: line 1285: found a tab character where an indentation space is expected 
```
In order to debug this, I needed to interactively run the gordo-deploy image, run the command the test failed on to generate `/tmp/out.yaml`, and then `sed -n 1280,1295p /tmp/out.yaml` the lines to see where things were going wrong.

So it would be great if this pre-check test could just point to the erroneous line in the argo template.

milesgranger:
Will close #712

### DIFF
--- a/tests/gordo/workflow/test_workflow_generator/test_workflow_generator.py
+++ b/tests/gordo/workflow/test_workflow_generator/test_workflow_generator.py
@@ -122,6 +122,14 @@ def test_argo_lint(argo_docker_image, fp_config, docker_client, repo_dir):
     Test the example config files, assumed to be valid, produces a valid workflow via `argo lint`
     """
 
+    # Verify doing a workflow generation on the config will generate valid yaml
+    # if it's not, running argo lint will generate wildly unhelpful error msgs.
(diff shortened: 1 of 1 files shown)


epa095 (approved):
LGTM

PR #852, opened 2020-01-13
Add high-priority to server deployment pod

### LINKED ISSUE #637: Add pod priority to server/watchman
K8s has a concept of pod-priority, where high priority pods are scheduled earlier, and kept allive longer than the meager pessant pods. We can set `podPriorityClassName: high-priority` on important pods, like the server and watchman (and maybe only them).

milesgranger:
Will close #637 

Doc ref: https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#priorityclass

### DIFF
--- a/gordo/workflow/workflow_generator/resources/argo-workflow.yml.template
+++ b/gordo/workflow/workflow_generator/resources/argo-workflow.yml.template
@@ -855,6 +855,7 @@ spec:
                         labels:
                           app: "{{'{{inputs.parameters.host-name}}'}}"
                       spec:
+                        priorityClassName: high-priority
                         containers:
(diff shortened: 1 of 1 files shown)


EinarJohnsen:
I see that `Non-preempting PriorityClasses (alpha)` is in alpha, but is that something we want to consider?
I am a bit uncertain how the scheduling works if you start two nearly simultaneously projects, but perhaps setting the `PreemptionPolicy: Never` to avoid preemting running pods?

milesgranger:
:arrow_up:  @epa095

milesgranger:
@epa095 tried the test project with a priority name which didn't exist and the server was scheduled without any trouble. :+1:

epa095 (approved):

PR #854, opened 2020-01-13
Make metadata/machine use more consistent. 

### LINKED ISSUE #830: Make metadata/machine naming consistent
As some post-refactoring pruning, parameters named `metadata` in places around `build_model` and friends can now be updated to `machine`. 

There is probably also some other small tid-bits that could be renamed similarly in way to make things more intuitive here while one is at it.

milesgranger:
Will close #830 

We use to return a semi-arbitrary 'metadata' object from `ModelBuilder.build` but now we return a complete `Machine`

Seems like good housekeeping to rename relevant variables/functions according to this.

### DIFF
--- a/gordo/builder/build_model.py
+++ b/gordo/builder/build_model.py
@@ -63,7 +63,7 @@ class ModelBuilder:
         ...     project_name='test-proj',
         ... )
         >>> builder = ModelBuilder(machine=machine)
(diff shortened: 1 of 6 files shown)


ryanjdillon (approved):
Looks good :+1: :tada:

PR #863, opened 2020-01-14
Convert server-to-sql into a `Reporter`

### LINKED ISSUE #865: Create `Reporter` API for reporting/logging built machines
We report/log Machine's two different ways in two different places (mlflow and "server-to-sql")

Should rather have a common interface for handling this; start with moving the postgres script in the workflow, then probably ML Flow after.

milesgranger:
Will close #865 

Converts the `server-to-sql` to a reporter called `PostgresReporter` along with the initial structure of having `reporters` and how they fit into the machine building process.


gordo/machine/machine.py
```diff
@@ -118,3 +118,13 @@ def to_dict(self):
...
+
+    def report(self):
+        """ 
+        Run any reporters in the machine's runtime for the current state.
+        """
+        # Avoid circular dependency with reporters which import Machine
+        from gordo.reporters.base import BaseReporter
+
+        for reporter in map(BaseReporter.from_dict, self.runtime["reporters"]):
+            reporter.report(self)
```
epa095, line 135: What about the "crash-on-failure-or-not" thing?

milesgranger: tja, upon implementing it felt quite strange to generally 'try/except' each and then for `Machine` to care about how to handle exceptions by checking `Reporter` for some 'thing'. 

Rather, IMO, handle it during the implementation of a `Reporter`. As such, if it should be allowed to fail, then _it_ should take some 'hardfail' flag as a parameter and set per config entry.

gordo/cli/workflow_generator.py
```diff
@@ -181,7 +178,23 @@ def workflow_generator_cli(gordo_ctx, **ctx):
...
     # for it?"
-    context["enable_influx"] = nr_of_models_with_clients > 0
+    enable_influx = nr_of_models_with_clients > 0
+    context["enable_influx"] = enable_influx
+
+    context["postgres_host"] = f"gordo-postgres-{config.project_name}"
+
+    # If enabling influx, we setup a postgres reporter to send metadata
+    # to allowing querying about the machine from grafana
+    if enable_influx:
```
ryanjdillon, line 188: Maybe we want to sometimes enable postgres and not influx? Could be `if enable_influx or enable_postgres:`

gordo/machine/validators.py
```diff
@@ -133,9 +133,22 @@ class ValidMachineRuntime(BaseDescriptor):
...
+        self._verify_reporters(value)
         value = fix_runtime(value)
         instance.__dict__[self.name] = value
 
+    def _verify_reporters(self, value: dict):
+        """
+        Verify the expected existence and structure of runtime.reporters
+        """
+        if "reporters" not in value:
+            value["reporters"] = list()
```
ryanjdillon: It seems off to modify the dict during validation. Could the runtime be a dataclass that would then enforce the empty list if one is not passed when the runtime "data" is initialized? That way there would be a one-stop-shop for looking at what are the requisite properties of `runtime` while avoiding an ambiguous injection of the field when missing. 

This of course could/should be another PR if it means some bigger overhauling.

gordo/reporters/postgres.py
```diff
@@ -0,0 +1,81 @@
...
+logging.basicConfig(level=logging.DEBUG)
+logger = logging.getLogger(__name__)
+
+
+db = PostgresqlExtDatabase(None)
+
+
+class PostgresReporter(BaseReporter):
+
+    db = db
```
ryanjdillon: A docstring could be good to describe the specifics of this interface.

gordo/workflow/config_elements/normalized_config.py
```diff
@@ -25,6 +25,7 @@ class NormalizedConfig:
 
     DEFAULT_CONFIG_GLOBALS = {
         "runtime": {
+            "reporters": [],
```
ryanjdillon, line 28: Ok, so it gets patched from here, so the validation insertion is just another check? It does seem a data class would be pretty nice, as it would address the patching stuff too.

gordo/workflow/workflow_generator/resources/argo-workflow.yml.template
```diff
@@ -451,9 +451,9 @@ spec:
                   apiVersion: v1
                   kind: Service
                   metadata:
-                    name: gordo-postgres-{{project_name}}
+                    name: {{postgres_host}}
```
ryanjdillon, line 477: Could we possibly ever want to have multiple postgres hosts? Maybe if we were to change the schema and do a canary, blue/green deployment or something? No change request here... just (possibly dumb) discussion.

tests/conftest.py
```diff
@@ -214,6 +214,28 @@ def gordo_ml_server_client(
...
+            remove=True,
+            detach=True,
+        )
+        import time
+
+        time.sleep(5)
+        yield
+    finally:
+        if postgres is not None:
+            postgres.kill()
```
ryanjdillon: This looks like it might work the same without the try/finally.

tests/gordo/machine/test_machine.py
```diff
@@ -0,0 +1,20 @@
+import pytest
+import peewee
+
+from gordo.machine import Machine
+from gordo.reporters.postgres import PostgresReporter, Machine as PostgresMachine
```
ryanjdillon, line 5: Maybe good to have `postgres.Machine` just be `postgres.PostgresMachine` on the regular?

ryanjdillon (commented):
I like it!

gordo/cli/workflow_generator.py
milesgranger: That seems outside the scope of this PR.

gordo/machine/validators.py
milesgranger: I don't think that because the dict is missing a key makes it invalid, but I agree, the runtime might be nice to have as a dataclass, and would address this anyhow, but should be addressed outside of this PR IMO.

gordo/workflow/workflow_generator/resources/argo-workflow.yml.template
milesgranger: Again, I think that is outside the scope of what this PR is doing. I think it's an entirely valid consideration, however.

tests/conftest.py
milesgranger: That's true if decorated with [@pytest.yield_fixture](https://pytest.readthedocs.io/en/2.8.7/yieldfixture.html), which this should probably be, but it'll be functionally the same. As it is now, a `KeyboardInterrupt` wouldn't guarantee the container would be killed; but from the docs it seems `yield_fixture` would in the same way... :+1:

tests/gordo/machine/test_machine.py
milesgranger: I think the point here is that `Machine` to a `Reporter` is independent of our interpretation. It's a database model representing a gordo machine. In that respect, it seems silly to mandate that users would prefix their database models with name of their database providers. I admit it's slightly awkward to import for a _test_ but it's not used anywhere inside the core of gordo.

gordo/workflow/workflow_generator/resources/argo-workflow.yml.template
```diff
@@ -1125,14 +1093,10 @@ spec:
           cpu: 200m
 
 
-  - name: sql-and-cleanup
+  - name: workflow-cleanup
```
epa095, line 1119: Remember that now the model-builder should depend on the postgres step in the workflow (if "influx is enabled").

gordo/reporters/base.py
```diff
@@ -0,0 +1,61 @@
...
+
+from gordo.machine import Machine
+
+
+class BaseReporter(abc.ABC):
+    @abc.abstractmethod
+    def report(self, machine: Machine):
+        """Report/log the machine"""
+
+    def to_dict(self) -> dict:
```
epa095, line 17: Can we not use the serializer here? (and with the load?)

gordo/builder/build_model.py
```diff
@@ -287,6 +288,7 @@ def _build(self) -> Tuple[sklearn.base.BaseEstimator, Machine]:
                 model_meta=self._extract_metadata_from_model(model),
             )
         )
+        machine.report()
```
epa095: This just rubs me the wrong way. I don't know why really. It is just a bit weird that a machine reports itself. I dont know what the alternatives are, and if they are nicer:
- Have the for loop here in build_model (maybe as a function),
- Have the for loop in `gordo/reporters/base.py `, maybe as a class-function?

gordo/reporters/base.py
milesgranger: Lovely.

gordo/workflow/workflow_generator/resources/argo-workflow.yml.template
milesgranger: Nice catch, safety third! :3rd_place_medal:

epa095:
Whats the status on this. Does it work now that we fixed the server?

milesgranger:
...debugging. I'll change it back to WIP. :+1:

milesgranger:
@epa095 Tested on minikube :ballot_box_with_check:

epa095 (approved):
LGTM

PR #864, opened 2020-01-14
Add possibility to set n_splits

### LINKED ISSUE #800: Un-hardcode number of splits for TimeSeriesSplit
In the model builder, the number of splits are currently hard-coded to `3`. This could be set using the `evaluation_config`, so some logic could be added there to allow for this. 

i.e.
```python
 n_splits = evaluation_config["n_splits"] if "n_splits" in evaluation_config else 3
 split_obj = TimeSeriesSplit(n_splits= n_splits)
```

EinarJohnsen:
n_splits was hardcoded to 3, but it is now possible to set in the config file under the evaluation key, it can now be set as 
```
evaluation:
  cv:
    sklearn.model_selection.TimeSeriesSplit: 
      n_splits: 5
```
for either a single machine or in the globals for all of them. Closes #800

### DIFF
--- a/gordo/builder/build_model.py
+++ b/gordo/builder/build_model.py
@@ -18,7 +18,7 @@ import tensorflow as tf
 import sklearn
 from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn import metrics
-from sklearn.model_selection import BaseCrossValidator, cross_validate, TimeSeriesSplit
+from sklearn.model_selection import BaseCrossValidator, cross_validate
 from sklearn.pipeline import Pipeline
 
 from gordo.util import disk_registry
@@ -212,8 +212,13 @@ class ModelBuilder:
                 scaler = self.machine.evaluation.get("scoring_scaler")
                 metrics_dict = self.build_metrics_dict(metrics_list, y, scaler=scaler)
 
+                split_obj = serializer.pipeline_from_definition(
+                    self.machine.evaluation.get(
+                        "cv",
+                        {"sklearn.model_selection.TimeSeriesSplit": {"n_splits": 3}},
+                    )
+                )
                 # Generate metadata about CV train, test splits
-                split_obj = TimeSeriesSplit(n_splits=3)
                 split_metadata = ModelBuilder.build_split_dict(X, split_obj)
 
                 cv_kwargs = dict(
--- a/tests/gordo/builder/test_builder.py
+++ b/tests/gordo/builder/test_builder.py
@@ -13,6 +13,7 @@ import sklearn.ensemble
 from sklearn.multioutput import MultiOutputRegressor
 from sklearn.linear_model import LinearRegression
 from sklearn import metrics
+from mock import patch
 
 import gordo
 from gordo.builder import ModelBuilder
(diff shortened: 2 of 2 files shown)


gordo/builder/build_model.py
```diff
@@ -212,8 +212,20 @@ def _build(self) -> Tuple[sklearn.base.BaseEstimator, Machine]:
                 scaler = self.machine.evaluation.get("scoring_scaler")
                 metrics_dict = self.build_metrics_dict(metrics_list, y, scaler=scaler)
 
+                try:
+                    n_splits = int(
+                        self.machine.evaluation["cv"][
+                            "sklearn.model_selection.TimeSeriesSplit"
+                        ]["n_splits"]
+                    )
```
milesgranger: Think we should might as well allow complete parameterization here with something like:
```python
serializer.pipeline_from_definition(self.machine.evalutation.get("cv", some_default))
```

EinarJohnsen: I see, I will fix this.

milesgranger (commented):
Looks good overall, but the premise of the test seems to be rather testing that the serializer works as expected (ie, getting the right amount of `n_splits`)

Do you think it may be better to have a very similar test instead which parametrizes over splitters and then uses mock to `assert split_method.called(...)` type pattern? To me, this may be more to the point as to what the test is ensuring. Basically, during the build process the `SomeSplitter.split` was called, and no one inadvertently hard codes a specific splitter. As serializer itself is, assumed to be, tested enough it's own that we don't necessarily need to verify that process.

tests/gordo/builder/test_builder.py
```diff
@@ -642,3 +643,38 @@ def test_setting_seed(seed, model_config):
...
+        {"sklearn.model_selection.TimeSeriesSplit": {"n_splits": 5}},
+        {
+            "sklearn.model_selection.TimeSeriesSplit": {
+                "n_splits": 5,
+                "max_train_size": 10,
+            }
+        },
+        {"sklearn.model_selection.ShuffleSplit": {"n_splits": 5}},
+    ),
+)
```
ryanjdillon: Sweet :+1:

tests/gordo/builder/test_builder.py
```diff
@@ -642,3 +643,38 @@ def test_setting_seed(seed, model_config):
...
+@patch("gordo.serializer.pipeline_from_definition")
+def test_n_splits_from_config(MockClient, cv):
+    data_config = get_random_data()
```
ryanjdillon: Maybe a docstring on this test's purpose would be good.

ryanjdillon (approved):

tests/gordo/builder/test_builder.py
```diff
@@ -642,3 +643,38 @@ def test_setting_seed(seed, model_config):
...
+            "sklearn.model_selection.TimeSeriesSplit": {
+                "n_splits": 5,
+                "max_train_size": 10,
+            }
+        },
+        {"sklearn.model_selection.ShuffleSplit": {"n_splits": 5}},
+    ),
+)
+@patch("gordo.serializer.pipeline_from_definition")
+def test_n_splits_from_config(MockClient, cv):
```
milesgranger: this is nit picking, I admit, but maybe rename this to `mocked_pipeline_from_definition` ?

milesgranger (approved):
Looks legit! One minor comment. :+1:

PR #872, opened 2020-01-15
Adding PVC to grafana, fixing persistence issue

### LINKED ISSUE #871: bug: Grafana suddenly empty
Sometimes grafana is suddenly empty, no datasource and no dashboard. It seems to happen when the cluster scales down, so maybe the pod is recreated even though it is a statefulset. I think the solution is to have PVC mounted into the pod, just like with influx.

epa095:
This closes #871 

Not tested since the clusters are a bit broken ATM.

Edit: Tested on minikube!

### DIFF
--- a/gordo/workflow/workflow_generator/resources/argo-workflow.yml.template
+++ b/gordo/workflow/workflow_generator/resources/argo-workflow.yml.template
@@ -358,6 +358,26 @@ spec:
           {% endif %}
         spec:
           serviceName: gordo-grafana-{{project_name}}
+          volumeClaimTemplates:
+            - metadata:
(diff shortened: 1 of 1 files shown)


milesgranger:
I have a feeling worfkow docker test started failing after we merged an upgrade to websocket. 
but maybe bumping the timeout a bit more might save it; locally it always ends up finishing, and re-running the tests will usually make it work again :man_shrugging:

milesgranger (dismissed):
Looks good enough to me.

milesgranger (approved):
> Tested on minikube

How delightful! :1st_place_medal:

PR #874, opened 2020-01-15
Update serializer & fix KerasRawModelRegressor

### LINKED ISSUE #873: Fix broken KerasRawModelRegressor after tensorflow >= 2.1
Tensorflow 2.1 moved `lr` from `Model.compile` and one will rather place that into the given optimizer:
```python
model.compile(optimizer=tf.keras.optimizers.SGD(learning_rate=0.001), ...)
```

Ok enough, but need to load the compile kwargs using the serializer, and will probably need to update that to allow such a thing.

milesgranger:
Will close #873 
Support loading compile kwargs which may have loadable objects in the values.

..should probably rename `serializer.pipeline_from_definition` because we're so darn generic. :stuck_out_tongue_winking_eye:


epa095 (approved):
LGTM!!

PR #879, opened 2020-01-17
Optimize argo lint of generated workflows

milesgranger:
Instead of building our own docker image which tends to have timeout issues, we can use `argocli` image and mount in a workflow generated on the host machine.

Seems to reduce the argo lint tests from ~5mins to 5 seconds. :turtle::dash:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -272,6 +272,17 @@ def influxdb(base_influxdb):
     base_influxdb.reset()
 
 
+@pytest.fixture(scope="session")
+def argo_version(repo_dir):
+    with open(os.path.join(repo_dir, "Dockerfile-GordoDeploy")) as f:
+        match = next(re.finditer(r'ARGO_VERSION="(\w\d+.\d+.\d+)"', f.read()), None)
+    if match is None:
+        raise LookupError(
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -272,6 +272,17 @@ def influxdb(base_influxdb):
     base_influxdb.reset()
 
 
+@pytest.fixture(scope="session")
+def argo_version(repo_dir):
+    with open(os.path.join(repo_dir, "Dockerfile-GordoDeploy")) as f:
+        match = next(re.finditer(r'ARGO_VERSION="(\w\d+.\d+.\d+)"', f.read()), None)
```
epa095, line 278: > Some people, when confronted with a problem, think "I know, I'll use regular expressions." Now they have two problems.

Just kidding. Love it. Or at least, I accept it and will let it live in my house.

epa095 (approved):

tests/conftest.py
milesgranger: haha, I've heard that. :laughing:  I've also heard Facebook doesn't allow regex in their code base, but I'm not certain how true that is.

PR #880, opened 2020-01-19
Add InsufficientDataAfterRowFilteringError to row_filter

EinarJohnsen:
The new error is raised if the row_filtering ends up with less (or equal) rows than the n_samples_threshold. Used to differentiate between not enough data and the row_filtering beeing to strict, removing to many values.

### DIFF
--- a/gordo/machine/dataset/datasets.py
+++ b/gordo/machine/dataset/datasets.py
@@ -32,6 +32,10 @@ class InsufficientDataError(ValueError):
     pass
 
 
+class InsufficientDataAfterRowFilteringError(ValueError):
+    pass
+
+
 def compat(init):
     """
     __init__ decorator for compatibility where the Gordo config file's ``dataset`` keys have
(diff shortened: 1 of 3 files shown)


gordo/machine/dataset/datasets.py
```diff
@@ -220,6 +224,13 @@ def get_data(self) -> Tuple[pd.DataFrame, Optional[pd.DataFrame]]:
                 data, self.row_filter, buffer_size=self.row_filter_buffer_size
             )
 
+            if len(data) <= self.n_samples_threshold:
+                raise InsufficientDataAfterRowFilteringError(
+                    f"The length of the genrated DataFrame ({len(data)}) does not exceed the"
+                    f"specified required threshold for the number of rows ({self.n_samples_threshold}),"
+                    f"and is caused by the row-filter."
```
milesgranger: Think you want to add a space at the end of each of these lines.

milesgranger (approved):
Seems reasonable to me. Just the one comment. :open_hands:

epa095 (approved):
LGTM!!!

PR #881, opened 2020-01-20
Remove gevent as explicit dependency

epa095:
Instead change gunicorn to gunicorn[gevent] and let that handle it


requirements.txt
```diff
@@ -37,23 +37,24 @@ cryptography==2.8         # via adal, azureml-core, pyopenssl, secretstorage
 cycler==0.10.0            # via matplotlib
 databricks-cli==0.9.1     # via mlflow
 dataclasses-json==0.3.7
+dataclasses==0.7          # via dataclasses-json
```
epa095: @milesgranger  do you know why it suddenly added this?

epa095: Can it be because I use py3.6 and not 3.7?

milesgranger: Oh, ya.. that's the backport fo sho...maybe. :man_shrugging:

milesgranger (approved):

PR #882, opened 2020-01-20
Client caching based on revision

### LINKED ISSUE #876: Client: Optimize caching with revisions in mind
We are caching metadata in the client, but we are not using the revision. It probably makes sense to have the cache have key `(machine, revision)`, not just `machine`. And should it be a LRU cache with a huge size? IDK, but that kind of makes sense to me.

### LINKED ISSUE #884: Client: Use revisions wherever possible.

epa095:
This closes #876 
This closes  #884

### DIFF
--- a/gordo/cli/client.py
+++ b/gordo/cli/client.py
@@ -21,7 +21,6 @@ from gordo.client.forwarders import ForwardPredictionsIntoInflux
 
 @click.group("client")
 @click.option("--project", help="The project to target")
-@click.option("--target", help="Single target, instead of all project targets")
 @click.option("--host", help="The host the server is running on", default="localhost")
 @click.option("--port", help="Port the server is running on", default=443)
 @click.option("--scheme", help="tcp/http/https", default="https")
@@ -61,6 +60,13 @@ def client(ctx: click.Context, *args, **kwargs):
 @click.command("predict")
 @click.argument("start", type=IsoFormatDateTime())
 @click.argument("end", type=IsoFormatDateTime())
+@click.option(
+    "--target",
+    help="A list of machines to target. If not provided then target all machines in the"
+    " project",
+    default=[],
+    multiple=True,
+)
 @click.option(
     "--data-provider",
     type=DataProviderParam(),
@@ -106,6 +112,7 @@ def predict(
     ctx: click.Context,
     start: datetime,
     end: datetime,
+    target: typing.List[str],
     data_provider: providers.GordoBaseDataProvider,
     output_dir: str,
     influx_uri: str,
@@ -139,7 +146,7 @@ def predict(
 
     # Fire off getting predictions
     predictions = client.predict(
-        start, end
+        start, end, targets=target,
     )  # type: typing.Iterable[typing.Tuple[str, pd.DataFrame, typing.List[str]]]
 
     # Loop over all error messages for each result and log them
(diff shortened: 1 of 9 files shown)


epa095:
Tested on aurora and it works!

mrdeveloperdude (commented):
Looks good to me

gordo/cli/client.py
```diff
@@ -185,20 +203,27 @@ def metadata(ctx: click.Context, output_file: typing.Optional[typing.IO[str]]):
 
 @click.command("download-model")
 @click.argument("output-dir", type=click.Path(exists=True))
+@click.option(
+    "--target",
+    help="A list of machines to target. If not provided then target all machines in the"
+    " project",
+    default=[],
+    multiple=True,
+)
```
milesgranger, line 212: Rather than repeating these blocks, do you think it's worth it to pass it higher up in the context? Maybe the ergonomics it worth it though?

gordo/client/io.py
```diff
@@ -16,15 +15,56 @@ class HttpUnprocessableEntity(BaseException):
...
+
+    pass
+
+
+class NotFound(BaseException):
+    """
+    Represents a 404
+    """
+
+    pass
```
milesgranger, line 43: Should we be using the werkzeug or requests' exceptions?

gordo/client/io.py
```diff
@@ -16,15 +15,56 @@ class HttpUnprocessableEntity(BaseException):
...
         if resp.status_code == 422:
-            raise HttpUnprocessableEntity()
+            raise HttpUnprocessableEntity(msg)
+        elif resp.status_code == 410:
+            raise ResourceGone(msg)
+        elif resp.status_code == 404:
+            raise NotFound(msg)
         elif 400 <= resp.status_code <= 499:
-            raise BadRequest(msg)
+            raise BadGordoRequest(msg)
```
milesgranger, line 99: Would this be a suitable change?
```suggestion
resp.raise_for_status()
```
[raise_for_status docs](https://2.python-requests.org/en/v2.0-0/user/quickstart/#errors-and-exceptions)

gordo/server/views/base.py
```diff
@@ -240,13 +240,13 @@ def get(self):
 
 class ModelListView(Resource):
     """
-    List the current models capable of being served by the model
+    List the current models capable of being served by the server
```
milesgranger, line 243: :eyes:

gordo/server/views/base.py
```diff
@@ -240,13 +240,13 @@ def get(self):
...
     """
 
-    @api.doc(description="List the metadata for all models capable of being served.")
+    @api.doc(description="List the name of the models capable of being served.")
     def get(self, gordo_project: str):
         available_models = os.listdir(g.collection_dir)
-        return jsonify({"models": available_models})
+        return jsonify({"models": available_models, "revision": g.current_revision})
```
milesgranger: May want to change the key to `current-revision` as there is a post request callback which appends `revision` to be the one that was asked for in the request.

gordo/client/io.py
epa095: I avoided werkzeug because I thought it would be nice to be able to seperate out the server as a extras_require, and then it would be cool not to require werkzeug in the client-part

gordo/cli/client.py
epa095: :man_shrugging:  Yeah maybe, but if felt a bit ugly having to dig it out, and with this there is a more clear mapping between the CLI and the python API

gordo/client/io.py
epa095: Yeah probably nicer, why re-invent the wheel lizm

gordo/server/views/base.py
epa095: damndamndamn goodie point

gordo/client/io.py
epa095: I changed my mind. It caused some lame changes upstream (where instead of just looking at the type of the exception we needed to look inside it to determine what path to take). I am still for it in the long run, just not in this PR.

gordo/server/views/base.py
epa095: I just rely on that one instead of adding myself.

gordo/client/io.py
```diff
@@ -16,15 +15,87 @@ class HttpUnprocessableEntity(BaseException):
...
+    Returns
+    -------
+     Union[dict, bytes]
+
+    Raises
+    ------
+    HttpUnprocessableEntitys
+        In case of a 422 from the server
+    ResourceGone
+     In case of a 410 from the server
```
ryanjdillon: Trivial, but the formatting here is a bit wonked.

ryanjdillon (commented):
Looks nifty :+1: 

One overarching thing as that there is still quite a bit of what we would call a *machine* now being referenced to as a "model". Perhaps this could be a moment to flush that out in the client.

ryanjdillon (approved):
I thought I approved this before, but maybe I was leaving that for Mr. milg.

PR #886, opened 2020-01-23
Replace gevent with gthread

milesgranger:
Using tensorflow>=2.1 has led to gevent has raising
```python
TypeError: cannot create weak reference to 'gevent._local.local' object
```

Replacing this with a more 'native' Async worker `gthread` seems to solve the issue and remove the gevent dependency.


ryanjdillon:
As per notes here #536 

Suggestions to use `uwgsi` with Kubernetes instead of `gunicorn
https://stackoverflow.com/questions/51610189/docker-kubernetes-gunicorn-celery-multiple-workers-vs-replicas/51855285#51855285

Guide on setting it up
https://uwsgi-docs.readthedocs.io/en/latest/Gevent.html

epa095 (approved):
LGTM, lets squash before merging

PR #887, opened 2020-01-23
Don't fail workflow if codecov fails

milesgranger:


epa095 (commented):
LGTM. Increased timeout would be better, but it dont seem like that action has it as a parameter

PR #891, opened 2020-01-27
Feature gates for mlflow and postgres reporters

### LINKED ISSUE #867: Use extras_require to reduce dependencies
readthedocs is failing because we have so many dependencies.

After compartmentalizing the concept of `reporters` we should be able to reduce dependencies for doc building by ignoring any postgres and mlflow deps.

so our doc install config would be `pip install .[docs]`

### LINKED ISSUE #868: Tracking issue: Reporters & feature gates.
Problem
-----------
We are essentially logging/reporting machines resulting from builds in two different ways (mlflow, and server-to-sql script) to two different sources.

- This causes any users of gordo to depend on these.
- Bloats our dependencies, which aside from being 'bad' also causes readthedocs to fail from memory consumption.
- Non-uniform way of implementing the logging/reporting machines when they are built.

Solution
-----------

- [x] Introduce concept of `Reporter` for postgres, which can be included in a Machine's runtime and will be called after the machine is built. (#865)

- [x]  Do the same for ML Flow logging. (#866)

- [ ] Place specific reporter requirements behind feature gates using `extras_require`, this will include only minimal deps (ie. excludes postgres and mlflow reporting dependencies) so _hopefully_ will put us back under memory use. (#867)

milesgranger:
Last issue to close #867 and will close #868


gordo/cli/cli.py
```diff
@@ -38,7 +38,6 @@
     NoSuitableDataProviderError: 70,
     InsufficientDataError: 80,
     InsufficientDataAfterRowFilteringError: 81,
-    MlflowLoggingError: 90,
```
ryanjdillon, line 39: Removing this import makes sense; however, having the mappings all in one place also makes sense. Talked with Erik about it, and his suggestion is to move the reporter exeception(s) to their own file/module adjacent to the reporter modules (i.e. `gordo/reporters/exceptions`, which then can be imported by both the reporter and cli modules.

ryanjdillon (dismissed):
Feature gates and reporters looking sweet :+1:

gordo/cli/cli.py
milesgranger: cc: @epa095

Ok. 

This implies that anyone also implementing a reporter needs to somehow get their exception classes into this file, and in this respect, I see nothing wrong with updating the mapping within the respective reporters; that would be the natural way to go about it in practice.

Atop of that, we'll then essentially be importing and mapping error types to codes for a feature which may not even be installed in the first place?

epa095: > This implies that anyone also implementing a reporter needs to somehow get their exception classes into this file

Yes, and they need to modify `gordo/cli/cli.py` to map it to a exit code. I am also fine with having the mlflow exception in a file by itself, if it is more natural to not have the exceptions in a common file. 

I don't think it is good for the ownership of `EXCEPTION_TO_EXITCODE` to be scattered around the codebase. It is good to be able to go to the mapping and immediately see the exit codes and what errors cause them. Having the map being updated from around the codebase makes it hard to know the complete mapping, and even opens the possibility for crashing exit codes. 

Part of the API of the reporters is what exceptions it raises, but not how the `cli` is supposed to handle that exception, that is for the cli to determine. We actually thought about another alternative earlier, to have the exception classes have a class constant `ERROR_CODE`, which contained the error code, instead of using the mapping. But it was rejected exactly because the exit code is not owned by the exception. 

One compromise I am willing to consider is the following:
If the `cli` gets a error it does not recognise in `EXCEPTION_TO_EXITCODE` it will see if it has a variable/method `recommended_exit_code`, **and** if this is between 100 and 120 it will exit with that code. Then we allow our non-existing future plugins to define their own exit codes, but we restrict them to the "unorganized region" of 100 to 120 where anything can happen.

Dockerfile-Client
```diff
@@ -29,12 +29,12 @@ ENV PATH "${HOME}/.local/bin:${PATH}"
...
 RUN pip install --no-deps -r prereq.txt --no-cache-dir
 
-COPY requirements.txt .
+COPY requirements/requirements.txt .
 RUN pip install -r requirements.txt --no-cache-dir
 
 # Install gordo, packaged from earlier 'python setup.py sdist'
 COPY --from=builder /code/dist/gordo-packed.tar.gz .
-RUN pip install gordo-packed.tar.gz
+RUN pip install gordo-packed.tar.gz[full]
```
epa095, line 37: Note that this does not use pinned versions :-/ So I think we should pip install the `txt` versions of mlflow / postgres. But then it gets quite important that those files are compatible with `requirement.txt`

gordo/cli/cli.py
milesgranger: > Yes, and they need to modify gordo/cli/cli.py to map it to a exit code.

This isn't possible unless they fork the repo though, and we'd expect users to maintain a fork when using Gordo?

> I don't think it is good for the ownership of `EXCEPTION_TO_EXITCODE` to be scattered around the codebase.

I would contend it's not scattered, as in this context, only `reporters` are going to be updating it. Or perhaps I misunderstand.

> Part of the API of the reporters is what exceptions it raises, but not how the cli is supposed to handle that exception, that is for the cli to determine. 

In that sense, the API is broken. If the CLI is to determine how to handle exceptions raised by arbitrary reporters, that won't scale out to other implementations, which goes back to the first point.

With all that said..  :man_shrugging: I think you're more set on this than I am, so I'll yield.

Dockerfile-Client
milesgranger: So I'm having doubts about this whole system of requirements, especially when introducing `extas_require` and now a directory dedicated to holding all these files. I've looked around at various projects, like Flask, and [their setup.py](https://github.com/pallets/flask/blob/master/setup.py) is seemingly reflective of the larger practice of not having such abundant and strict requirements. 

In this case, saying we need to install the hard pinned requirements essentially says that our project is so unstable it can't be trusted to be installed from flexible requirements. I think this would open up the project to becoming more robust.

What do you think?

ryanjdillon: From a science standpoint, I think the pinned deps are a good thing for reproducibility, which isn't a problem Flask has. If they aren't pinned,  I think it may make sense to include the deps in the metadata somehow, but then that would affect the serializer... hmmm.

milesgranger: [sklearn](https://github.com/scikit-learn/scikit-learn/blob/master/setup.py), [pytorch](https://github.com/pytorch/pytorch/blob/master/setup.py), and [mxnet](https://github.com/deepinsight/mxnet/blob/master/python/setup.py) also don't use pinned versions for installs, from the science standpoint as you suggest.

epa095: Flasks setup.py is solving a different problem than our pinned requirement files. There is a difference between listing the requirements needed to install a library/application, vs specifying a deploy environment. That is the point of the pinned requirement files. Think of them more as belonging to the dockerfile than belonging to the setup.py. If we had the dockerfiles in a seperate repository (I am not saying we should do that, but if we did) then I think we would have put the pinned requirement files in there instead of here.

epa095: But miles, we are also not using pinned versions in our install_requires...

setup.py
```diff
@@ -1,12 +1,31 @@
...
+                if r.strip() and not r.startswith("#") and not r.startswith("-")
+            )
+    return req
+
+
+extras_require = {
+    "docs": requirements("docs_requirements.in"),
+    "mlflow": requirements("mlflow_requirements.in"),
+    "postgres": requirements("postgres_requirements.in"),
+    "tests": requirements("requirements.txt", "test_requirements.txt")
```
epa095: :-1:

epa095 (changes requested):

setup.py
epa095: But in generall things make me happy in this PR :smile:  :man_cartwheeling:

.github/workflows/main.yml
```diff
@@ -24,15 +24,15 @@ jobs:
...
           restore-keys: |
             ${{ runner.os }}-${{ matrix.python-version }}-pip-
 
       - name: Install
         run: |
           pip install --upgrade pip
-          pip install -r requirements.txt
-          pip install -r test_requirements.txt
+          pip install -r requirements/requirements.txt
+          pip install -e .[tests]
```
epa095: Do we want to remove the `-e` maybe?

.github/workflows/main.yml
```diff
@@ -24,15 +24,15 @@ jobs:
...
+          key: ${{ runner.os }}-${{ matrix.python-version }}-pip-${{ hashFiles('requirements/*requirements.txt') }}
           restore-keys: |
             ${{ runner.os }}-${{ matrix.python-version }}-pip-
 
       - name: Install
         run: |
           pip install --upgrade pip
-          pip install -r requirements.txt
-          pip install -r test_requirements.txt
+          pip install -r requirements/requirements.txt
```
epa095: Did we want to rename `requirements.txt` to `full_requirements.txt` or something?

epa095: Given that it is in fact not just the output of requirements.in anymore

gordo/cli/cli.py
```diff
@@ -36,7 +39,7 @@
...
+    ReporterException: 90,
```
epa095, line 42: :+1:

epa095 (commented):
LOOOOVE IT :heart_decoration:

.github/workflows/main.yml
milesgranger: Agreed!

epa095 (approved):

PR #892, opened 2020-01-27
Raise better error on missing data

### LINKED ISSUE #890: Return empty data instead of error for `get_data()`
Moved from issue in gordo-infrastructure:

@henrikny 
> .get_data() method throws the error One or more series in the tag list is missing data for the specified period
This is kind of annoying when you're trying to fetch data for 50+ tags and you have no idea which one is the culprit
Suggestion:
Option to return empty data

milesgranger:
Will close #890 

Catch individual `IndexError`s during `join_timeseries` and accumulate a list of the series name's that caused the errors and reflect those back out in the final `InsufficientDataError`

### DIFF
--- a/gordo/machine/dataset/base.py
+++ b/gordo/machine/dataset/base.py
@@ -12,6 +12,10 @@ import numpy as np
 logger = logging.getLogger(__name__)
 
 
+class InsufficientDataError(ValueError):
+    pass
+
+
 class GordoBaseDataset:
 
     _params: Dict[Any, Any] = dict()  # provided by @capture_args on child's __init__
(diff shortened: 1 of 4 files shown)


gordo/machine/dataset/base.py
```diff
@@ -115,6 +119,14 @@ def join_timeseries(
 
         """
         resampled_series = []
+        series_iterable = list(series_iterable)
```
epa095: Carefull! This loads all the series in memory, and they can be big. The rest of the stuff is intended to be lazy (using generator).

milesgranger: ya ya.. trying to think of a nicer way around this. :(

ryanjdillon:
Very nice :)

gordo/machine/dataset/base.py
```diff
@@ -115,77 +119,99 @@ def join_timeseries(
...
         return new_series.dropna()
+
+    @staticmethod
+    def _resample(
+        series: pd.Series,
+        resampling_startpoint: datetime,
+        resampling_endpoint: datetime,
+        resolution: str,
+        aggregation_methods: Union[str, List[str], Callable] = "mean",
+    ):
```
ryanjdillon, line 156: Are there tests that cover this bad boy already?

ryanjdillon (approved):

gordo/machine/dataset/base.py
milesgranger: A couple in here: https://github.com/equinor/gordo/blob/master/tests/gordo/machine/dataset/test_dataset.py

PR #900, opened 2020-01-29
dataframe_from_dict: parse for int and sort

### LINKED ISSUE #899: Update documentation and maybe tests when from_dict()ing a json response
We had an issue with Tilstandomatic where the dictionary returned from the /anomaly/predict was correct, but when from_dict() was used, the order of the rows appeared wrong. And they were wrong.

The reason was that the row number was used as the index, but it is represented as a string, not numeric.

Maybe we should have a test for this. Normally we use datestamps, so we don't notice this behaviour, a bit by luck.

milesgranger:
Will close #899 

Attempt to sort the dataframe index as integers as well, followed by sorting.

### DIFF
--- a/gordo/server/utils.py
+++ b/gordo/server/utils.py
@@ -179,8 +179,10 @@ def dataframe_from_dict(data: dict) -> pd.DataFrame:
     try:
         df.index = df.index.map(dateutil.parser.isoparse)  # type: ignore
     except (TypeError, ValueError):
-        logger.debug("Could not parse index to pandas.DatetimeIndex")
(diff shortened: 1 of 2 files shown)


flikka (approved):
There are probably arguments for doing somethings differently in this "domain", supporting other indexes etc. but for the given situation I think this is a sensible improvement!

flikka (commented):
One potential improvement could be to actually have a pandas dataframe with a datetime index and an integer one, do to_dict and json.dumps, and then take it back again - producing these strings in the index.

PR #902, opened 2020-01-30
Add 'dataset' key for build metadata

### LINKED ISSUE #779: Add more metadata about the actual dataset
We have a metadata for dataset (eg. time taken to fetch data), another interesting thing:
* What was the _actual_ start and end timestamp (based on presence of data)?
* For each tag, how many datapoints were there, originally?

milesgranger:
Part of #779 
This sets up the ability to add more explicit `dataset` build metadata as its own key withing `BuildMetadata`. 

In a seperate PR(?), should probably change the metadata that is returned from `TimeSeriesDataset.get_metadata` as it basically gives what `.to_dict()` gives; as well as additional stuff about the dataset itself which happened during the build.

### DIFF
--- a/gordo/builder/build_model.py
+++ b/gordo/builder/build_model.py
@@ -31,6 +31,7 @@ from gordo.machine import Machine
 from gordo.machine.metadata import (
     BuildMetadata,
     ModelBuildMetadata,
+    DatasetBuildMetadata,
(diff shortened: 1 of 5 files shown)


ryanjdillon (approved):
LGTM :+1:

PR #903, opened 2020-01-30
Add assets to dataset resolvers

ryanjdillon:
Adds resolving of assets in tilstandomatic CSV files that are not currently supported. 

As per the issue on `gordo-infrastructure`:
https://github.com/equinor/gordo-infrastructure/issues/508


flikka (changes requested):
Nice! Would it be an idea to make a status list of the assets (files/tag-groups) that we are missing, ie. don't find in the datalake? Using the tilstandomatic csv's, as you probably have done now. You can omit the --predictive flag on ioc_gordo to get a gordo model for every tilstandomatic model with IMS tags.

Also -> I think Oseberg is in the datalake, so I guess you can add that. But it is easier to get an overview if we have a list of _all_ assets that tilstandomatic has, and our status on them. Maybe log it on the issue in gordo-infrastructure that you refer to?

gordo/machine/dataset/data_provider/ncs_reader.py
```diff
@@ -20,11 +20,15 @@ class NcsReader(GordoBaseDataProvider):
...
         "1101-sfb": "/raw/corporate/IMS Statfjord/sensordata/1101-SFB",
         "1110-gfa": "/raw/corporate/Aspen MS - IP21 Gullfaks A/sensordata/1110-GFA",
+        "1111-gfb": "/raw/corporate/Aspen MS - IP21 Gullfaks B/sensordata/1111-GFB",
+        "1112-gfc": "/raw/corporate/Aspen MS - IP21 Gullfaks C/sensordata/1112-GFC",
         "1125-kvb": "/raw/corporate/PI System Operation Norway/sensordata/1125-KVB",
         "1130-troa": "/raw/corporate/Aspen MS - IP21 Troll A/sensordata/1130-TROA",
         "1138-val": "/raw/corporate/PI System Operation Norway/sensordata/1138-VAL",
         "1163-gdr": "/raw/corporate/PI System Manager Sleipner/sensordata/1163-GDR",
         "1180-nor": "/raw/corporate/PI System Operation North/sensordata/1180-NOR",
+        "1175-kri": "/raw/corporate/Aspen MS - IP21 Kristin/sensordata/1175-KRI",
```
EinarJohnsen: This path seems to be incorrect. 
Also, would like if this and the next line was changed so it is in sorted order on the number.

ryanjdillon:
@flikka The assets I have gone over so far are those that had no YAMLs autogenerated. 

I sent some status files over email to be forwarded to IOC and Omnia, do you mean a status file we keep in the repo?

Oseberg A and D are in the Datalake, while only Oseberg B tags are in the tilstandomatic CSV files. I have a list of the Oseberg tags in the CSV file I sent over email.

gordo/machine/dataset/data_provider/ncs_reader.py
ryanjdillon: Nice catch. I was sure that you were wrong and double checked and... dangit! :)

EinarJohnsen: Haha, you can't trick a trickster :smile: 
Otherwise it looks good :tada:

ryanjdillon:
Talked with Einar about a problem I observed with regards to the Odeberg A, B D (sub) assets. For each, all tags are prefixed with OSF, so when Odeberge B is added to the datalake, we will need to take the Gudrun approach and inject some other unique prefix for each sub-asset to match on.

flikka:
> Oseberg A and D are in the Datalake, while only Oseberg B tags are in the tilstandomatic CSV files. I have a list of the Oseberg tags in the CSV file I sent over email.

Hmm. Let's talk.

I found
Oseberg C: OSC-FT  -13-010A
Oseberg South: OSS-PI  -13-8853
Oseberg East: OSE-PZI -13-0793
Oseberg Feltsenter: OSF-13-TI  -0123B

but not oseberg B tags :-)

flikka (approved):
I think we can just add these, and keep 'em coming

ryanjdillon:
This can now be done auto-magically via gordo-infrastructure PR [517](https://github.com/equinor/gordo-infrastructure/pull/517).

ryanjdillon:
Though Kristian and I talked about this, just in case anybody digs this up...

The problem I described graphically:
```
raw
└── corporate
    └── AspenMS-IP21OsebergFieldCentre
        └── sensordata
            ├── 1772-OSA
            │   └── OSF-23-FI-201
            ├── 1774-OSD
            │   └── OSF-25-FIC-0096
            └── -1-OSF-unknown
                └── OSF-13-FQI-0916CY
```

PR #904, opened 2020-01-30
Upgrade to pandas 1.0.0

EinarJohnsen:
Opens the possibility to use pandas own query and eval on dataframes on column names with multiple spaces.


milesgranger (approved):
Looks good enough to me, if it works on the test project. @epa095 ?

PR #905, opened 2020-01-31
Add Mlflow report example notebook

### LINKED ISSUE #861: Add notebook that demos Mlflow querying
This was never added to the repo. Need to cleanup a little and remove some sercrets before adding.

ryanjdillon:
Added the example notebook for querying with the MlFlow client with updates for using the reporters.

closes #861

### DIFF
--- a/examples/Gordo-Reporters-MlFlow.ipynb
+++ b/examples/Gordo-Reporters-MlFlow.ipynb
@@ -1,0 +1,383 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Using the Gordo Mlflow reporter with AzureML\n",
+    "\n",
+    "## Building on a cluster\n",
+    "When a gordo workflow is generated from a YAML config using `kubectl apply -f config.yml`, the model is built by the model builder pod. If a remote logging \"reporter\" was configured in the `config.yml`, then at the end of the model building step the metadata will be logged with the specified reporter. \n",
+    "\n",
+    "**Note**\n",
+    "When using the MLflow reporter, the cluster running the workflow must have the AzureML workspace credentials set to the environment variable `AZUREML_WORKSPACE_STR` as well as the `DL_SERVICE_AUTH_STR`.\n",
+    "\n",
+    "The cluster should use the workspace credentials associated with the deployment stage associated with that cluster, e.g. \"production\", \"staging\", \"testing\", etc.\n",
+    "\n",
+    "While reporters can be defined in the globals runtime when using the workflow generator, they must be defined by machine when building locally."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "\n",
+    "from azureml.core.workspace import Workspace\n",
+    "from azureml.core.authentication import InteractiveLoginAuthentication\n",
+    "import mlflow\n",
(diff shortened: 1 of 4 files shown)


ryanjdillon:
@flikka Do you have any notes you think we should add for the AzureML sdk?

examples/Gordo-Reporters-MlFlow.ipynb
```diff
@@ -0,0 +1,353 @@
...
+    "        name: the-model-name\n",
+    "        # Can be configured individually\n",
+    "        reporter: \"gordo.reporter.mlflow.MlFlowReporter\"\n",
+    "globals:\n",
+    "      runtime:\n",
+    "        builder:\n",
+    "          remote_logging:\n",
+    "            enable: True\n",
+    "      # Can be configured for all machines\n",
+    "      reporter: \"gordo.reporter.mlflow.MlFlowReporter\"\n",
```
milesgranger: A reporter goes inside `runtime.reporters`

examples/Gordo-Reporters-MlFlow.ipynb
```diff
@@ -0,0 +1,353 @@
...
+    "            - Tag1\n",
+    "            - Tag2\n",
+    "          target_tag_list:\n",
+    "            - Tag3\n",
+    "            - Tag4\n",
+    "          train_end_date: '2019-03-01T00:00:00+00:00'\n",
+    "          train_start_date: '2019-01-01T00:00:00+00:00'\n",
+    "          asset: asgb\n",
+    "          data_provider:\n",
+    "            type: RandomDataProvider\n",
```
flikka: I find it a bit confusing this config. It has "asset:asgb" but tag names that are just examples. And it uses RandomDataProvider.

What about making a config that would work, if you only inserted tags we recognize? Meaning, get rid of the data_provider and the asset:, and fix the reporter according to @milesgranger feedback. And test the config, that it works, also with the mlflow reporter?

flikka (changes requested):

examples/Gordo-Reporters-MlFlow.ipynb
ryanjdillon: Ok, the notebook is now updated with an example from the test project, including setting interactive login with the data provider. 

@milesgranger I also updated how the logging methods pull the secrets to make it a tad simpler, just in case you have any thoughts on that.

flikka:
@ryanjdillon closed it?

examples/Gordo-Reporters-MlFlow.ipynb
```diff
@@ -0,0 +1,379 @@
...
+    "kind: Gordo\n",
+    "metadata:\n",
+    "  name: test-project\n",
+    "spec:\n",
+    "  deploy-version: 0.50.0\n",
+    "  config:\n",
+    "    machines:\n",
+    "      - dataset:\n",
+    "          tags:\n",
+    "            - TRA-35TT8566.PV\n",
```
flikka: No real tags I'm afraid. The tag names are not approved for public audience. But we kan keep the prefix, and just alter some numbers/letters perhaps?

flikka (changes requested):
Looks good, a part from the real tags which should be "anonymized" (change the numbers/letters a bit?)

examples/Gordo-Reporters-MlFlow.ipynb
ryanjdillon: Ahh... I misunderstood. Got it :+1:

Issue #522, opened 2019-10-09
Cleanup filter_rows.py after upgrading to pandas v1


EinarJohnsen:
Cleanup code in filter_rows after upgrade to pandas 0.25 as it should support quoting column names with backticks to refer to names with spaces.

milesgranger:
Depends on #386

epa095:
Link to the pandas issue?

EinarJohnsen:
https://github.com/pandas-dev/pandas/issues/28893
There is a PR on it, fixing other stuff and also our issue.
https://github.com/pandas-dev/pandas/pull/28215

hwalinga:
As the implementer of that PR I would like to remind you that you should not rely on differentiating column names by spaces. In fact, this will raise an ambiguity warning in the new PR, as the information of multiple spaces will get lost when the query is parsed as Python. 

I.e. 
```
df = pd.DataFrame({"A A": [1, 2], "A  A": [3, 4]})
df.query("`A A` > 1")
```

This will raise an error. As on the parser end, the only information that is received is ``('`', 'A', 'A', '`', '>', '1')``, even with two spaces between the two A's, thus one could be referring. to 
```
"A A"
``` 
or 
```
"A  A"
``` 
This will therefore raise an error. 

Implementing such a feature would require a different approach, and in general I would highly advise against using multiple spaces to differentiate your column names. 

For now, what will improve with the new PR, is that, if you only would have a `"A  A"` in your column names, you can still use it in a query. The error will only be raised if there is an unresolvable ambiguity in your column names.

hwalinga:
**update:** Inspired by your hope of multiple spaces being supported in the .query function, I implemented a different approach that catches does as well. So the PR will support that as well. 

(You still cannot use the hashtag and unpaired quotes.)

EinarJohnsen:
Thank you @hwalinga 
I noticed you made a commit for that. The point of it was more to allow having multiple spaces, as it is not always the case where the columns have only one space. The example given was more of a "special" case, since we do not often use spaces to differentiate column names, but we need to support names possibly containing multiple spaces. 
And as the support for names with was added, it seemed like a somewhat unnecessary limitation to only have support for one space :smile: 
Thank you for taking your time to inform us.

EinarJohnsen:
In the next pandas release (0.26), this should be possible. I have built pandas from source with version 0.26.0.xx, and it works as we expect it to.There is a couple of tests that has to be changed slightly, but otherwise it looks good.

PR #908, opened 2020-02-03
Cleanup filter_rows.py

EinarJohnsen:
Remove custom ast implementation to allow multiples spaces in column names, as it is supported in pandas 1.0.0. Cloeses #522

### DIFF
--- a/gordo/machine/dataset/filter_rows.py
+++ b/gordo/machine/dataset/filter_rows.py
@@ -1,46 +1,8 @@
-import ast
 import logging
 import pandas as pd
 import numpy as np
 
 logger = logging.getLogger(__name__)
-
-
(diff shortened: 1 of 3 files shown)


milesgranger (approved):
LGTM. :ok_man:

EinarJohnsen:
Just to have mentioned it, this removes the possibility to have quotes inside quotes. 
I don't know why we chose to have it before, but if we merge this it is no longer supported: 
` " 'A' > 2"`. One should either remove the single quotes or use backticks. If the column name contains spaces one must use backticks e.g. ``` "`Col A` > 2"  ```. 
We could do some "clever" string replacement to still support quote inside quote, but I am not sure what advantages it would give. Any thoughts @epa095 @milesgranger

PR #911, opened 2020-02-03
Server: support reporting expected models

### LINKED ISSUE #883: Server: Also return expecting models for current revision
Today the server returns the models it has and which are ready to serve, but it would be useful to get a complete list of models.

milesgranger:
Will close #883

### DIFF
--- a/gordo/server/views/base.py
+++ b/gordo/server/views/base.py
@@ -7,7 +7,6 @@ import traceback
 import timeit
 import typing
 
-import yaml
 import pandas as pd
 from flask import Blueprint, current_app, g, send_file, make_response, jsonify, request
 from flask_restplus import Resource, fields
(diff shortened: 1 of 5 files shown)


gordo/server/views/base.py
```diff
@@ -268,7 +268,14 @@ def get(self, gordo_project: str):
         )
 
 
+class ExpectedModels(Resource):
+    @api.doc(description="Models that the server expects to be able to serve.")
+    def get(self, gordo_project: str):
+        return jsonify({"expected-models": current_app.config["EXPECTED_MODELS"]})
```
epa095, line 274: It makes sense to me that this also reports the current version the server is running on.

epa095 (commented):
I imagined that this would be another key in the return value of ModelListView, so it would give the expected models in addition to the actual models if revision is the latest. I dont know what I think about the pros and cons of having it as a seperate endpoint. Maybe ok?

epa095 (dismissed):
LGTM!

gordo/server/views/base.py
```diff
@@ -243,10 +243,19 @@ class ModelListView(Resource):
...
+        "served, and those expected to be able to serve."
+    )
     def get(self, gordo_project: str):
         available_models = os.listdir(g.collection_dir)
-        return jsonify({"models": available_models})
+        return jsonify(
+            {
+                "models": available_models,
+                "server-version": __version__,
+                "expected-models": current_app.config["EXPECTED_MODELS"],
```
epa095: so yeah, this was a bit tricky... `current_app.config["EXPECTED_MODELS"]` is not the expected models for a old revision, only for the current. So what do return for old revision? Remove the key? Should it just be the same as models? Should it be empty? Uffda, IDK. 


Also, I noobed when I in my previous comment said "current version", I meant the revision for the request. I.e. if none is provided then whatever is the latest revision the server is serving, or the passed revision.

milesgranger: Could maybe change the key to something like `current-revision-expected-models`?

epa095:
I change my mind! I am ok with having it as a seperate endpoint!

epa095 (approved):
lovely

PR #912, opened 2020-02-04
Update TimeSeriesDataset metadata

### LINKED ISSUE #779: Add more metadata about the actual dataset

milesgranger:
Will close #779 

Remove current metadata, as it is basically a duplicate
of what `.to_dict()` does in the resulting machine.dataset
storage.

Updates the underlying `.join_timeseries` to record
information about each series which is loaded along with
the final joining of those series.


ryanjdillon (approved):
LGTM :+1:

PR #917, opened 2020-02-05
NotFound on invalid model metadata request

### LINKED ISSUE #915: Server: return a 4xx code on metadata request for non-existing machines.
Requesting the metadata for a non-existing machine gives a 200 result with the following metadata:
```json
{"gordo-server-version": "0.48.3", 
"metadata": {}, 
"env": {"MODEL_COLLECTION_DIR": "/gordo/models/ep-tp-483/1580813025723"}, 
"revision": "1580813025723"}
```

It should probably return a 4xx code to indicate that it is a non-existing machine. Example; 

`curl 127.0.0.1:8000/gordo/v0/ep-tp-483/non-exdsdf/metadata`

milesgranger:
Will close #915

### DIFF
--- a/gordo/serializer/serializer.py
+++ b/gordo/serializer/serializer.py
@@ -6,7 +6,6 @@ import os
 import re
 import pickle
 
-from os import path
 from typing import Union, Any  # pragma: no flakes
 
 from sklearn.pipeline import Pipeline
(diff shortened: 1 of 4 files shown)


epa095 (approved):
LGTM!

PR #918, opened 2020-02-05
Return 404 req to download non-existing model

### LINKED ISSUE #916: Server: return a 4xx code download request for non-existing machines.
Today there is a FileNotFoundException raised in the server, which is not catched, and the result is a 500 SERVER ERROR return :-/

milesgranger:
Will close #916


epa095 (approved):
LGTM!

PR #919, opened 2020-02-05
Include features' thresholds per fold in metadata

milesgranger:
We currently log into metadata the calculated thresholds per feature and the aggregate. It has been requested to additionally log the existing intermediate calculations of the "feature thresholds per fold" as well.

Example:

```python
>>> machine.metadata.build_metadata.model.model_meta['feature-thresholds-per-fold']
{'tag-0': {'fold-0': 0.5027252655852338,
           'fold-1': 0.4664650358022078,
           'fold-2': 0.5429959691823347},
 'tag-1': {'fold-0': 0.562261042573828,
           'fold-1': 0.6089732143556913,
           'fold-2': 0.5880519662881121},
 'tag-2': {'fold-0': 0.5893215013091214,
           'fold-1': 0.47895224187839747,
           'fold-2': 0.22596768265410738},
 'tag-3': {'fold-0': 0.4760469371831675,
           'fold-1': 0.6181816933881628,
           'fold-2': 0.5593841253053533}}

```

### DIFF
--- a/gordo/machine/model/anomaly/diff.py
+++ b/gordo/machine/model/anomaly/diff.py
@@ -68,6 +68,10 @@ class DiffBasedAnomalyDetector(AnomalyDetectorBase):
             metadata["feature-thresholds"] = self.feature_thresholds_.tolist()
         if hasattr(self, "aggregate_threshold_"):
             metadata["aggregate-threshold"] = self.aggregate_threshold_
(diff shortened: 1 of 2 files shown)


flikka (approved):
Looks good, just double check with @alexisc67 that it is the _individual tag_ threshold that they want per fold, and not the _aggregated threshold_ per fold.

PR #928, opened 2020-02-07
Pin werkzeug==0.16.0 in req.in for flask-restplus

### LINKED ISSUE #927: Non-pinned installs/Docs fails from rest-pluss/Werkzeug
It fails to build docs because it tries to install non-pinned dependencies, and suddenly Werkzeug 1.0 was released, and flask-restpluss is broken with that, as per https://github.com/noirbizarre/flask-restplus/issues/777

We can:
1. Wait until a fixed version of flask-restpluss is released, or
1. Pin Werkzeug<=0.16

Also, we should probably rely on pinned dependencies for the release pipeline, and rather maybe add a test in PR's that it works to install with latest pypy non-pinned reqs.

epa095:
This closes #927 .

### DIFF
--- a/requirements/full_requirements.txt
+++ b/requirements/full_requirements.txt
@@ -27,7 +27,7 @@ cachetools==4.0.0
 catboost==0.20.2
 cchardet==2.1.5
 certifi==2019.11.28       # via msrest, requests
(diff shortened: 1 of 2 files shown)


EinarJohnsen (approved):
LGTM

PR #929, opened 2020-02-07
Total anomaly change

henrikny:
This PR will apply the following changes:

- remove unnecessary function for the calculation for the feature thresholds (easier to read)
- new calculation of total anomaly (scaled and unscaled analogue):
from 

`data["total-anomaly-scaled"] = np.linalg.norm(data["tag-anomaly-scaled"], axis=1)`
to
`data["total-anomaly-scaled"] = np.square(data["tag-anomaly-scaled"]).mean(axis=1)`

- the total model threshold validation is simplified

### DIFF
--- a/tests/gordo/machine/model/anomaly/test_anomaly_detectors.py
+++ b/tests/gordo/machine/model/anomaly/test_anomaly_detectors.py
@@ -208,28 +208,6 @@ def test_diff_detector_cross_validate(return_estimator: bool):
 
     assert cv_results_da.keys() == cv_results_sk.keys()
 
-
-# simulate LSTM outptu shorter than input
-@pytest.mark.parametrize("y_pred_shape", ((100, 2), (100, 2)))
-@pytest.mark.parametrize("y_true_shape", ((100, 2),))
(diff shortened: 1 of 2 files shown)


epa095:
The build-doc tests fail because of https://github.com/equinor/gordo/issues/927 . When that is fixed and merged you need to rebase on master, and then that test should pass. 

Then there is the `machine` test, which you need to change according to your changes. The easiest is probably for you to run the tests locally. `pip install -r requirements/test_requirements.txt` and then `python setup.py testmachine` to run the test-suite.

tests/gordo/machine/model/anomaly/test_anomaly_detectors.py
```diff
@@ -209,27 +209,6 @@ def test_diff_detector_cross_validate(return_estimator: bool):
     assert cv_results_da.keys() == cv_results_sk.keys()
 
 
-# simulate LSTM outptu shorter than input
```
flikka: Why is this removed?

epa095 (approved):
LGTM :+1:

PR #934, opened 2020-02-17
Add batching to handle errors with log_batch calls

ryanjdillon:
# Problem :doughnut:  
An error was encountered when, what turned out to be, `MlflowClient.log_batch` request payloads were too large.

# Solution :milk_glass: 
Buffer these to the least number of permissibly sized requests.

### DIFF
--- a/gordo/reporters/mlflow.py
+++ b/gordo/reporters/mlflow.py
@@ -4,7 +4,7 @@ import json
 import logging
 import os
 import tempfile
-from typing import List
+from typing import Dict, List, Union, Tuple
 from uuid import uuid4
 
 from azureml.core import Workspace
@@ -185,7 +185,7 @@ def epoch_now() -> int:
     return _datetime_to_ms_since_epoch(datetime.now(tz=UTC))
 
 
-def get_batch_kwargs(machine: Machine) -> dict:
+def get_machine_log_items(machine: Machine) -> Tuple[List[Metric], List[Param]]:
     """
     Create flat lists of MLflow logging entities from multilevel dictionary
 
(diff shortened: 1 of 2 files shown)


gordo/reporters/mlflow.py
```diff
@@ -185,7 +185,66 @@ def epoch_now() -> int:
     return _datetime_to_ms_since_epoch(datetime.now(tz=UTC))
 
 
-def get_batch_kwargs(machine: Machine) -> dict:
+def split_log_item_batches(
+    metrics: List[Metric],
+    params: List[Param],
+    n_max_metrics: int = 200,
```
epa095: My parranoia sais that we should set this to 199, and n_max_params to 99 :-p

epa095: Are we not sending / caring about tags?

gordo/reporters/mlflow.py
```diff
@@ -185,7 +185,66 @@ def epoch_now() -> int:
...
+    Returns
+    -------
+    log_batches: List[Dict[str, Union[Metric, Param]]]
+        List of MlflowClinet.log_batch keyworkd arguments, split to quatnitites
+        that respect limits present for MLFlow and AzureML.
+    """
+
+    def _calc_n_batches(n: int, n_max: int):
+        """
+        Calculate the number of batches required to log items with batches of n_max
```
epa095: ... to log *n* items ...

epa095 (commented):
LGTM, but have you tested it?

gordo/reporters/mlflow.py
ryanjdillon: I manual-integration tested it with those values and things worked, but it couldn't hurt.

Due to the AzureML limits, the tag limits wouldn't come into affect here, as we are only using one tag per run, and the AzureML each limits (I discovered through manual testing) are independent of each other, as opposed to the MLFlow total sum limit that we won't reach due to the Azure limits.

ryanjdillon: Oh, and I there is a unit test to check the batching logic.

ryanjdillon:
Running the `ioc-1901` configuration that previously failed now passes with these changes and logs are successfully generated on AzureML.

epa095 (approved):
LGTM

PR #936, opened 2020-02-17
 Builder: Use un-cached runtime properties

### LINKED ISSUE #935: Reporter information appears to be cached and not overwritten
Example: 
1. You built a machine with the `remote_logging: True`.
2. Then, you change to `remote_logging: False`
If the cache gets a hit (ie. you havent changed other things in the machine), the `remote_logging:` change is "ignored", since the reporters patched into the machine when this was set to `True` were cached with that machine. 

This would probably not be a big issue, BUT, we had a situation where the MLflow logging failed. But the machine was built, and cached with the MLflowReporter, so changing the remote_logging to False didn't stop the (failing) attempt to do remote logging, as the machine with the reporter attached was retrieved from the cache.

Solution: Either include the reporter in the cache key, or "patch" the machine with the information

epa095:
This closes #935

### DIFF
--- a/gordo/builder/build_model.py
+++ b/gordo/builder/build_model.py
@@ -133,6 +133,9 @@ class ModelBuilder:
                 metadata["metadata"][
                     "user_defined"
                 ] = self.machine.metadata.user_defined
+
+                metadata["runtime"] = self.machine.runtime
+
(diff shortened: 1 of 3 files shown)


milesgranger (approved):
Nailed it!

PR #942, opened 2020-02-19
Fix model list view on empty project

milesgranger:
Closes #940

PR #943, opened 2020-02-19
Store models under dedicated folder on disk

epa095:
Before they were stored in /gordo/models/{{project_name}}/{{project_revision}},
now they are stored in
/gordo/models/{{project_name}}/models/{{project_revision}}. The problem was that
the /revision endpoint ended up listing out model_register as well.


epa095:
I cherry-picked this on top of #942 , and they both work!

milesgranger (approved):

PR #947, opened 2020-02-20
Add checks in ncs_reader to handle missing files

### LINKED ISSUE #870: ncs_reader: Handle empty/missing files
There is a bit of an inconsistency in how gordo handles missing data around newyear relative to the rest of the year. The tagfiles in the datalake are given as one file per year. And if we try to read a file and it is not there then we will raise an exception. That means that if you try to read data from e.g. 2019-10-01 -> 2020-01-01, and there is no data in 2020 gordo will crash with filenotfound even though it is only missing one day. But if you ask for e.g. 2019-01-01 -> 2019-03-01 and it is missing data in march, gordo will not complain but rather forward fill the data.

I think it makes sense to catch the filenotfound error and ignore it, and rather let the [InsufficientDataError](https://github.com/equinor/gordo/blob/ccd878f9d55c044975034415aec49ed6f37a2db5/gordo/machine/dataset/datasets.py#L206) be responsible.

EinarJohnsen:
Verifiy that the tag exists, otherwise it crashes with `FileNotFound`. Also ignores missing files instead of crashing, e.g. the years 2018-2020 is specified, but 2019 is missing. Then the 2019 will be discarded (a message is logged) and the generated dataframe will have data from 2018 and 2020. Closes #870

### DIFF
--- a/gordo/machine/dataset/data_provider/ncs_reader.py
+++ b/gordo/machine/dataset/data_provider/ncs_reader.py
@@ -123,6 +123,22 @@ class NcsReader(GordoBaseDataProvider):
                 yield filtered
 
     @staticmethod
+    def _verify_tag_path_exist(
+        adls_file_system_client: core.AzureDLFileSystem, path: str
+    ):
+        """
+        Verify that the tag path exists, if not the `adls_file_system_client.info` will raise a FileNotFound error.
+
+        Parameters
+        ----------
+        adls_file_system_client: core.AzureDLFileSystem
+            the AzureDLFileSystem client to use
+        path : str
+            Path of tag to be checked if exists.
+        """
+        adls_file_system_client.info(f"{path}")
+
+    @staticmethod
     def read_tag_files(
         adls_file_system_client: core.AzureDLFileSystem,
         tag: SensorTag,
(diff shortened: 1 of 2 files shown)


flikka:
Nice and clean. I have one question though, from a "usability" point of view:

If we, like you mention in the comment, have data from 2018 and 2020, but not 2019 - I wonder if we should rather raise some error then? I mean, the realistic scenario (afik) is that we miss years in the tail or head (like around new-year, just started up in january, or have rare data, only in december, but not yet january)? 

Maybe it is risky to silently allow a whole missing year "in-between"? Not superstrong feelings, just a though.

epa095:
> Nice and clean. I have one question though, from a "usability" point of view:
> 
> If we, like you mention in the comment, have data from 2018 and 2020, but not 2019 - I wonder if we should rather raise some error then? I mean, the realistic scenario (afik) is that we miss years in the tail or head (like around new-year, just started up in january, or have rare data, only in december, but not yet january)?
> 
> Maybe it is risky to silently allow a whole missing year "in-between"? Not superstrong feelings, just a though.

Yeah, but maybe this should be somewhere else than in the ncs reader, like in the dataset. Then it can be a bit more general than "error if 1 year is missing" and rather be "error if more than x months are missing." it also allows for catching the case where data is missing between July 2018 and August 2019.

flikka (approved):
Looks good, and as @epa095 points out,  a more sophisticated "check for big wholes in data" logic can be somewhere else (and in a separate issue/PR)

PR #949, opened 2020-02-24
Keep pods if log_level is set to `debug`

### LINKED ISSUE #933: Workflow: Dont delete pods if we are in debug mode
Here we need some logic!
https://github.com/equinor/gordo/blob/a21f254dcc616f387be1ae51d0082d8d5449f1a8/gordo/workflow/workflow_generator/resources/argo-workflow.yml.template#L1110

EinarJohnsen:
Keep the pods in the cluster if the `log_level` is set to `debug`, otherwise they are deleted. Closes #933


epa095 (approved):
LOOOOvely! 
veldig bra

PR #954, opened 2020-02-25
Modify data InfluxDB schema written in Client forwarder.

### LINKED ISSUE #878: Optimize influx schema for quering
The influx schema consists of several measurements, e.g. model-input, model-output, resampled. The model-input/output schema has one `field` per tag, and then the machine name as a `tag`. This has the consequence that when querying for the output for a given machine you get a big matrix back with data for every possible sensor, not only the sensors used by that machine. This matrix has mostly `none` values (for all the sensors which are not part of the particular machine), but still becomes quite huge when running on projects with many sensors, and causes grafana to be quite slow and memory-hungry. 

Some alternatives: 
1. Add the sensor as a tag in addition to the machine, and change the field to e.g. `value`. 
1. Somehow encode somewhere in influx which sensors belong to a machine. Then the `Sensor` variable in grafana can be refactored to only contain these sensors, and then the query will be `select mean($Sensors)`, which maybe (maybe not) makes influx return a more narrow matrix?
1. Other?

Should we also ditch `resampled`? It made sense before we sendt up `model-input`. For the pure auto-encoder case it would be more effecient to ditch `model-input`, but for the general X-Y case there is no other way (that I can see) to know what the input-sensors are (short of fetching them from postgress, which we could).

ryanjdillon:
In order to optimize InfluxDB queries (where sparse matrices of all known tags to the db were returned), restructure the schema so a query will only return measurements with tags associated with a machine.

So it now goes from this:

```
                                    machine    0   1   2   3   4
2019-01-01 00:00:00+00:00  some-target-name    4   5   6   7   8
```

To this:
```
                                    machine sensor_name sensor_value
2019-01-01 00:00:00+00:00  some-target-name           0            4
2019-01-01 00:00:00+00:00  some-target-name           1            5
2019-01-01 00:00:00+00:00  some-target-name           2            6
2019-01-01 00:00:00+00:00  some-target-name           3            7
2019-01-01 00:00:00+00:00  some-target-name           4            8
```

Where `sensor_name` is an InfluxDb "tag" (i.e. is indexed) and `sensor_value` is an InfluxDB "field" (i.e. not indexed).

Checklist
- [x] Modify code to change schema
- [x] Update the Grafana queries

closes #878

### DIFF
--- a/gordo/client/forwarders.py
+++ b/gordo/client/forwarders.py
@@ -4,8 +4,7 @@ import abc
 import itertools
 import logging
 import time
-import typing
-from typing import Optional
+from typing import Any, Dict, Optional
 
 import pandas as pd
 import numpy as np
@@ -135,6 +134,9 @@ class ForwardPredictionsIntoInflux(PredictionForwarder):
         Takes a multi-layed column dataframe and write points to influx where
         each top level name is treated as the measurement name.
 
+        How the data is written via InfluxDB DataFrameClient in this method
+        determines the schema of the database.
+
         Parameters
         ----------
         predictions: pd.DataFrame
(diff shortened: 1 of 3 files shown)


hjarraya (approved):
LGTM

epa095 (approved):
LGTM!

PR #963, opened 2020-03-09
Fix parquet file reader.

hjarraya:
Previous parquet reader was taking a long time to parse date-time columns.

### DIFF
--- a/gordo/machine/dataset/data_provider/file_type.py
+++ b/gordo/machine/dataset/data_provider/file_type.py
@@ -1,5 +1,4 @@
 import pandas as pd
-import pyarrow.parquet as pa
 import numpy as np
 
 from abc import ABCMeta, abstractmethod
@@ -100,10 +99,8 @@ class ParquetFileType(FileType):
     def read_df(self, f: BinaryIO) -> pd.DataFrame:
         columns = self.time_series_columns.columns
         datetime_column = self.time_series_columns.datetime_column
-        pf = pa.ParquetFile(f)
-        table = pf.read(columns)
-        df = table.to_pandas()
-        df[datetime_column] = df[datetime_column].apply(
-            lambda v: pd.to_datetime(v, utc=True)
+        df = pd.read_parquet(f, engine="pyarrow", columns=columns).set_index(
+            datetime_column
         )
-        return df.set_index(datetime_column)
+        df.index = pd.to_datetime(df.index, utc=True)
+        return df
--- a/gordo/machine/dataset/data_provider/ncs_reader.py
+++ b/gordo/machine/dataset/data_provider/ncs_reader.py
@@ -3,6 +3,7 @@ import logging
 from collections import OrderedDict
 from concurrent.futures import ThreadPoolExecutor
 from datetime import datetime
+import timeit
 from typing import Iterable, List, Optional
 from urllib.parse import quote
 
@@ -21,7 +22,6 @@ from gordo.machine.dataset.sensor_tag import SensorTag
 from gordo.util import capture_args
 
 logger = logging.getLogger(__name__)
-
 
 time_series_columns = TimeSeriesColumns("Time", "Value", "Status")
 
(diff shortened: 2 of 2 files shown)


gordo/machine/dataset/data_provider/ncs_reader.py
```diff
@@ -346,13 +346,16 @@ def read_tag_files(
                 if dry_run:
                     logger.info("Dry run only, returning empty frame early")
                     return pd.Series()
-
+                before_downloading = timeit.default_timer()
                 with adls_file_system_client.open(file_path, "rb") as f:
                     df = file_type.read_df(f)
                     df = df.rename(columns={"Value": tag.name})
+                    df.index = pd.to_datetime(df.index, utc=True)
```
epa095: This looks a bit weird for the csv case now, with double converting

hjarraya: I am not sure in which date format in the current parquet files in datalake, if you think this is a double conversion, I'd rather have it than not.

epa095: The point is that the way it was before was that `file_type.read_df(f)` handles all conversion including time parsing. Now it still does that for the csv case (and then parses again here) while for the parquet it's only here. You should either handle the timeseries parsing inside `read_df(f)` for both the csv and parquet readers, or in none of them.

gordo/machine/dataset/data_provider/ncs_reader.py
```diff
@@ -346,13 +346,16 @@ def read_tag_files(
...
                     df = df[~df["Status"].isin(remove_status_codes)]
                     all_years.append(df)
-                    logger.info(f"Done parsing file {file_path}")
+                    logger.info(
+                        f"Done parsing in {(timeit.default_timer()-before_downloading):.2f} seconds for file {file_path}"
```
flikka: I suggest rewording this logging statement to clarify that it is both downloading and parsing that is being timed.

gordo/machine/dataset/data_provider/file_type.py
```diff
@@ -100,10 +99,6 @@ def __init__(self, time_series_columns: TimeSeriesColumns):
     def read_df(self, f: BinaryIO) -> pd.DataFrame:
         columns = self.time_series_columns.columns
         datetime_column = self.time_series_columns.datetime_column
-        pf = pa.ParquetFile(f)
-        table = pf.read(columns)
-        df = table.to_pandas()
-        df[datetime_column] = df[datetime_column].apply(
-            lambda v: pd.to_datetime(v, utc=True)
```
flikka, line 107: As @epa095 points out, if you remove the pd.to_datetime from the read_df in the ParquetFileType, and put it into the read_df in ncs_reader instead, you should remove it from the CsvFileType. Alternatively, just add the pd.to_datetime statement to ParquetFileType again, and remove it from read_df.

flikka (changes requested):
Two comments:
* The pd.to_datetime should be _either_ in the *FileType classes _or_ in the read_df. Not in "half of" the FileType classes AND in the read_df function, that is very confusing. 
* A slightly unrelated issue (perhaps a good idea to fix here, or in a separate issue): The Parquet files appears to _not_ be sorted on the timeindex. This will probably cause problems (either crashing, or even worse, silent logic catastropies) if the dataframes are _not_ resampled. Usually we do resample, and then I _think_ there is an automatic sorting happening.

gordo/machine/dataset/data_provider/file_type.py
hjarraya: moved it to ParqutFileType.

gordo/machine/dataset/data_provider/ncs_reader.py
hjarraya: You have a good point, moved to ParqutFileType.

hjarraya:
Added sort_index after reading each file.

flikka (approved):

flikka:
Goodie 👍

PR #977, opened 2020-04-01
add describe statistics and input histogram estimations

hjarraya:


gordo/machine/dataset/datasets.py
```diff
@@ -231,6 +232,26 @@ def get_data(self) -> Tuple[pd.DataFrame, Optional[pd.DataFrame]]:
...
+            hists = {}
+            for tag in X.columns:
+                step = round((X[tag].max() - X[tag].min()) / 100, 6)
+                if step < 9e-07:
+                    hists[str(tag)] = {}
+                    continue
+                outs = pd.cut(
+                    X[tag],
+                    bins=np.arange(
+                        round(X[tag].min() - step, 6),
```
henrikny: Why the +/- step?

henrikny (commented):
I am not convinced that using bins in such a way is particularly useful.
It is not a robust method.
Imagine you have a extreme outlier.
Your steps would be very large, and 99% of all your observations would reside inside one of them.

A simple list of quantiles is sufficient (e.g. at every 0.1 step), in addition to sd, min, max and arithmetic mean would be good enough.
Also, easier to show in Grafana.
If Gaussian, only the mean and standard deviation can be collected as sufficient statistics.

gordo/machine/dataset/datasets.py
hjarraya: to include extreme values buckets

hjarraya: it is more descriptive than quantiles, make it easier to visualise.

henrikny (approved):

PR #983, opened 2020-04-28
Statuses for gordo's models

koropets:
As a user I want to know what models succeeded/failed when building in Gordo


hjarraya (approved):
LGTM

PR #984, opened 2020-04-29
[MRG] List filtering

henrikny:
This will augment the function `pandas_filter_rows` to also handle a list of filtering criteria.

We met some limitations when using the concatenated string logic on `pd.DataFrame.eval` and the number of logical parts would exceed 32 (current dependency of numpy and pandas ) or 242 (latest release).

Using a list instead, `pd.DataFrame.eval` can handle 100 items (both in current dependency and latest release) and returns a `numpy.ndarray`.


henrikny:
I had not intention of linking this with #890 and have no idea how that happened or can be undone.

hjarraya (approved):
LGTM

PR #985, opened 2020-04-30
change Exceptions inheritance to "Exception"

devProdigy:
Fix for catching the errors from Gordo.
If custom exception is inherited from `BaseException` then it crashes the app that is using Gordo.
Usually in code if all exceptions wanted to be caught we can see smth like `except Exception as e`.

Accordingly to [docs](https://docs.python.org/3/library/exceptions.html#Exception) 
```
All built-in, non-system-exiting exceptions are derived from this class. All user-defined exceptions should also be derived from this class.
```

Don't know Gordo but I think that this won't fail the code, but please correct me if I'm wrong.

### DIFF
--- a/gordo/client/io.py
+++ b/gordo/client/io.py
@@ -5,7 +5,7 @@ from typing import Union, Optional
 import requests
 
 
-class HttpUnprocessableEntity(BaseException):
+class HttpUnprocessableEntity(Exception):
     """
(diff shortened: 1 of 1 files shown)


koropets (approved):

hjarraya (approved):
LGTM

PR #990, opened 2020-05-18
Daily window smoothed threshold + anomaly score

hugobettmach:
Adds the smoothed threshold and anomaly calculation as discussed with the SMEs. In summary:
* keep the current threshold as it is, at least until we decide to officially adopt the method below
* define a time window that should for now be defaulted to 1 day: 144 time steps @ 10T resolution
* calculate another threshold using the same logic here but with that window: `scaled_mse.rolling(144).min().max()`
* make sure another that threshold is set to the third fold like the current one here
* also calculate the feature_thresholds using the same method
* add the new thresholds to the metadata. keep the current ones in there for now
* calculate a scaled and unscaled version of the smoothed total_anomaly using the same rolling window and the median, e.g. `data["smooth-total-anomaly-scaled"] = data["total-anomaly-scaled"].rolling(144).median()`. Do the same for tag-level anomalies.
* recalculate anomaly confidence with the new threshold and the new total_anomaly, also for the tag-level anomaly confidence.

### DIFF
--- a/gordo/cli/workflow_generator.py
+++ b/gordo/cli/workflow_generator.py
@@ -105,7 +105,7 @@ def workflow_cli(gordo_ctx):
 @click.option(
     "--split-workflows",
     type=int,
-    default=50,
+    default=30,
     help="Split workflows containg more than this number of models into several "
     "workflows, where each workflow contains at most this nr of models. The "
     "workflows are outputted sequentially with '---' in between, which allows "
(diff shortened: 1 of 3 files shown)


hjarraya (approved):
LGTM

hjarraya:
Please run black formatting to fix the build.

hjarraya (changes requested):
LGTM, fix build, run black formatting.

koropets (changes requested):
Looks good, but there is still some failed test [here](https://github.com/equinor/gordo/pull/990/checks?check_run_id=700141362)

hjarraya (approved):
LGTM

PR #1000, opened 2020-06-11
Fixing a couple of minor bugs

koropets:
- An issue with empty `window` argument during building the model with custom window smoothed threshold
- Fixing models statuses management for large gordos split by independent Argo-workflows, the issue, in this case: if some of these workflows failed the whole gordo is going to be failed
- Fixing dependencies of the `gordo-client` argo workflow step for gordos with `influx.enable=true`


hjarraya (approved):
LGTM

PR #1001, opened 2020-06-17
[WIP] adds filtering of global thresholds

henrikny:
There is a need to filter training data based on a global threshold.
This is meant as a pure sanity-check of the data, as extremely high/low values have been seen (e.g. 7e+80). 

Alternatively, such a filtering could be done on the data, and drop-periods could be added to the model specific model filter list in a separate step.
To simply handle it directly in the filter-step seems simpler.
Disadvantage is that one does not know wether or not there has been removed data due to global threshold filtering criteria.
I do, however, not see this as a big problem.

### DIFF
--- a/gordo/machine/dataset/filter_rows.py
+++ b/gordo/machine/dataset/filter_rows.py
@@ -138,10 +138,6 @@ def pandas_filter_rows(df, filter_str: str, buffer_size: int = 0):
     apply_buffer(pandas_filter, buffer_size=buffer_size)
     df = df[list(pandas_filter)]
     logger.info("Shape of data after numerical filtering: %s", df.shape)
-    logger.info("Applying global min/max filtering")
-    mask = ((df > -1000) & (df < 50000)).all(1)
-    df = df[mask]
-    logger.info("Shape of data after global min/max filtering: %s", df.shape)
     return df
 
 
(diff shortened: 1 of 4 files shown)


gordo/machine/dataset/filter_rows.py
```diff
@@ -133,8 +134,15 @@ def pandas_filter_rows(df, filter_str: str, buffer_size: int = 0):
     # and returns a pd.Series
     else:
         pandas_filter = df.eval(filter_str)
+
     apply_buffer(pandas_filter, buffer_size=buffer_size)
-    return df[list(pandas_filter)]
+    df = df[list(pandas_filter)]
+    logger.info("Shape of data after numerical filtering: %s", df.shape)
+    logger.info("Applying global min/max filtering")
+    mask = ((df > -1000) & (df < 50000)).all(1)
```
koropets: I do not think this is a good place to put these values here. We could have these values by default, but the user should be able to change these values as `min_global_threshold` and `max_global_threshold` through the gordo resource config.

hjarraya: Also, can we double-check these values?

henrikny: Sure! I agree. This was obviously a "quick fix", and to show what is wanted.
It is however not completely clear to me how those parameters would be picked up from the config and where they eventually would be evaluated.
Any advice would be appreciated.

hugobettmach: I agree with @koropets that these default values should be set to function args defaults with the possibility to customize them from the config, same way as we are passing the `row_filter_buffer` for example.
@henrikny I also have no problem with this filtering being located here.
@hjarraya these were the values discussed yesterday with the SME's and seem to be reasonable defaults from their perspective.

koropets (approved):

PR #1010, opened 2020-07-16
[MRG] No forward fill and augments drop status codes

henrikny:
Trying to avoid the problem of having flatlined signals caused by forward filled values by using
1) interpolation in sted of forward fill
and
2) max interpolation of future points: 8h.

Also, more "non-normal" status codes are added as default to the drop-criteria:
(https://techsupport.osisoft.com/Documentation/PI-AF-SDK/html/T_OSIsoft_AF_Asset_AFValueStatus.htm)

One test is removed, and another adapted to the new returned data frame shape.

### DIFF
--- a/gordo/machine/dataset/base.py
+++ b/gordo/machine/dataset/base.py
@@ -234,4 +234,4 @@ class GordoBaseDataset:
         # given granularity in minutes
         granularity = int(re.findall(r"\d+", resolution)[0])
 
-        return resampled.interpolate(limit=(480 / granularity)).dropna()
+        return resampled.interpolate(limit=(480 / granularity))#.dropna()
--- a/gordo/machine/dataset/base.py
+++ b/gordo/machine/dataset/base.py
@@ -4,6 +4,7 @@ import abc
 import logging
 from typing import Iterable, Union, List, Callable, Dict, Any, Tuple
 from datetime import datetime
+import re
 
 import pandas as pd
 import numpy as np
@@ -228,4 +229,9 @@ class GordoBaseDataset:
             resampled.columns = pd.MultiIndex.from_product(
                 [[series.name], resampled.columns], names=["tag", "aggregation_method"]
             )
-        return resampled.dropna()
+
+        # ensure interpolation max eight hours (8 x 60 minutes) forward
+        # given granularity in minutes
+        granularity = int(re.findall(r"\d+", resolution)[0])
+
+        return resampled.interpolate(limit=(480/granularity)).dropna()
--- a/gordo/machine/dataset/base.py
+++ b/gordo/machine/dataset/base.py
@@ -4,7 +4,7 @@ import abc
 import logging
 from typing import Iterable, Union, List, Callable, Dict, Any, Tuple
 from datetime import datetime
-import re
+from copy import copy
 
 import pandas as pd
 import numpy as np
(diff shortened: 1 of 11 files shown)


hugobettmach:
@henrikny I've tried this out briefly on a few pumps. Results _as usual_ depend on the given equipment but what I see in several pumps is troubling. My guess is that there are some tags that wreak havoc.
* Sometimes most of the data is kept e.g. **A-23PA001A**
* Other times upwards of 90% of the data is removed! e.g. **A-21PA101A**, **T-50PA001A** or **23-KZ01A** (this last one, the Sleipner gas compressor is extreme, **since no data remains**...)
* In the cases where most data is removed, counter-intuitively when changing the resolution from `'10T'` to `'5T'` or `'1T'` the amount of data dropped is even larger! I think this may be at least in part because of the `row_filter_buffer`, but not only, since I also tried setting that to `0`.

Tagging @eua89 and @hjarraya here so they can follow the discussion

gordo/machine/dataset/base.py
```diff
@@ -230,4 +231,9 @@ def _resample(
             resampled.columns = pd.MultiIndex.from_product(
                 [[series.name], resampled.columns], names=["tag", "aggregation_method"]
             )
-        return resampled.fillna(method="ffill")
+
+        # ensure interpolation max eight hours (8 x 60 minutes) forward
+        # given granularity in minutes
+        granularity = int(re.findall(r"\d+", resolution)[0])
```
koropets: A resolution could have various numbers of [units](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Timedelta.html). My suggestion is:
```
         td = pd.Timedelta(resolution)
         granularity = td.total_seconds() / 60
```

gordo/machine/dataset/base.py
```diff
@@ -230,4 +231,9 @@ def _resample(
...
+
+        return resampled.interpolate(limit=int(480 / granularity)).dropna()
```
koropets: `limit` could be a optinal:
```
         interpolate_limit = None
         if self.interpolate_limit is None:
              interpolate_limit = int(480 / granularity)
         elif self.interpolate_limit != 0:
              interpolate_limit = self.interpolate_limit
         if interpolate_limit is not None:
              return resampled.interpolate(limit=interpolate_limit).dropna()
        else:
              return resampled.fillna(method="ffill")

koropets (changes requested):

henrikny (commented):
I will implement the suggested changes :)

gordo/machine/dataset/base.py
HugoBettMach: @henrikny I've tried this on several pumps and it seems to be working fine. It actually stops the data from dead tags being forward-filled which was one of our main issues. Very promising!

Of course on certain tags, especially the calculated diff tags, the change in behaviour is very apparent and I've tried to sketch it below. Diff tags in some of the models have a value calculated every 6 hours, before they used to have a step like aspect and now are interpolated. I don't particularly see a big problem with this but would like to hear your opinion.
ffillvslininterp

hugobettmach (commented):
@henrikny I've tried this on several pumps and it seems to be working fine. It actually stops the data from dead tags being forward-filled which was one of our main issues. Very promising!

Of course on certain tags, especially the calculated diff tags, the change in behaviour is very apparent and I've tried to sketch it below. Diff tags in some of the models have a value calculated every 6 hours, before they used to have a step like aspect and now are interpolated. I don't particularly see a big problem with this but would like to hear your opinion.
ffillvslininterp

gordo/machine/dataset/base.py
henrikny: Good to hear that it seems to solve some of the issues we've been having with regards to "frozen values".

Regarding the ffill vs. interpolation on calculated diff tags, the actual raw data is also a time-aggregated statistic, probably an average over e.g. 6 hours. In these cases, I think linear interpolation is actually a better method than forward filling. The only situation where this would not be the case, is if we were dealing with nominal scaled data having digital states such as 1=starting, 2=running, 3=shutting down, 4=stopped. As far as I know, we are not dealing with any categorical type of data.

PR #1022, opened 2020-08-13
Upgrade tensorflow up to 2.2.0

koropets:


hugobettmach (commented):
Is 2.3.0 breaking stuff?

koropets:
@HugoBettMach Yes, looks like
```
>               history.epoch = self.model.history.epoch
E               AttributeError: 'History' object has no attribute 'epoch'
```

koropets:
@HugoBettMach But looks like this line becomes unnecessary with tensorflow 2.2.0
```python
                 history = History()
                 history.history = self.model.history.history
                 history.params = self.model.history.params
-                history.epoch = self.model.history.epoch
                 state["history"] = history
```
because of `self.model.history.params` containing `epoch`
```python
{'verbose': 0, 'epochs': 1, 'steps': 1}
```

hugobettmach:
Yes looks like, just drop it

hugobettmach:
There are still issues with having a newer version of TensorFlow, right? Is it better then to keep 2.1.0 and I revert back to Python 3.7?

PR #1034, opened 2020-08-26
[WIP] Adds filter periods

### LINKED ISSUE #1031: Include extreme outlier removal workflow as part of gordo
* possibly as a step in the building workflow, possibly after this line https://github.com/equinor/gordo/blob/77780438c6bd0d565229de3c86daabaa83df07a7/gordo/builder/build_model.py#L181
* able to declare desired configuration (e.g. method `['no_cleaning', 'rolling_median_iqr', 'ifor', etc etc]` and args to such methods in the yaml config, both at global and dataset levels. same as is done with model configs currently.

henrikny:
Median and isolation forest filtering directly available in Gordo.

Resolves #1031

There seem to be some randomness or rounding issues coming from the isolation forest algorithm and/or the exponential weighting function.
The isolation forest has its `random_state` parameter sat.
At every execution, the shape of the resulting filtered data frame changes, making it impossible to set a fixed assert.

In addition to this, evaluating it I get different results for using
- local `pytest` direct on file (e.g. `pytest tests/gordo/machine/dataset/test_filter_periods.py`)
- local `pytest`on testset (e.g. `python setup.py testmachine`)
- direct console execution
- CircleCI jobs via commit to GitHub

Base functions used:
https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.IsolationForest.html
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.ewm.html

**Reviewers:**
- Please check if the placement of the code seems reasonable.
- Please do a quick overlook over the logic, and comment if the functions could be written more understandable or efficient.
- Please rapport if the attached test scheme is sufficient or not.
- Any insight on the issues with the always-changing filtered data shape

### DIFF
--- a/gordo/machine/dataset/datasets.py
+++ b/gordo/machine/dataset/datasets.py
@@ -264,7 +264,7 @@ class TimeSeriesDataset(GordoBaseDataset):
                 )
 
         if self.filter_periods:
-            data, drop_periods = self.filter_periods.filter_df(data)
+            data, drop_periods, _ = self.filter_periods.filter_data(data)
             self._metadata["filtered_periods"] = drop_periods
 
         x_tag_names = [tag.name for tag in self.tag_list]
--- a/gordo/machine/dataset/datasets.py
+++ b/gordo/machine/dataset/datasets.py
@@ -253,6 +253,15 @@ class TimeSeriesDataset(GordoBaseDataset):
                     f"specified required threshold for number of rows ({self.n_samples_threshold})."
                 )
 
+        if self.filter_periods:
+            filter_method = self.filter_periods.get("filter_method")
+
+            self.filter_periopds.get("filter")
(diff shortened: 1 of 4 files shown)


gordo/machine/dataset/datasets.py
```diff
@@ -253,6 +259,13 @@ def get_data(self) -> Tuple[pd.DataFrame, Optional[pd.DataFrame]]:
                     f"specified required threshold for number of rows ({self.n_samples_threshold})."
                 )
 
+        if self.filter_periods:
+            filtered_periods = filter_periods(
+                data=data, granularity=self.resolution, kwargs=self.filter_periods
```
koropets: Expression `kwargs=self.filter_periods` will not be working here

koropets (changes requested):

koropets (approved):

Issue #1038, opened 2020-09-10
Speeding up row filtering


tommyod:
Expressing the following function as convolution:

https://github.com/equinor/gordo/blob/0887eb0555c9a4691b34174921a00ad6a52a308d/gordo/machine/dataset/filter_rows.py#L8

Yields a spedup of ~1000x.

Here is a sketch of how the function might look like as a convolution:

```python

def apply_buffer2(mask: pd.Series, buffer_size: int = 0):
    # TODO: Cases when len(kernel) > len(array) are not handled, but should be
    array = 1 - np.array(mask.to_numpy(), dtype=float)
    kernel = np.ones(buffer_size*2 + 1, dtype=float)
    ans = np.convolve(a=array, v=kernel, mode='same')
    mask.values[:] = (ans < 1)

```
Here are simple timings of the existing solution vs. the proposed solution:
```
Speedup of factor 1291.21 for array of shape 10^4
Speedup of factor 1465.28 for array of shape 10^5
Speedup of factor 882.95 for array of shape 10^6
Speedup of factor 640.74 for array of shape 10^7
```

PR #1040, opened 2020-09-17
[MRG] Speed up filtering

henrikny:
This speeds up the filtering and buffer application.
Solves #1038

Adds `know_filter_periods` to config class.
Removes unnecessary error-class.
Also fixes a bug when evaluating global low/high thresholds.

### DIFF
--- a/gordo/machine/dataset/datasets.py
+++ b/gordo/machine/dataset/datasets.py
@@ -14,7 +14,11 @@ from gordo.machine.dataset.data_provider.providers import (
     RandomDataProvider,
     DataLakeProvider,
 )
-from gordo.machine.dataset.base import GordoBaseDataset, InsufficientDataError, ConfigurationError
+from gordo.machine.dataset.base import (
+    GordoBaseDataset,
+    InsufficientDataError,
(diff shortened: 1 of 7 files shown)


gordo/machine/dataset/datasets.py
```diff
@@ -240,32 +238,50 @@ def get_data(self) -> Tuple[pd.DataFrame, Optional[pd.DataFrame]]:
...
                     f"specified required threshold for the number of rows ({self.n_samples_threshold}), "
-                    f" after applying the specified row-filter."
+                    f" after applying the specified numerical row-filter."
                 )
 
-        if self.low_threshold and self.high_threshold:
+        if isinstance(self.low_threshold, int) and isinstance(self.high_threshold, int):
+            assert (
+                self.low_threshold < self.high_threshold
+            ), "Low threshold need to be larger than high threshold"
```
koropets: I think it is better to have this raised some custom exception as we did with `InsufficientDataError` just to keep in mind that in the future we will go to some  sort of custom retry policy logic, and in this case, we will just ignore some type of exceptions

PR #1088, opened 2020-12-01
[WIP] init commit for flakehell and black checks in the project

devProdigy:
Test adding [flakehell](https://github.com/life4/flakehell/blob/master/README.md) code checker and [black](https://black.readthedocs.io/en/stable/) formatter to the project and CI.
Check should be made only for the new changes brought in the PRs, old code will be "unified" with time.
Styles and best practices to be checked/allowed/forbidden by checker - TBD. Feel free to contribute :)


flikka:
Isn't black already invoked with the test https://github.com/equinor/gordo/blob/master/tests/test_formatting.py?

devProdigy:
> Isn't black already invoked with the test https://github.com/equinor/gordo/blob/master/tests/test_formatting.py?

hm, seems like it's, but in a little bit weird way (through tests). And I can't see usual configs for black.
Maybe it' better to check for formatting and code-quality before running tests - cause tests running is time consuming and usually it's better to know that you misses smth as early as you can.
Maybe it's a good idea additionally to have `pre-commit hooks` for such tasks?

So we'll have (before local push and on the CI): 
- flake checks of code quality;
- black formatting checks.

@flikka What do you think?

Update:
Run Black on the project when it was really turned on and got lots of error (you can see it in the CI job for this PR).
Added just one param to `pyproject.toml` file:
```
[tool.black]  
line-length = 120
```
Or maybe I missed smth?

flikka:
> hm, seems like it's, but in a little bit weird way (through tests). And I can't see usual configs for black.

The configs were kept default (therefore no configs) intentionally, to make it as "standard" as possible (according to Black).

> Maybe it's a good idea additionally to have pre-commit hooks for such tasks?

Agreed, many commits in Gordo's PR's are  "fix formatting, or fix black", so I totally agree! One can also make it integrated in your own dev environment, for example in PyCharm with auto-black on save or similar.

> Run Black on the project when it was really turned on and got lots of error (you can see it in the CI job for this PR).

If you see the CI for other PRs, or main, you can see that black is run, with no errors. The reason you get errors now is precisely because you also change the black config by setting the `line-length = 120` parameter, thus changing black's line length from 88 to 120. This will then cause _a lot_ of changes, naturally causing black to fail. As it should. If you want to introduce this (should be discussed among the Gordo developers, it is quite a massive change, and some people have strong feelings about this) - you should create a PR that _only_ does that, and tell others that you are doing it. It wil change almost all of the files in the repo, so merging other ongoing PRs can be hellish. 

Keep up the good work :+1:

devProdigy:
Closing this PR due to some issues.
New PR is [here](https://github.com/equinor/gordo/pull/1099)

PR #1299, opened 2023-02-03
AB#98780: Update docker image to latest debian stable

toera:


Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN cat /code/requirements/full_requirements.txt | grep tensorflow== > /code/pre
     && cat /code/requirements/full_requirements.txt | grep scipy== >> /code/prereq.txt \
     && cat /code/requirements/full_requirements.txt | grep catboost== >> /code/prereq.txt
 
-FROM python:3.9-slim-buster
+FROM python:3.9-slim
```
koropets:
```suggestion
FROM python:3.9-slim-bullseye
```
using `3.9-slim` would potentially implicitly break runtime's back-compatibility between Debian releases

PR #1385, opened 2024-06-26
Feat: Upgrade `tensorflow` to `2.16.1` and `keras` to `3.4.0`

RollerKnobster:
To accommodate the upgrade, next changes to the code were made:

* `tensorflow` no longer has keras wrappers, which are now used from `scikeras` library, recommended for use by them. They have largely the same interface, but slightly different innerworkings, which were accomodated for.
* Since `keras 3.*` now clones the input `regressor` that is provided to the `BaseEstimator` into a `regressor_` attribute, and ONLY updates the `regressor_` one after fitting, metadata extraction in `gordo.builder.build_model.ModelBuilder` was changed to seek the `regressor_` in the `steps`.
* Accessing `keras.optimizers` has changed in `keras 3.*` to having to use the `keras.optimizers.get` method and providing a `class_name` and `config` to deserialize. Relevant for lstm and feedforward autoencoders.
* Model config now requires for `tensorflow.keras.models.Sequential` to define `input_shape` in its layers, otherwise model is not properly built after compiling and prior to fitting. Relevant for `KerasRawModelRegressor`.

`KerasBaseEstimator` underwent the most changes.

* We now need to run the `__init__` of the KerasRegressor with the expected `kwargs` for proper initialisation of the object, but the `kwargs` will always take precedence for `fit`, `predict` and `compile`, so they are mostly for making `keras` happy.
* Saving model for pickling was changed ƒrom `h5` format, which is now considered legacy to `keras` native zip format, and the file is now stored to a temporary file and read into a buffer instead of using the `h5py` library, since `save_model` and `load_model` now exclusively expect an actual file instead of an io buffer.
* Current implementation of model preparation for fit depends on the `__call__` method, which is no longer used in `keras 3.*`, and was changed to `_prepare_model` to replicate the same behaviour right before we call our own `fit` since it requires the `X` and `Y` to be present to set the `n_features` and `n_features_out`.
* `History` is no longer returned from calling `fit` and must be extracted from under `model.history`.
* Manually stored history now resides in `self._history` instead of `self.history` to avoid attribute name clash.

Adjustments to tests were made:

* `input_shape` now resides under `input.shape` in `model.layers`.
* `optimizer_kwargs.lr` is now `optimizer_kwargs.learning_rate`

### DIFF
--- a/gordo/builder/build_model.py
+++ b/gordo/builder/build_model.py
@@ -556,7 +556,9 @@ class ModelBuilder:
         # which also had a GordoBase as a parameter/attribute, but will satisfy BaseEstimators
         # which can take a GordoBase model as a parameter, which will then have metadata to get
         for key, val in model.__dict__.items():
-            if key == "regressor":  # keras3 clones the regressor into regressor_ and never updates original regressor
+            # keras3 clones the regressor into regressor_ and never updates original regressor,
+            # but still stores both in attributes
+            if key == "regressor":
                 continue
             if isinstance(val, Pipeline):
                 metadata.update(
--- a/gordo/builder/build_model.py
+++ b/gordo/builder/build_model.py
@@ -557,7 +557,7 @@ class ModelBuilder:
         # which can take a GordoBase model as a parameter, which will then have metadata to get
         for key, val in model.__dict__.items():
             if key.endswith(
-                    "_"
+                "_"
             ):  # keras3 clones the regressor into regressor_ and never updates original regressor
                 if isinstance(val, Pipeline):
                     metadata.update(
--- a/gordo/builder/build_model.py
+++ b/gordo/builder/build_model.py
@@ -555,13 +555,16 @@ class ModelBuilder:
         # Continue to look at object values in case, we decided to have a GordoBase
         # which also had a GordoBase as a parameter/attribute, but will satisfy BaseEstimators
         # which can take a GordoBase model as a parameter, which will then have metadata to get
-        for val in model.__dict__.values():
-            if isinstance(val, Pipeline):
-                metadata.update(
-                    ModelBuilder._extract_metadata_from_model(val.steps[-1][1])
-                )
-            elif isinstance(val, GordoBase) or isinstance(val, BaseEstimator):
-                metadata.update(ModelBuilder._extract_metadata_from_model(val))
+        for key, val in model.__dict__.items():
+            if key.endswith(
+                    "_"
+            ):  # keras3 clones the regressor into regressor_ and never updates original regressor
+                if isinstance(val, Pipeline):
+                    metadata.update(
+                        ModelBuilder._extract_metadata_from_model(val.steps[-1][1])
+                    )
+                elif isinstance(val, GordoBase) or isinstance(val, BaseEstimator):
+                    metadata.update(ModelBuilder._extract_metadata_from_model(val))
         return metadata
 
     @property
--- a/gordo/builder/build_model.py
+++ b/gordo/builder/build_model.py
@@ -556,15 +556,14 @@ class ModelBuilder:
         # which also had a GordoBase as a parameter/attribute, but will satisfy BaseEstimators
         # which can take a GordoBase model as a parameter, which will then have metadata to get
         for key, val in model.__dict__.items():
-            if key.endswith(
-                "_"
-            ):  # keras3 clones the regressor into regressor_ and never updates original regressor
-                if isinstance(val, Pipeline):
-                    metadata.update(
-                        ModelBuilder._extract_metadata_from_model(val.steps[-1][1])
-                    )
-                elif isinstance(val, GordoBase) or isinstance(val, BaseEstimator):
-                    metadata.update(ModelBuilder._extract_metadata_from_model(val))
+            if key == "regressor":  # keras3 clones the regressor into regressor_ and never updates original regressor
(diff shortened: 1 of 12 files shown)


gordo/builder/build_model.py
```diff
@@ -555,13 +555,16 @@ def _extract_metadata_from_model(
...
-            if isinstance(val, Pipeline):
-                metadata.update(
-                    ModelBuilder._extract_metadata_from_model(val.steps[-1][1])
-                )
-            elif isinstance(val, GordoBase) or isinstance(val, BaseEstimator):
-                metadata.update(ModelBuilder._extract_metadata_from_model(val))
+        for key, val in model.__dict__.items():
+            if key.endswith(
+                "_"
+            ):  # keras3 clones the regressor into regressor_ and never updates original regressor
```
koropets:
```suggestion
            if key == "regressor":  # keras3 clones the regressor into regressor_ and never updates original regressor
                continue
```
endwith is too broad to fix one back-compatibility issue

gordo/machine/model/models.py
```diff
@@ -78,11 +76,22 @@ def __init__(
...
         self.kwargs: Dict[str, Any] = kwargs
+        self._history = None
+
+        # This new keras wrapper expects most of these kwargs to be set to the model attributes and uses them for
+        # defaults in some places, but always gives precedence to kwargs passed to respective fit, predict and compile
+        # methods, so this is just to make it happy again
+        _expected_kwargs = {
+            *self._fit_kwargs,
+            *self._predict_kwargs,
+            *self._compile_kwargs,
```
koropets:
```suggestion
            *KerasRegressor._fit_kwargs,
            *KerasRegressor._predict_kwargs,
            *KerasRegressor._compile_kwargs,
```
in case of self not clear what an exact source of these constants

gordo/machine/model/models.py
```diff
@@ -301,25 +313,25 @@ def get_params(self, **params):
             Parameters used in this estimator
         """
         params = super().get_params(**params)
-        params.pop("build_fn", None)
+        params.pop("model", None)
```
koropets:
```suggestion
```
unnecessary call

tests/gordo/machine/model/test_raw_keras.py
```diff
@@ -80,8 +81,12 @@ def test_raw_keras_part_of_pipeline():
                             layers:
                                 - tensorflow.keras.layers.Dense:
                                     units: 4
+                                    input_shape:
+                                        - 4
```
koropets:
```suggestion
                                    input_shape: [4]
```
just suggestion

tests/gordo/machine/model/test_raw_keras.py
```diff
@@ -80,8 +81,12 @@ def test_raw_keras_part_of_pipeline():
...
                                 - tensorflow.keras.layers.Dense:
                                     units: 1
+                                    input_shape:
+                                        - 1
```
koropets:
```suggestion
```

koropets (approved):

gordo/machine/model/factories/feedforward_autoencoder.py
```diff
@@ -88,8 +88,9 @@ class (e.x. Adam(lr=0.01,beta_1=0.9, beta_2=0.999)).  If no arguments are
 
     # Instantiate optimizer with kwargs
     if isinstance(optimizer, str):
-        Optim = getattr(keras.optimizers, optimizer)
-        optimizer = Optim(**optimizer_kwargs)
```
hugobettmach, line 92: I think this previous approach still works though. It's what we have implemented in aion and runs fine. But I'm ok with changing it.

gordo/machine/model/models.py
```diff
@@ -2,25 +2,23 @@
...
 from copy import copy, deepcopy
 from importlib.util import find_spec
 
-import h5py
 import tensorflow.keras.models
 from tensorflow.keras.models import load_model, save_model
 from tensorflow.keras.preprocessing.sequence import pad_sequences, TimeseriesGenerator
-from tensorflow.keras.wrappers.scikit_learn import KerasRegressor as BaseWrapper
-from tensorflow.keras.callbacks import History
+from scikeras.wrappers import KerasRegressor
```
hugobettmach, line 16: Interesting I wasn't aware about this. Good you found this out.

gordo/machine/model/models.py
```diff
@@ -78,11 +76,22 @@ def __init__(
...
+
+        # This new keras wrapper expects most of these kwargs to be set to the model attributes and uses them for
+        # defaults in some places, but always gives precedence to kwargs passed to respective fit, predict and compile
+        # methods, so this is just to make it happy again
+        _expected_kwargs = {
+            *KerasRegressor._fit_kwargs,
+            *KerasRegressor._predict_kwargs,
+            *KerasRegressor._compile_kwargs,
+        }
+        KerasRegressor.__init__(
```
hugobettmach, line 91: This is indeed a bit strange, but oh well, as long as it works. Fortunately we won't need this wrapping business in aion.

hugobettmach (commented):
Wow quite a lot of tricky changes to make, good work @RollerKnobster in finding these out. It's hard to have an opinion if they will work or not but once we have a pre-release of gordo-anomaly I will use my local dieta to test it

